PR #8, opened 2019-03-07
Tor proxy, always-use-proxy, bind to localhost clightning

nixbitcoin:
Sets a Tor proxy for all outgoing clightning connections, which allows users to connect to other .onion Lightning nodes. Forces clightning to always use Tor proxy. Binds clightning to Tor HS at 127.0.0.1:9735


jonasnick:
Nice, will look at this later.

jonasnick:
ACK

PR #9, opened 2019-03-14
Add proxy and hidden service to liquidd

nixbitcoin:
This commit adds Tor support to liquidd. Should it include the external ip option to advertise the hidden service for inbound connections?

### DIFF
--- a/modules/nix-bitcoin.nix
+++ b/modules/nix-bitcoin.nix
@@ -153,9 +153,10 @@ in {
     ";
     services.liquidd.listen = true;
     services.liquidd.proxy = config.services.tor.client.socksListenAddress;
+    services.liquidd.port = 7042;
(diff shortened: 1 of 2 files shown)


modules/nix-bitcoin.nix
```diff
@@ -151,7 +151,15 @@ in {
       mainchainrpcuser=${config.services.bitcoind.rpcuser}
       mainchainrpcport=8332
     ";
-
+    services.liquidd.listen = true;
+    services.liquidd.proxy = config.services.tor.client.socksListenAddress;
+    services.tor.hiddenServices.liquidd = {
+      map = [{
+        port = 7042; toPort = 7042;
```
jonasnick: use `config.services.liquidd.port`?

jonasnick:
ACK

PR #13, opened 2019-03-25
NixOS fixes + /var/lib/bitcoind/blocks

clefru:
Fixes mostly for NixOS powered deployments.
Plus a fix for creating /var/lib/bitcoind/blocks.

Thanks for your work, Jonas!


jonasnick:
This is excellent. Just recently someone report that deployments from NixOS didn't work, but I didn't get around debugging it. Thank you very much for your contribution!

Initially I thought "import from derivation" was more idiomatic and elegant for some reason but having the dependencies managed by nix is clearly the way to go.

PR #14, opened 2019-03-26
Import bitcoind banlist in separate service

jonasnick:
Otherwise it'll take a couple of seconds to wait for every addban RPC to time out. CC @nixbitcoin

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ The `nix-bitcoin.nix` module provides the two profiles "minimal" and "all":
 
 * minimal
     * bitcoind (pruned) with outbound connections through Tor and inbound connections through a hidden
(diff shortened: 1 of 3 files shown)


jonasnick:
perhaps the banlist adder should be its own service

jonasnick:
Made banlist import its own service

nixbitcoin:
What benefits are there from making banlist import its own service over fixing waiting on the addban RPC to time out in the bitcoind postStart?

jonasnick:
Sorry, should have mentioned this. The benefit is that if addban is called in bitcoind postStart it takes a very long time to restart the bitcoind service because it only counts as started once the postStart script is done.

PR #15, opened 2019-03-26
node2nix cleanups

clefru:
This mostly fixes a problem with spark-wallet not having access to the qrcode-terminal package. The rest are clean-ups. Sadly spark-wallet -Q still doesn't print me my access URL.


jonasnick:
Looks good, thanks again. I had noticed the crash with `-Q`. That's not the case anymore, but do you see a QR code? I don't.

pkgs/nanopos/supplement.nix
```diff
@@ -3,681 +3,6 @@
...
-      name = "yallist";
-      packageName = "yallist";
-      version = "3.0.3";
-      src = fetchurl {
-        url = "https://registry.npmjs.org/yallist/-/yallist-3.0.3.tgz";
-        sha512 = "S+Zk8DEWE6oKpV+vI3qWkaK+jSbIK86pCwe2IF/xwIpQ8jEuxpw9NyaGjmp9+BoJv5FV2piqCDcoCtStppiq2A==";
-      };
-    };
-  };
+  sources = {};
```
jonasnick: wouldn't it be better if we remove the `--supplement` flag from generate.sh?

jonasnick:
Oh, I should've read the OP more closely ... you're saying that it doesn't print a QR code for you.

pkgs/nanopos/supplement.nix
clefru: I felt symmetry of "generate.sh" among all node2nix-packaged packages was more important than those empty files. I am happy to customize it.

clefru:
> Looks good, thanks again. I had noticed the crash with `-Q`. That's not the case anymore, but do you see a QR code? I don't.

I do think that this is the case because of this line: https://github.com/shesek/spark-wallet/blob/e1c7018c3120336be2e1f5d4c0ea94b8c3be2146/src/app.js#L74
```
    if (qrterm && !url.includes('://localhost:')) {
```
and my console prints HTTP server running on http://localhost:9737. We probably have to start it with the onion endpoint supplied to "--public-url" 
```
    --public-url <url>       override public URL used for QR codes [default: http(s)://{host}/]
```

jonasnick:
Yup, that works.

jonasnick:
Working on a fix. ACK a0b4d5b9325ddbbc3aafd9051aae50a1acb2aa9c

PR #16, opened 2019-03-27
Fix spark wallet QR code display

jonasnick:
CC @clefru @nixbitcoin

### DIFF
--- a/modules/nix-bitcoin.nix
+++ b/modules/nix-bitcoin.nix
@@ -172,12 +172,7 @@ in {
     services.nix-bitcoin-webindex.enable = cfg.modules == "all";
     services.clightning.autolisten = cfg.modules == "all";
     services.spark-wallet.enable = cfg.modules == "all";
-    services.tor.hiddenServices.spark-wallet = {
-      map = [{
-        port = 80; toPort = 9737;
-      }];
-      version = 3;
-    };
+    services.spark-wallet.onion-service = true;
     services.electrs.enable = false;
     services.electrs.high-memory = false;
     services.tor.hiddenServices.electrs = {
--- a/modules/spark-wallet.nix
+++ b/modules/spark-wallet.nix
@@ -4,6 +4,19 @@ with lib;
 
 let
   cfg = config.services.spark-wallet;
+  dataDir = "/var/lib/spark-wallet/";
+  onion-chef-service = (if cfg.onion-service then [ "onion-chef.service" ] else []);
+  run-spark-wallet = pkgs.writeScript "run-spark-wallet" ''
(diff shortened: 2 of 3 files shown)


modules/spark-wallet.nix
```diff
@@ -20,16 +22,49 @@ in {
...
+        }
+        echo $CMD > ${dataDir}/startscript.sh
+        echo Running $(cat ${dataDir}/startscript.sh)
+        chown -R 'clightning:clightning' ${dataDir}
+        chmod u+x ${dataDir}/startscript.sh
+      '';
       serviceConfig = {
-        ExecStart = "${pkgs.spark-wallet}/bin/spark-wallet --ln-path ${cfg.ln-path} -k -c /secrets/spark-wallet-login";
+        PermissionsStartOnly = "true";
+        ExecStart = "${pkgs.bash}/bin/bash ${dataDir}/startscript.sh";
```
clefru: I would use pkg.writeScript here instead of generating a script inside of preStart. Can't the work of preStart not be done inside of ExecStart as well?

jonasnick: Yes, that would be nicer. But we still need to do some work as root because the user doesn't have permissions to read `/var/lib/tor/onion/spark-wallet/hostname`.

clefru: How about just copying & chowning hostname in preStart to ${dataDir}? Seems like an equal amount of work :)

But overall ACK, doesn't seem to hard to refactor later.

nixbitcoin: > Yes, that would be nicer. But we still need to do some work as root because the user doesn't have permissions to read `/var/lib/tor/onion/spark-wallet/hostname`.

I am working on this fixing `nodeinfo` so it can be run as operator. If the hostnames are tor group readable, clightning user could be added to host.

modules/nix-bitcoin.nix
```diff
@@ -168,12 +168,7 @@ in {
     services.nix-bitcoin-webindex.enable = cfg.modules == "all";
     services.clightning.autolisten = cfg.modules == "all";
     services.spark-wallet.enable = cfg.modules == "all";
-    services.tor.hiddenServices.spark-wallet = {
-      map = [{
-        port = 80; toPort = 9737;
-      }];
-      version = 3;
-    };
+    services.spark-wallet.onion-service = true;
```
nixbitcoin: Would you consider handling all modules onion services like this from now on? Or would it make the modules less usable for other projects since those might not want to use a hidden service out of the box?

jonasnick: > Would you consider handling all modules onion services like this from now on?

Only if necessary as in this case

jonasnick:
Now based on #20 onion manager

PR #17, opened 2019-03-28
Nurify

clefru:
This is mostly about the second commit:

Make repository importable as NUR (including an overlay)

https://github.com/nix-community/NUR is a Nix community project that
aims to make out of tree derivations more easily discoverable and
accessible to Nix users. Converting the nix-bitcoin repo to conform to
that style is a minor change and enhances reusability of its
components. For instance, I could slap on the clightning module more
easily onto my existing bitcoin node without having to redeploy the
whole as nixops driven installation. Having the repo in NUR style
would make that easier.


modules/default.nix
```diff
@@ -0,0 +1,13 @@
+{
+  bitcoind = ./bitcoind.nix;
+  clightning = ./clightning.nix;
+  default = ./default.nix;
+  electrs = ./electrs.nix;
+  lightning-charge = ./lightning-charge.nix;
+  liquid = ./liquid.nix;
+  nanopos = ./nanopos.nix;
+  nix = ./nix-bitcoin.nix;
```
jonasnick: s/nix/nix-bitcoin ?

default.nix
```diff
@@ -0,0 +1,15 @@
+{ pkgs ? import <nixpkgs> {} }:
+{
+  # The `lib`, `modules`, and `overlay` names are special
+  lib = import ./lib { inherit pkgs; }; # functions
+  modules = import ./modules; # NixOS modules
+  overlays = import ./overlays; # nixpkgs overlays
```
jonasnick: The NUR template says that lib, modules and overlays are optional. Since we don't seem to use them wouldn't it be better to remove them?

modules/default.nix
clefru: Fixed

default.nix
clefru: I removed lib and overlays. I would retain "modules" given that things inside of the modules/ dir are actual NixOS modules. Also that's where I think most of the value comes from, as packaging is one thing, but correctly wiring everything up is another :)

jonasnick:
ACK

PR #19, opened 2019-03-29
Use rust stableinstead of nightly, update electrs, specify electrum-rpc-addr

nixbitcoin:
@romanz has enabled rust stable upstream. This commit transitions electrs from rust nightly to rust stable 1.31. It also updates the electrs version to 0.5.0. I also specified the electrum-rpc-addr to make the hidden service at port 50001 more reliable.


nixbitcoin:
Tested on Debian 9 Stretch

PR #20, opened 2019-03-29
Give operator access to onion hostnames through new onion-manager module

jonasnick:


modules/nix-bitcoin.nix
```diff
@@ -42,6 +42,7 @@ in {
     ./liquid.nix
     ./spark-wallet.nix
     ./electrs.nix
+    ./onion-manager.nix
```
clefru: You missed the historic opportunity to call this onion-chef, onion-dispenser, or onion-dealer! :)

jonasnick: *almost missed ;P

PR #22, opened 2019-03-29
Add network-libvirtd.nix for libvirtd environments.

clefru:
Example deployment for libvirtd.

FYI: 4GB of ram doesn't seem to be enough for bitcoind to sync.

### DIFF
--- a/network-libvirtd.nix
+++ b/network-libvirtd.nix
@@ -1,0 +1,30 @@
+{
+  bitcoin-node =
+    { config, pkgs, ... }:
+    {
+      deployment.targetEnv = "libvirtd";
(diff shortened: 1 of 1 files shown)


jonasnick:
I can't test this right now but looks good. I'm a bit hesitant to clobber the root dir with example files. Right now we have network-vbox.nix mostly just for the tutorial. I think we should move the network.nix files in their own directory.

That bitcoind doesn't sync with 4GB ram is surprising. IIRC it worked on @nixbitcoin's apu2c4 with 4GB. Reducing the dbcache setting to the default 450MB should help, but is perhaps not enough.

PR #23, opened 2019-03-31
Update Nix to 2.2.1 and add sandbox fix

nixbitcoin:


README.md
```diff
@@ -324,14 +333,24 @@ On the machine you are deploying from:
...
 
+	Sandboxing is enabled by default on Nix <2.2. This may cause problems on your system. If you get an error similar to
+
+	```
+	error: cloning builder process: Operation not permitted
+	error: unable to start build process
+	/tmp/nix-binary-tarball-unpack.hqawN4uSPr/unpack/nix-2.2.1-x86_64-linux/install: unable to install Nix into your default profile
+	```
+
+	try disabling sandboxing by adding `sandbox = false` to `/etc/nix/nix.conf`.
```
jonasnick: I haven't tested nix 2.2 but apparently this happens when nix isn't installed as root because sandboxing needs root permissions therefore imo better to specify that this happens because nix isn't isntalled as root and that alternatively you can install as root or build as user (as recommended here https://nixos.org/nix/manual/#conf-sandbox).

nixbitcoin: Fixed in 36b37eb

jonasnick: only fixed in one of the two locations

nixbitcoin: Fixed in a7eb091. Will be more careful next time.

PR #24, opened 2019-04-02
Project directory fixups

jonasnick:
@clefru does non-broken.nix serve a purpose or can we delete it?


clefru:
It's a shortcut .nix file to build all packages marked as non-broken found in many other NUR repos. Looks like most people use it in conjunction with travis.
https://github.com/nix-community/nur-combined/blob/master/repos/yurrriq/.travis.yml

I personally will leave it in my NUR repo, as I found it handy to just mark packages broken. I'll probably try to setup cachix for my own repo and then see how useful this file really is.

jonasnick:
I see. It looks like the [template repo](https://github.com/nix-community/nur-packages-template) was updated 2 days ago and `non-broken.nix` was deleted and instead there's a `ci.nix` now. If you think that's useful I could add `ci.nix` - at least its name isn't as obnoxious.

jonasnick:
I'm going ahead with this as I can't evaluate what's going on in the NUR. Can revisit later.

PR #29, opened 2019-04-10
Remove profiles and replace with options to enable/disable each module separately in configuration.nix

nixbitcoin:
This PR seeks to give the user more granular control over which modules he wants to run on his nix-bitcoin node. Above each option the user will find documentation explaining what the module is and when to enable/disable it.

Tested on top of 1efe510 and f0b78ee

### DIFF
--- a/configuration.nix
+++ b/configuration.nix
@@ -11,9 +11,25 @@
     # This is not needed when deploying to a virtual box.
     #./hardware-configuration.nix
   ];
-  services.nix-bitcoin.enable = true;
-  # FIXME Install and use minimal or all modules
-  services.nix-bitcoin.modules = "all";
+  # FIXME Enable modules by uncommenting their respective line. Disable modules by commenting out their respective line.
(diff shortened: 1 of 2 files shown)


configuration.nix
```diff
@@ -11,9 +11,25 @@
     # This is not needed when deploying to a virtual box.
     #./hardware-configuration.nix
   ];
-  services.nix-bitcoin.enable = true;
-  # FIXME Install and use minimal or all modules
-  services.nix-bitcoin.modules = "all";
+  # FIXME Enable modules by uncommenting their respective line. Disable modules by commenting out their respective line.
+    # Enable this module to use the nix-bitcoin node configuration. Only disable this if you know what you are doing.
```
jonasnick: Nit: Can you remove the indentation of the whole block to be on the same level as FIXME? And would also be nice to have a fixed `textwidth` (i.e. width in number of chars). The rest of the text has width 78.

PR #30, opened 2019-04-10
Split up and enhance documentation

nixbitcoin:
This PR slims down README.md and moves its content to an install.md and a useage.md in a docs directory. Overall documentation is brought to a common format and enhanced. The documentation replaces profiles with the new model in https://github.com/jonasnick/nix-bitcoin/pull/29, so it only really makes sense in conjunction with that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,16 +1,16 @@
 nix-bitcoin
 ===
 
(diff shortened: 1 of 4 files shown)


jonasnick:
*usage ;P

PR #41, opened 2019-04-19
Update usage.md with recurring-donations instructions

nixbitcoin:


nixbitcoin:
https://github.com/jonasnick/nix-bitcoin/issues/35

jonasnick:
I think ideally a new entry in usage.md wouldn't be needed. What do you think about moving item 2. and 4. to configuration.nix?

nixbitcoin:
Implemented in 4364e7c. I agree for at least as long as recurring-donations is experimental. What is your long-term vision for `docs/usage.md`?

jonasnick:
looks good to me. `docs/usage.md` should only contain explanations for things which have no place in configuration.nix.

PR #42, opened 2019-04-23
Switch to multi-user mode in nix installation instructions

jonasnick:


docs/install.md
```diff
@@ -55,24 +55,24 @@ The following steps are meant to be run on the machine you deploy from, not the
 	sudo apt-get install curl git gnupg2 dirmngr
 	```
 
-2. Install Latest Nix with GPG Verification
+2. Install latest Nix in "multi-user mode" with GPG Verification
```
nixbitcoin: Don't forget to make these changes in both tutorials.

jonasnick: Good catch, fixed.

PR #43, opened 2019-04-23
Make recurring donations persistent (across reboots)

jonasnick:


jonasnick:
Actually this is not necessary

PR #45, opened 2019-04-26
electrs ssl

nixbitcoin:
This PR modifies electrs to use a self-signed ssl certificate to provide additional connection security. This is vital for the Electrum Android App and should align connecting to electrs with standard Electrum usage.

Relevant Issue: https://github.com/jonasnick/nix-bitcoin/issues/33

### DIFF
--- a/modules/electrs.nix
+++ b/modules/electrs.nix
@@ -32,6 +32,16 @@ in {
         default = 50001;
         description = "Override the default port on which to listen for connections.";
     };
+    onionport = mkOption {
+        type = types.ints.u16;
+        default = 50002;
+        description = "Override the default port on which to listen for connections.";
+    };
+    nginxport = mkOption {
+        type = types.ints.u16;
+        default = 50003;
+        description = "Override the default port on which to listen for connections.";
+    };
   };
 
   config = mkIf cfg.enable {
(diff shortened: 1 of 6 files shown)


nixbitcoin:
Tested on top of https://github.com/jonasnick/nix-bitcoin/pull/44

modules/electrs.nix
```diff
@@ -48,7 +58,7 @@ in {
     systemd.services.electrs = {
       description = "Run electrs";
       wantedBy = [ "multi-user.target" ];
-      requires = [ "bitcoind.service" ];
+      requires = [ "bitcoind.service" "nginx.service"];
```
jonasnick: I don't think `requires` is necessary here. electrs works fine without nginx.

nixbitcoin: I see what you mean. The electrs systemd service doesn't need nginx, but SSL termination, which is essential to accessing electrs, only works if nginx is enabled.

jonasnick: mmh, only for android afaik, but okay

nixbitcoin: The way it's setup the Tor Hidden Service will direct traffic to the nginx reverse proxy which will only pass it on to electrs if the traffic uses SSL. So if nginx is disabled, electrs is unreachable, whether the user is on android or any other OS.

jonasnick: ah, got it

PR #68, opened 2019-05-17
Fix package dependencies

jonasnick:
Without this PR nix-bitcoin installs multiple versions of bitcoind and one of them is with GUI. To get rid of GUI bitcoin with this PR, run `nixops deploy`, reboot and run `nix-collect-garbage -d` on your target.

### DIFF
--- a/modules/clightning.nix
+++ b/modules/clightning.nix
@@ -73,7 +73,7 @@ in {
 
     systemd.services.clightning = {
       description = "Run clightningd";
-      path  = [ pkgs.bitcoin ];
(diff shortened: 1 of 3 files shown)


jonasnick:
>  Is there a way to check if miniupnpc is really disabled?

Not easily. But we could look into the symbols linked into the binary. If there's no miniupnp in `/nix/store` it's highly unlikely that it would come from somewhere else.

> Should I document nix-collect-garbage -d in the usage.md for future updates like this?

Nix garbage collection only removes things from the nix store. But this should be done on a regular basis to free disk space.

PR #70, opened 2019-05-18
Mention disabling Hyper-Threading and mention ECC

jonasnick:


README.md
```diff
@@ -67,7 +67,7 @@ Hardware requirements
 ---
 * Disk space: 300 GB (235GB for Bitcoin blockchain + some room)
   * Bitcoin Core pruning is not supported at the moment because it's not supported by c-lightning. It's possible to use pruning but you need to know what you're doing.
-* RAM: 2GB of memory
+* RAM: 2GB of memory. ECC memory is better.
```
nixbitcoin: Can you explain why ECC memory is better?

docs/install.md
```diff
@@ -125,6 +125,13 @@ See [usage.md](usage.md) for usage instructions, such as how to update.
 
 Tutorial: install and configure NixOS for nix-bitcoin on your own hardware
 ---
+
+## 0. Preparation
+
+1. Optional: Disable Simultaneous Multi-Threading (SMT) in the BIOS
+
+    Researchers recommend disabling (SMT), also known as Hyper-Threading Technology in the Intel® world to significantly reduce the impact of speculative exacution-based attacks (https://mdsattacks.com/).
```
nixbitcoin: Another mitigation would be installing intel microcode updates, possibly through: https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/hardware/cpu/intel-microcode.nix

README.md
jonasnick: To avoid bitflips resulting in weak private keys or nonces. And to make rowhammer (slightly) harder.

jonasnick: Unfortunately we can't make a footnote for that in the README.

jonasnick:
Added commit to mention firmware updates

README.md
jonasnick: Perhaps a bit too hard to swallow in the README.

nixbitcoin: I agree. As long as it doesn't have any trade-offs users should be aware of.

PR #73, opened 2019-05-27
Update pinned nixpkgs

jonasnick:


nixbitcoin:
Tested and works on my end. What updates does this change include? Why don't you use one of the latest signed commits? Security theater?

jonasnick:
> What updates does this change include?

Updates to the latest stable und unstable packages as part of regular update intervals.

> Why don't you use one of the latest signed commits?

Forgot this time, will do next time. Next truly signed commit (i.e. not signed by github) is about 10 commits below tip.

nixbitcoin:
Should we pick trusted NixOS maintainers whose signed commits we merge into our pinned nixpkgs? 

Possibly these could be maintainers who are verifiable through WoT and/or whose keys we can verify in person. However, I don't know how much security this provides since the assumption would unrealistically be that their repository is fully reviewed and backdoor free.

jonasnick:
Yeah that could be done, but we'd need consistent commits from them to not miss important updates for too long.

Issue #57, opened 2019-04-28
Harden Nginx and Tor


nixbitcoin:
**Time Horizon:** Short term

**Description:** Apply `nix-bitcoin-services.defaultHardening` to both network facing services Nginx and Tor. Additionally make Nginx run as unprivileged user `nginx` instead of `root`.

**Motivation:** Continuation of nix-bitcoin hardening efforts.

**Implementation:**

nixbitcoin:
@jonasnick Should we port [this](https://github.com/NixOS/nixpkgs/pull/60646) to nix-bitcoin?

PR #74, opened 2019-05-30
nginx with defaultHardening measures and not running as root

nixbitcoin:
Based on NixOS 19.03 services/modules/services/web-servers/nginx.nix, https://github.com/NixOS/nixpkgs/pull/60646, and our own defaultHardening.

IMO this Closes https://github.com/fort-nix/nix-bitcoin/issues/57 because it hardens nginx and removes root. Regarding Tor, it already has the following systemd hardening measures:
```
PrivateTmp = mkIf (!cfg.controlSocket.enable) "yes";
PrivateDevices = "yes";
ProtectHome = "yes";
ProtectSystem = "strict";
InaccessiblePaths = "/home";
ReadOnlyPaths = "/";
ReadWritePaths = [ torDirectory torRunDirectory ];
NoNewPrivileges = "yes";
```

### DIFF
--- a/modules/nginx.nix
+++ b/modules/nginx.nix
@@ -1,0 +1,709 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  nix-bitcoin-services = pkgs.callPackage ./nix-bitcoin-services.nix { };
+  cfg = config.services.nginx;
+  virtualHosts = mapAttrs (vhostName: vhostConfig:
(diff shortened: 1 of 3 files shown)


nixbitcoin:
Closing PR, because of lack of interest, increased maintenance work resulting from running our own nginx module, and nginx being well-vetted enough to reasonably trust without additional hardening.

Issue #75, opened 2019-05-30
WoT verify pinned nixpkgs


nixbitcoin:
**Time Horizon:** Short term

**Description:** Update both nixpkgs and nixpkgs-unstable only to signed commits by trusted maintainers whose keys have been verified through the WoT or in person.

**Motivation:** Continuation of nix-bitcoin hardening efforts.

**Implementation:** Find trustworthy maintainers who regularly commit signed commits to either nixpkgs or nixpkgs-unstable, verify their PGP keys through WoT or in person, write standard to only pin those commits, follow the standard.

nixbitcoin:
This has turned out to be security theater. Maybe there will be an actual solution someday like [flake authentication](https://github.com/NixOS/nix/issues/2849)

PR #78, opened 2019-06-04
WoT verify pinned nixpkgs

nixbitcoin:
This PR sets up criteria and procedure for verifying pinned nixpkgs and nixpkgs-unstable commits. Additionally it provides a list of nixpkgs maintainer keys that fully or partially meet the criteria. The standards described could be automated in a script for easy verification.

It is worth discussing how much security this additional effort actually provides.

Closes https://github.com/fort-nix/nix-bitcoin/issues/75

### DIFF
--- a/contrib/pinning-nixpkgs.md
+++ b/contrib/pinning-nixpkgs.md
@@ -1,0 +1,57 @@
+# Verifying pinned nixpkgs and nixpkgs-unstable commits
+
+Goal
+---
+All pinned nixpkg commits should be signed by a trusted nixpkgs maintainer with a verified GPG key.
+
+Terms
+---
+**Trusted:** An individual who has no history of security breaches, is in good standing with the Nix community, has been a NixOS maintainer for at least a year, and is not pseudonymous.
+
+**Nixpkgs maintainer:** An individual who is included in the [nixpkgs maintainers list](https://github.com/NixOS/nixpkgs/blob/master/maintainers/maintainer-list.nix).
+
+**GPG key:** A rsa>2048, ed25519, or 256e GPG/PGP key. 
+
+**Verified:** Verified through _at least_ two of the following (in descending preference)
+* **WoT:** GPG key has at least three marginally trusted signatures or one fully trusted signature. 
+* **Marginally trusted signatures:** Signatures made by keys trusted by at least one nix-bitcoin developer.
+* **Fully trusted signatures:** Signatures made by keys of nix-bitcoin developers.
+* **Keybase:** GPG key is listed in a Keybase profile that includes the Nixpkgs maintainer's GitHub. 
+* **Website:** GPG key fingerprint is displayed or GPG key is downloadble on a nixpkgs maintainer's personal website.
+
+**nix-bitcoin developer:** jonasnick B1A70E4F8DCD0366, nix-bitcoin D11F9AD5308B3BA
+
+Verified List
+---
+Nixpkgs maintainer | GPG key | Verifiable through 
+--- | --- | ---
+[Elis Hirwing](https://github.com/etu) | 0xD57EFA625C9A925F | [WoT](https://pgp.cs.uu.nl/paths/b1a70e4f8dcd0366/to/d57efa625c9a925f.html), [Website](https://elis.nu/)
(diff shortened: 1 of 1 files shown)


contrib/pinning-nixpkgs.md
```diff
@@ -0,0 +1,57 @@
...
+[Michael Weiss](https://github.com/primeos) | 0x130826A6C2A389FD | [Keybase](https://keybase.io/primeos)
+[Alyssa Ross](https://github.com/alyssais) | 0x736CCDF9EF51BD97 | [WoT](https://pgp.cs.uu.nl/mk_path.cgi?FROM=b1a70e4f8dcd0366&TO=736CCDF9EF51BD97&PATHS=trust+paths) *not in strong set anymore*, [Website](https://twitter.com/qyliss)
+[Sébastien Maret](https://github.com/smaret) | 0x86E30E5A0F5FC59C | [WoT](https://pgp.cs.uu.nl/paths/b1a70e4f8dcd0366/to/86e30e5a0f5fc59c.html)
+[Sondre Niles](https://github.com/sondr3) | 0x25676BCBFFAD76B1 | [Keybase](https://keybase.io/sondre)
+[Tadeo Kondrak](https://github.com/tadeokondrak) | 0xFBE607FCC49516D3 | [Keybase](https://keybase.io/tdeo)
+[Domen Kozar](https://github.com/domenkozar) | 0xC2FFBCAFD2C24246 | [Keybase](https://keybase.io/ielectric) 
+
+Procedure
+---
+1. Search for most recent signed commit on [nixos-19.03](https://github.com/NixOS/nixpkgs-channels/commits/nixos-19.03) or [nixos-unstable](https://github.com/NixOS/nixpkgs-channels/commits/nixos-unstable)
```
jonasnick: s/or/and

nixbitcoin:
> What if we desparately need a fix that's not yet buried in a signed commit?

We can always manually add a commit hash to nixpkgs-pinned.nix in emergencies. The standard could reflect this. In that case we would just not use the script and revert to manually updating.

> What if we miss an important fix by accident during the regular update because it's not yet buried in a signed commit?

Quite a few of the recent commits in https://github.com/NixOS/nixpkgs-channels/commits/nixos-19.03 and https://github.com/NixOS/nixpkgs-channels/commits/nixos-unstable are properly signed according to our specifications, which should include all the previously buried updates. The script could introduce a time limit on how recent the commit must be, and throw an error if the latest properly signed commit is too old.

> Before we merge this we need a script that does everything automatically because it's quite boring to search for a signed commit from the list. Perhaps this whole PR can be a script so we don't have gpg keys in both a markdown file and a script which makes maintainability harder? Ideally the script would recv the keys, clone both nixos-unstable and nixos-19.03 into a temporary dir and then produce a nixpkgs-pinned.nix file.

Yes, will do. What do you think about this script getting the fingerprints from a list of manually verified maintainer key fingerprints signed by one of our keys? This could be a way to cryptographically certify  keys with information from different sources, like Keybase, websites, and GPG WoT. Or would there be a more elegant way to verify fingerprints from Keybase proofs, WoT, and websites in a scripted way?

contrib/pinning-nixpkgs.md
nixbitcoin: ACK

fiksn:
Interesting approach, but I don't seem to exactly understand the threat model. Ok, supply chain attacks in a sense that somebody puts a backdoored bitcoind in nixpkgs. But why should a "commit-chain" that was extended by one of the trusted maintainers be better? A backdoor could be well hidden and the people could work on a completely unrelated features. What I want to say, if there is a backdoor you are screwed one way or the other. There should be a quorum of nix-bitcoin maintainers saying ok this stuff from nixpkgs seems sane.

nixbitcoin:
Threat model: GitHub slips a malicious update into the tree, attacker gets write access to the git repo and merges backdoored code without review

At least when a trusted maintainer has signed the top commit we know that the state exists locally somewhere with somebody who probably looks at the changes to his nixpkgs very carefully.

I suggest I implement this first and when we have more development resources we set up a quorum to review relevant nixpkgs.

nixbitcoin:
#162 is the PR to be merged

PR #79, opened 2019-06-11
Update to spark 0.2.8 and fix rate lookup

### LINKED ISSUE #62: Fix spark-wallet price lookup
Spark-wallet with `nix-bitcoin-services.allowTor` doesn't allow looking up price. Using torsocks to wrap the price request gives an error. Should be fixed ASAP if we merge #52 before. 

Upstream issue: https://github.com/shesek/spark-wallet/issues/78

jonasnick:
Fixes #62 

I couldn't verify that the spark we're getting from the registry is the correct one due to https://github.com/shesek/spark-wallet/issues/95. But we're anyway relying on npm to ensure the integrity of all the other packages.


configuration.nix
```diff
@@ -34,7 +34,7 @@
   # Enable this module to use spark-wallet, a minimalistic wallet GUI for
   # c-lightning, accessible over the web or through mobile and desktop apps.
   # Only enable this if clightning is enabled.
-  # services.spark-wallet.enable = true;
+  services.spark-wallet.enable = true;
```
nixbitcoin: Why are we enabling spark-wallet by default?

modules/spark-wallet.nix
```diff
@@ -15,6 +15,9 @@ let
       CMD="$CMD --public-url http://$(cat /var/lib/onion-chef/clightning/spark-wallet)"
       ''
     }
+    # Use rate provide wasabit because default (bitstamp) doesn't accept
```
nixbitcoin: small typo `wasabit`

pkgs/spark-wallet/supplement.nix
```diff
@@ -1,4 +1,4 @@
-# This file has been generated by node2nix 1.6.1. Do not edit!
+# This file has been generated by node2nix 1.6.0. Do not edit!
```
nixbitcoin: Just out of curiosity why did we downgrades node2nix versions?

nixbitcoin:
I can now see the usd price again! Good work.

Why don't we pull `spark-wallet-^0.2.8` from the [signed Github releases](https://github.com/shesek/spark-wallet/releases/download/v0.2.8/spark-wallet-0.2.8-npm.tgz) in `node-packages.nix` page like @shesek recommends? How did you create the `node-packages.nix`? Running `./generate.sh` in `pkgs/spark-wallet` gives me a different file.

pkgs/spark-wallet/supplement.nix
jonasnick: 1.6.0 is the latest version in stable. The last generation was done by clefru - not sure where he got 1.6.1 from.

modules/spark-wallet.nix
jonasnick: fixed

configuration.nix
jonasnick: fixed

jonasnick:
> Why don't we pull spark-wallet-^0.2.8 from the signed Github releases in node-packages.nix page like @shesek recommends?

When clefru created the node2nix PR he changed that to using the npm registry because for some reason this fixed the qr code display which didn't work before.

> How did you create the node-packages.nix?

Are you on unstable? I ran generate.sh in the nix-bitcoin shell.

nixbitcoin:
I still can't reproduce node-packages.nix with `./generate.sh`, but I got spark-wallet working from @shesek's [signed GitHub release](https://github.com/shesek/spark-wallet/releases/download/v0.2.8/spark-wallet-0.2.8-npm.tgz). I have commited my results in [my own spark-0.2.8 branch](https://github.com/nixbitcoin/nix-bitcoin/tree/spark-0.2.8).

shesek:
> since npm distributed dependencies could be compromised.

Note that the `npm-shrinkwrap.json` file inside the spark npm package commits to the sha512 integrity checksums of the entire dependency tree. You're guaranteed to get the exact same dependencies that I had when I signed the package.

PR #82, opened 2019-07-01
Resize VirtualBox disk to 300gb

ldub:
I've put together a script to help fix #50. I'm new to NixOps and I'm not sure how to make NixOps run these commands for me. If can give me a quick pointer for how this is supposed to be done, I'll update this PR.

Example run:
```
$ ./scripts/vbox-resize-disk1.sh
++ DEPLOYMENT=bitcoin-node
++ MACHINE=bitcoin-node
+++ nixops export -d bitcoin-node
+++ nix-shell -p jq --command 'jq -r '\''..|."virtualbox.disks"?|select(.!=null)'\'' | jq -r .disk1.path'
++ DISK_FILE='/Users/lev/VirtualBox VMs/nixops-81b16311-9af4-11e9-abb6-d0817ad44941-bitcoin-node/disk1.vdi'
++ nixops stop -d bitcoin-node
bitcoin-node> shutting down... [running]
bitcoin-node> [running] Connection to 192.168.56.115 closed by remote host.

bitcoin-node> [running]
bitcoin-node> [running]
bitcoin-node> [running]
bitcoin-node> [running]
bitcoin-node> [running]
bitcoin-node> [running]
bitcoin-node> [running]
bitcoin-node> [running]
bitcoin-node> [poweroff] ++ VBoxManage modifyhd --resize 266240 '/Users/lev/VirtualBox VMs/nixops-81b16311-9af4-11e9-abb6-d0817ad44941-bitcoin-node/disk1.vdi'
0%...10%...20%...30%...40%...50%...60%...70%...80%...90%...100%
++ nixops start -d bitcoin-node
bitcoin-node> starting...
bitcoin-node> Waiting for VM "nixops-81b16311-9af4-11e9-abb6-d0817ad44941-bitcoin-node" to power on...
bitcoin-node> VM "nixops-81b16311-9af4-11e9-abb6-d0817ad44941-bitcoin-node" has been successfully started.
bitcoin-node> waiting for IP address... 192.168.56.115
bitcoin-node> waiting for SSH........................................................................
++ nixops ssh -d bitcoin-node bitcoin-node -- '(echo d; echo n; echo p; echo; echo; echo; echo w; ) | fdisk /dev/sda'

set -ex
Welcome to fdisk (util-linux 2.33.1).
Changes will remain in memory only, until you decide to write them.
set -ex
Be careful before using the write command.


Command (m for help): Selected partition 1
Partition 1 has been deleted.

Command (m for help): Partition type
[...]

PR #83, opened 2019-07-03
macOS installation instructions

ldub:
Added the instructions to run nix-bitcoin with virtualbox on a macOS host.

I'll rebase this on top of master when #82 is merged.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -6,7 +6,8 @@ Have a look at the options in the [NixOps manual](https://nixos.org/nixops/manua
 
 # Tutorials
 1. [Install and configure NixOS for nix-bitcoin on VirtualBox](#tutorial-install-and-configure-nixos-for-nix-bitcoin-on-virtualbox)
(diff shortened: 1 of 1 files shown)


docs/install.md
```diff
@@ -123,6 +124,104 @@ You can also build Nix from source by following the instructions at https://nixo
...
+
+2. Confirm that nix-linuxkit works
+
+    ```
+	nix-build /Users/lev/.cache/nix-linuxkit-builder/example.nix
+	```
+
+	As the installer says, run a `nix-build` to make sure that you are able to build linux binaries. The `example.nix` is specifically configured to force a x86_64-linux build. Remove the generated `result` folder afterwards.
+
+## 4. Nixops deployment
```
jonasnick: In general we should avoid repeating the same instructions in this document if possible. How about just referring to the Nixops deployment section in the other tutorial (with the exception of running deploy with the `--option system x86_64-linux` flag)?

ldub:
Added references from macOS install section to debian install section. 
Also fixed a typo and added some separators.

PR #87, opened 2019-07-12
Always chown bitcoin/liquid data directories

jonasnick:
This allows using an existing datadir, for which the user may not have permissions, f.e. an external drive. CC @stevenroose

### DIFF
--- a/modules/bitcoind.nix
+++ b/modules/bitcoind.nix
@@ -207,15 +207,13 @@ in {
       preStart = ''
         if ! test -e ${cfg.dataDir}; then
           mkdir -m 0770 -p '${cfg.dataDir}'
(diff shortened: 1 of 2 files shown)


stevenroose:
ACK f707d970ae28c76420761575e3e0d428be0f9a37

stevenroose:
@nixbitcoin Yeah. I mounted it in the nix hardware-configuration file thing and I precreated the directories. I actually had an old Core and electrs directory and wanted to see if they would work still. But they were owned by root so it didn't work.

nixbitcoin:
Did the electrs preStart, specifically `chown -R '${cfg.user}:${cfg.group}' ${cfg.dataDir}` succeed in making the electrs directory work from the mounted external harddrive?

PR #88, opened 2019-07-12
add BTCPayServer

jb55:
This adds BTCPayServer to nix-bitcoin! This is just the package for now, no module yet.

I also still need to do package update instructions for the `nuget-package.json`. I'm using my [dotnet2nix](https://github.com/jb55/dotnet2nix) project to generate them

Please test and let me know if you run into any issues. This uses a new dotnetcore build support which isn't in nixpkgs yet.


pkgs/nbxplorer/default.nix
```diff
@@ -0,0 +1,28 @@
+{ stdenv, lib, bash, dotnet, callPackage, writeText, makeWrapper,
```
jb55: I realize I have extra arguments here I forgot to remove after I refactored everything. Will remove these

pkgs/btcpayserver/default.nix
```diff
@@ -0,0 +1,27 @@
+{ stdenv, dotnet, lib, bash, callPackage, writeText, makeWrapper, writeScript, dotnet-sdk,
```
jb55: I need to fix the extra arguments here as well

kcalvinalvin:
Can't tell which deps are needed so can't test right now. How long does the build take?

I have working packages for NBXplorer and btcpayserver at https://github.com/kcalvinalvin/calvin-nix-conf

Only downside at the moment to my build is that it takes long.

jb55:
it looks like the slowest part of your build is simply adding packages to the package set: https://github.com/kcalvinalvin/calvin-nix-conf/blob/aab817b4b526e7ec8c43dd814575d3ac62726b47/pkgs/btcpayserver/default.nix#L33

I get around this by generating the package set myself. I noticed you are parsing the dotnet install log. There's a `project.assets.json` that gets generated in the obj dir. This is what my [dotnet2nix](https://github.com/jb55/dotnet2nix) project converts to json/nix. your bash script is definitely simpler though. I don't think I could use that approach though, I need the sha512 in the project.assets.json for building the nuget package set.

one issue I had is that the compiled BTCPayServer binary that is generated requires additional environment variables set before running. So I use wrapProgram, but then the binary complains that the argv[0] doesn't match the dll name. I'll try stealing your dotnet exec approach...

jb55:
To answer your question, this install takes about 1 minute and 36 seconds, which is mostly just to download the dependencies, as this riveting asciinema demonstrates: https://asciinema.org/a/GPGIT5f4Swzdb3P2uItVIZJRC

> Can't tell which deps are needed so can't test right now. 

not sure what you mean here, as long as you do the same as the asciinema linked above it should work.

Another issue that crops up is the mismatch between generated dependencies and runtime versions. dotnet will complain. This can happen with different nixpkgs versions. I think there should be a way to make it not complain so I'm going to look into that today.

kcalvinalvin:
> install takes about 1 minute and 36 seconds

Ah ok, much faster than what I have.

> not sure what you mean here, as long as you do the same as the asciinema linked above it should work.

I was trying to just take the package and build it on my NixOS config, I can give it a try on my rasp pi by taking the code whole.

>Another issue that crops up is the mismatch between generated dependencies and runtime versions. dotnet will complain. This can happen with different nixpkgs versions. I think there should be a way to make it not complain so I'm going to look into that today.

Could you post the log? Not sure if I ran into that issue as well. Can take a looksie.

jb55:
> I was trying to just take the package and build it on my NixOS config, I can give it a try on my rasp pi by taking the code whole.

it might be easiest to add a nix-bitcoin reference in your nixpkgs overrides:

```nix
{ pkgs }:
let nix-bitcoin = import (pkgs.fetchFromGitHub {
      owner = "fort-nix";
      repo  = "nix-bitcoin";
      rev   = "030c50f36448160afa2700f2260a9254cd4fd734";
      sha256 = "0qds3hw88nq0qka3nx8q9p39ibw6jq1n8z5fv7gik3zksgdbhpzf";
    }) { inherit pkgs; };
in {
  packageOverrides = super: {
    inherit (nix-bitcoin) btcpayserver nbxplorer;
  };
}
```

then you can just do `nix build nixpkgs.{btcpayserver,nbxplorer}`

with the added bonus of being able to reference these packages in local systemd services.

jb55:
I got the build working for 19.03, but now it breaks on nixos-unstable. It seems the way I'm doing it only works when targeting specific versions of the dotnet runtime. I didn't have any more time today to figure out how to make this more flexible.

jb55:
scratch that, I have a workaround where I generate different nuget package sets for different versions of the dotnetcore sdk. seems to be working, new rebase incoming...

jb55:
OK so I think we're all good now. Latest changes fixed a bunch of things:

* NBXplorer/BTCPayServer now work on nixos-19.03 and nixos-unstable. It selects the proper dependencies depending on the corresponding dotnet-sdk version.

* The binary names have been fixed from `BTCPayServer-Release` to `BTCPayServer`, etc. This uses the `dotnet exec` trick that @kcalvinalvin used in his version.

Please test and let me know how it goes!

ldub:
Sweet. Build works fine on my machine now.

jonasnick:
Also builds fine in our nix-shell, i.e. with the pinned stable nix packages. Afaics the only things missing is a `generate.sh` to create the `nuget-packages.json` similar to `pkgs/spark-wallet/generate.sh` and removing the extra arguments as you mentioned above.

jb55:
yup will do that next, thanks for testing in the meantime.

nixbitcoin:
How do you imagine BTCPayServer usage being integrated into nix-bitcoin? We could automatically integrate with the user's nix-bitcoin clightning node and let the user specify an xpub for on-chain payments in configuration.nix. Ideally the user would only need to change two variables in configuration.nix "services.btcpayserver.enable & services.btcpayserver.xpub" to get it up and running.

jb55:
@nixbitcoin I'm pretty new to btcpayserver, but I noticed you could just set the xpub from the interface? I imagine most people would set up the instance manually, because it includes ways for checking that your xpub is correct. So a minimal nix config should be ok for now.

I'll have to think about clightning integration, I haven't tested that yet. It worked for me out of the box in the impure setting, but I'm not sure if it *just works* in sandboxed builds.

jonasnick:
Hey @jb55, I'd really like to add BTCPayServer in the coming weeks. Are you still interested in continuing with this?

jb55:
@jonasnick yeah I just need to find time to put together the update instructions...

kcalvinalvin:
> @jonasnick yeah I just need to find time to put together the update instructions...

What do you need help on specifically?

jb55:
@kcalvinalvin the thing I'm having trouble with is syncing the dependencies with the runtime version of dotnet-sdk in nixpkgs. even things aren't aligned perfectly the compile fails. need some way to make updates reliable/automatable, right now it takes some finagling since I'm not sure exactly how the build time/runtime dependencies work.

right now I have to run `dotnet public -c Release -r linux-x64` to generate the nuget dependencies. these dependencies are determined by the current installed dotnet-sdk.

then I type `dotnet2nix` to generate a json of these dependencies which are tied to the current dotnet-sdk version. The build will fail/complain for various reasons related to this that I have yet to figure out.

kcalvinalvin:
Is this still being worked on?

jb55:
it looks like the packaging process is even more complicated now since btcpayserver uses custom nuget repos.

jb55:
to support this I would have to update `dotnet2nix` to load the nuget.config and fallback to the additional repos if there's a 404 when hitting the nuget api.

jb55:
dropping this approach as it seems too difficult to maintain and I don't have time to maintain it

Issue #86, opened 2019-07-10
electrs breaks when the datadir variable is set


stevenroose:
I set the `electrs.dataDir` variable to some other location and I'm getting these logs in electrs:
```
Jul 10 08:55:55 dasilva-nix s4j86k2qkxz8473v1gy7xh3q5jfxp743-unit-script-electrs-pre-start[1425]: chmod: cannot access '/media/bitcoin-extra/electrs/startscript.sh': No such file or directory
```

So I suppose this startscript.sh script is somehow hard-coded into a path, but then the datadir is used to invoke it. Why isn't that script really hardcoded and just put in some hardcoded location?

stevenroose:
So I guess it would either need a change like this:
```diff
diff --git a/modules/electrs.nix b/modules/electrs.nix
index 524bbd1..657c4c9 100644
--- a/modules/electrs.nix
+++ b/modules/electrs.nix
@@ -66,7 +66,7 @@ in {
       preStart = ''
         mkdir -m 0770 -p ${cfg.dataDir}
         chown 'electrs:electrs' ${cfg.dataDir}
-        echo "${pkgs.electrs}/bin/electrs -vvv ${index-batch-size} ${jsonrpc-import} --timestamp --db-dir ${cfg.dataDir} --daemon-dir /var/lib/bitcoind --cookie=${config.services.bitcoind.rpcuser}:$(cat /secrets/bitcoin-rpcpassword) --electrum-rpc-addr=127.0.0.1:${toString cfg.port}" > /var/lib/electrs/startscript.sh
+        echo "${pkgs.electrs}/bin/electrs -vvv ${index-batch-size} ${jsonrpc-import} --timestamp --db-dir ${cfg.dataDir} --daemon-dir /var/lib/bitcoind --cookie=${config.services.bitcoind.rpcuser}:$(cat /secrets/bitcoin-rpcpassword) --electrum-rpc-addr=127.0.0.1:${toString cfg.port}" > ${cfg.dataDir}/startscript.sh
         chown -R 'electrs:electrs' ${cfg.dataDir}
         chmod u+x ${cfg.dataDir}/startscript.sh
         '';    
```

Or like this:
```diff
diff --git a/modules/electrs.nix b/modules/electrs.nix
index 524bbd1..04e3ade 100644
--- a/modules/electrs.nix
+++ b/modules/electrs.nix
@@ -68,11 +68,11 @@ in {
[...]

PR #93, opened 2019-07-19
Eliminate startscript.sh in electrs module

stevenroose:
Fixes https://github.com/fort-nix/nix-bitcoin/issues/86.

I'm new to Nix development, so I have no idea how I can test these things easily..
I'm also not super confident in the correctness, so please review thoroughly :)

### DIFF
--- a/modules/electrs.nix
+++ b/modules/electrs.nix
@@ -66,13 +66,10 @@ in {
       preStart = ''
         mkdir -m 0770 -p ${cfg.dataDir}
         chown 'electrs:electrs' ${cfg.dataDir}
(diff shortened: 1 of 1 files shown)


jonasnick:
I test PRs like that by deploying to virtualbox as explained in the docs with `electrs` enabled and check its logs.

stevenroose:
Hmm, it seems like the `$()` notation doesn't work inside `ExecStart`... Does it make sense to do this PR then?

jonasnick:
That's probably the reason why I made a startscript in the first place. Sorry for not remembering earlier.

nixbitcoin:
> Hmm, it seems like the $() notation doesn't work inside ExecStart... Does it make sense to do this PR then?

Yes, that seems to be the case. I'm getting the same error. I've created another PR #96 that fixes the problem by placing the electrs startscript in a temporary dir.

PR #95, opened 2019-07-24
Add travis ci builds

jb55:
This builds everything to make sure everything works on either the latest release or unstable branch. In the future we could do more interesting nixops builds to test modules, but for now we just nix-build everything and make sure it compiles.

We test the build with two different versions of nixpkgs:

  - nixpkgs-channels/nixos-19.03
  - nixpkgs-channels/nixos-unstable

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,8 +30,6 @@ env:
   - PKG=elementsd STABLE=1
   - PKG=elementsd STABLE=0
   - PKG=electrs STABLE=1
-  global:
-    secure: "bC32LNscv/BXG5fxQuO2laB4cU5lB1bMeYymZk5c/ZFwXMeK5L4DBmU8VXQANTsZA9bJcqWAcwWVu11jxan3n8lCNsAB/6JcRSluPzOkpwewldurwEMT8OUO2GP85zVjPmGEO8tyZL+yTWZr+3dRZgi7zi+rYTtDx8uhg7qH/g8zsZMLCHYi2eNQpiIFl8VnzX1gIGN83FSBMq9HdWinOD+Te+cpGcKnXcnd0LjMmFD3dD1d4Bs4TL1j36owRLPugD0lZbi9W6NipIaQShc30f5+34DoUWV8/Q9UxFvbPsmnDJ4hY7RRssQP1DbanzEAnjiB//PLmzNzAHJrw4eS3AScnTaZD7Tly/0VqQG7tXk1kheSvrmWkrQ/jeF0R5v+YW/e563sPs+J4QFxteCIUP5EPKohsmK4Jw6kB9rBH/RK4RDRCdhsneum4HTDJNXBFZH6jpa0B2L1nFZ9NmLgeO1dcPlSrVKIMvcZ5wGXsLjyalo4yrUSsGuT2a+96cKtfj936H70GPjGnzZq2QwbxJ6UOd02JPEEHHhe0WrKGut9qk+Rb3hPcgvrkhVLvHGY0N98DzpuR86UTWsRm7trTD69H49sjToVbx/ylOAvhkdK3Jv3+EpYNdkKIX6AYdIV32CTkk/Zmrig0ov2mQsnbWY5BB6ycSBAfIJu0/enhR8=" # nix-bitcoin cachix signing key
 script: |
   printf '%s (%s)\n' "$NIX_PATH" "$VER"
-  nix-build -A $PKG | cachix push nix-bitcoin
+  nix-build -A $PKG
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,35 @@
+language: minimal
+
+# Retry installing nix due to nondeterministic error
+#   Fatal error: glibc detected an invalid stdio handle
+# see:
+#   https://github.com/nh2/static-haskell-nix/pull/27#issuecomment-502652181
+#   https://github.com/nixos/nix/issues/2733
+install: |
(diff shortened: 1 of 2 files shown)


jb55:
uhh it looks like the build hits a timelimit and always fails, we might have to split the build into a matrix for each package.

jb55:
Timeouts are fixed as far as I can tell. I split individual packages into their own jobs in the build matrix, for both unstable and stable.

You can see the current build progress for this PR here: https://travis-ci.org/jb55/nix-bitcoin/builds/563236262

.travis.yml
```diff
@@ -0,0 +1,29 @@
+language: nix
+nix: 2.2.2
+
+env:
+  global:
+    - STABLE: "NIX_PATH=nixpkgs=https://github.com/nixos/nixpkgs-channels/archive/nixos-19.03.tar.gz"
+    - UNSTABLE: "nixpkgs=https://github.com/nixos/nixpkgs-channels/archive/nixos-unstable.tar.gz"
```
jonasnick: How about we pin the nixpkgs to the same versions as in `pkgs/nixpkgs-pinned.nix`? This would avoid having to figure out if an error is related to an update of the nixpkgs or if it's actually a problem with the package. Also presumably most users use the pinned versions. Then when we update the pinned dependencies and travis produces an error we would easily notice that the error is related to the update of the nixpkgs and nothing else.

jb55: Yeah that makes sense, will update

jb55:
I've pushed a few changes, it now:

- Tries to install nix 5 times in a row incase there's a random segfault
- Tests against the pinned nixpkgs versions

spark-wallet looks legit broken, not sure if this is a know issue

jb55:
Here's the build for this version: https://travis-ci.org/jb55/nix-bitcoin/builds/563345171

jb55:
it looks like the electrs build fails because the rocksdb dependency build takes more than 10 minutes without any output...

jonasnick:
Hm, I wonder if it's somehow using the wrong version. For both `STABLE` and `UNSTABLE` 
```
nix-instantiate --eval -E 'with import <nixpkgs> {}; lib.version or lib.nixpkgsVersion'
```
prints `"19.09pre186559.2bcc9260e4c"`, but on my machine it's `"19.03pre-git"` for stable and `"19.09pre-git"`. Additionally, on my machine I need to run 
```
nix eval -f pkgs/nixpkgs-pinned.nix 
```
with `--raw` because otherwise I'm getting
```
warning: Nix search path entry '"/nix/store/p6icmxdc5i6x8xz5fq17hzmxxjd80232-source"' does not exist, ignoring
```
meaning that it's confused by the quotes. So perhaps in the travis run `$STABLE/$UNSTABLE` and therefore `$NIX_PATH` are empty?

jb55:
@jonasnick yeah I noticed that as well. will investigate and fix the next chance I get.

jb55:
ok I think I found a solution. new build is running here https://travis-ci.org/jb55/nix-bitcoin

I'm also using caching (via https://cachix.org/) now so packages are only rebuilt if necessary

.travis.yml
```diff
@@ -0,0 +1,35 @@
...
+  - PKG=nanopos STABLE=1
+  - PKG=nanopos STABLE=0
+  - PKG=spark-wallet STABLE=1
+  - PKG=spark-wallet STABLE=0
+  - PKG=elementsd STABLE=1
+  - PKG=elementsd STABLE=0
+  - PKG=electrs STABLE=1
+  - PKG=electrs STABLE=0
+  global:
+    secure: A/jsU3/hz6ARBt9NLn9G8jJqiMDFnjfWomkuVrfL6Y742Oc0j7kmTXNlVeuW++wsm5SbQHNYcJ3cPDwDILMPx1USzpBwFt0hLLYp1qzibRN7FF28Xds52T1jG1CdN0/ydejPnpXzDGqxSjm+ftY+LfJMAz5ybOgzqUyamUnXuAjlQJXCeMB37H2jELKZ7qnqQPoO9LBnvmLnG+i8uAYGzYOnh3la7ZVBdT7hE28uzsPwbotKNGqmmVpS/v3hXV9yP8JF3TQrI16/glHT+JCGPyxglv/IPIXGchlcD8oRhRYHrN9Dxw1ZE4jzbDGoSAjkDHygPM1RDAJDPsZReQUkq1NXuUY/yVSD4tVs+Boc+opaHT3XBhkPS2WctiP7oE2ddZIfXIG8fu7ThvAkwpi9UXcxkpuFobEIZkZ46/jOdNToWb2z42Lzpacpv95nnobyyP5s9JhuU01winRMLQK14YHfeFVgoIxkjhRkodxGFpy2Jole1KNfBPmoqytqsrcSFoASyQK7Vwpz6vF8ja+iRyFBr3a/V5Dcbx/hZ6vt2ZFdLGy7uPLT4wuYAHjj+LRJvUNnx0nxabC6P/JuTqMlaonTbhFYIaj8rwN9ldq6VgeualHVqFWG7rmcgs+ehTeA5d9/jhb2jzsLSFRwwycAqKpG7ch41PnY/JqviT47SyQ=
```
jonasnick: Can you add a comment explaining what this does? If I understand correctly this sets environment variables, but which ones exactly?

jb55: this is the key needed for signing and pushing build artifacts to `nix-bitcoin.cachix.org`. I'll add a comment.

jb55: I ended up just removing this because I couldn't get the env var working

PR #96, opened 2019-07-26
Move electrs startscript to tempdir and fix nits

### LINKED ISSUE #86: electrs breaks when the datadir variable is set

nixbitcoin:
This PR fixes #86 by placing startscript in a systemd RuntimeDirectory only accessible by the electrs user as per @jonasnick's suggestion. It also sets a cfg.user and cfg.group variable and removes the hardcoded electrs user and group in several places.

I'm a bit concerned that I can't pass `${RuntimeDirectory}` from `serviceConfig` to `preStart` and had to hardcode `/run/electrs/startscript.sh`, but I've seen this done in other nix modules upstream. Will this be an issue?

Closes #93

### DIFF
--- a/modules/electrs.nix
+++ b/modules/electrs.nix
@@ -21,6 +21,16 @@ in {
       default = "/var/lib/electrs";
       description = "The data directory for electrs.";
     };
+    user = mkOption {
+      type = types.str;
+      default = "electrs";
+      description = "The user as which to run electrs.";
+    };
+    group = mkOption {
+      type = types.str;
+      default = cfg.user;
+      description = "The group as which to run electrs.";
+    };
     high-memory = mkOption {
       type = types.bool;
       default = false;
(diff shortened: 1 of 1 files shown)


modules/electrs.nix
```diff
@@ -66,13 +77,14 @@ in {
       preStart = ''
         mkdir -m 0770 -p ${cfg.dataDir}
         chown 'electrs:electrs' ${cfg.dataDir}
```
jonasnick: Can you remove that? Seems even less useful now?

jonasnick: Perhaps also move the `chown -R '${cfg.user}:${cfg.group}' ${cfg.dataDir}` right below `mkdir`.

nixbitcoin:
Do you think `/run/electrs/startscript.sh` could cause problems?

jonasnick:
@nixbitcoin No. What kind of problems are you thinking of?

nixbitcoin:
> @nixbitcoin No. What kind of problems are you thinking of?

If somebody were to change RuntimeDirectory, startscript would no longer be located at `/run/electrs/startscript.sh`. However, I see no reason for a user doing this.

PR #97, opened 2019-07-26
Revert to old electrs pkg format for update

nixbitcoin:
Crates.io doesn't seem to have the newest electrs version 0.7.0, instead it only has 0.6.2. To update electrs I reverted to the old format without crate2nix, which we will change back when electrs gets updated upstream. Does this work for you @jb55 ?


jonasnick:
Probably easier to just ask romanz to run `cargo publish`?

romanz:
Will do tomorrow, thanks for letting me know :)

romanz:
> Will do tomorrow, thanks for letting me know :)

Done: https://crates.io/crates/electrs (also released a new patch version).

Issue #84, opened 2019-07-07
Enable validatepegin for Liquid


stevenroose:
Would be cool to have a bool flag in the liquid config that when true (and perhaps if bitcoind.enable is true), it adds the `validatepegin=1` and the necessary `mainchainrpc*` fields to have Liquid validate pegins.

nixbitcoin:
ACK. Will add when I have time.

PR #99, opened 2019-07-26
Enable validatepegin for Liquid

nixbitcoin:
Closes https://github.com/fort-nix/nix-bitcoin/issues/84

I've made `validatepegin` false-by-default in configuration.nix. Should it be an opt-out? Is setting the option in liquid.conf even necessary? Seems to be enabled by default according to https://github.com/ElementsProject/elements/blob/master/share/examples/liquid.conf "# Validate pegin claims. All functionaries must run this. (default: 1)" @stevenroose

### DIFF
--- a/modules/liquid.nix
+++ b/modules/liquid.nix
@@ -10,6 +10,7 @@ let
     ${optionalString cfg.testnet "testnet=1"}
     ${optionalString (cfg.dbCache != null) "dbcache=${toString cfg.dbCache}"}
     ${optionalString (cfg.prune != null) "prune=${toString cfg.prune}"}
+    ${optionalString (cfg.validatepegin != null) "validatepegin=${if cfg.validatepegin then "1" else "0"}"}
 
     # Connection options
     ${optionalString (cfg.port != null) "port=${toString cfg.port}"}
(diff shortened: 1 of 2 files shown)


jonasnick:
> I've made validatepegin false-by-default in configuration.nix

I think it should be true in nix-bitcoin.nix and null by default in liquid.nix (use liquid-daemon defaults) 

> Is setting the option in liquid.conf even necessary? 

Apparently no, it already validatespegin, which is why I had to add mainchainrpcpassword in the first place.

nixbitcoin:
I can't imagine a situation in which somebody would want to disable validatepegin, and this PR would be useful. Can you?

jonasnick:
If you don't have bitcoind then validatepegin=false is useful and that may be the case if you use the module outside of nix-bitcoin.

stevenroose:
I agree with @jonasnick. In a situation where you have a bitcoind available, it should be default, but to make the liquid module more user-friendly as a standalone module, I'd not force it by default always.

modules/liquid.nix
```diff
@@ -10,6 +10,7 @@ let
     ${optionalString cfg.testnet "testnet=1"}
     ${optionalString (cfg.dbCache != null) "dbcache=${toString cfg.dbCache}"}
     ${optionalString (cfg.prune != null) "prune=${toString cfg.prune}"}
+    ${optionalString (cfg.validatepegin != true) "validatepegin=0"}
```
stevenroose: We can't assume Liquid defaults, right? Either the `cfg.validatepegin` should a `true|false|unset` variable or it should always be added to the config.

jonasnick: Yes the line should be something like
```
    ${optionalString (cfg.validatepegin != null) "validatepegin=${if cfg.validatepegin then "1" else "0"}"}
```

PR #101, opened 2019-07-28
Update electrs to v0.7.1

nixbitcoin:
New and better version of https://github.com/fort-nix/nix-bitcoin/pull/97 now that @romanz updated the crate.

@jb55 can you test deploying electrs and verify electrs sha256.


jb55:
LGTM but I don't use electrs anymore or have an easy way to verify hash legitimacy.

PR #105, opened 2019-08-01
 Replace liquidd with elementsd package

jonasnick:
Based on #104

The liquid repo/binaries are deprecated and instead `elementsd` is used to access the liquid network (see https://github.com/Blockstream/liquid). I tried to change the liquid module as little as possible. In order to interact with liquid you have to use `elements-cli` instead of `liquid-cli`.


configuration.nix
```diff
@@ -46,7 +46,7 @@
   # Enable this module to use liquidd, a daemon for an inter-exchange
   # settlement network linking together cryptocurrency exchanges and
   # institutions around the world.
-  # services.liquidd.enable = true;
+  services.liquidd.enable = true;
```
jonasnick: Oops sorry, fixing...

modules/liquid.nix
```diff
@@ -189,17 +189,17 @@ in {
...
-        echo "mainchainrpcpassword=$(cat /secrets/bitcoin-rpcpassword)" >> '${cfg.dataDir}/liquid.conf'
+        echo "rpcpassword=$(cat /secrets/liquid-rpcpassword)" >> '${cfg.dataDir}/elements.conf'
+        echo "mainchainrpcpassword=$(cat /secrets/bitcoin-rpcpassword)" >> '${cfg.dataDir}/elements.conf'
       '';
       serviceConfig = {
         Type = "simple";
         User = "${cfg.user}";
         Group = "${cfg.group}";
-        ExecStart = "${pkgs.liquidd}/bin/liquidd ${cmdlineOptions}";
+        ExecStart = "${pkgs.elementsd}/bin/elementsd ${cmdlineOptions}";
```
nixbitcoin: Don't we need to specify `-chain=liquidv1` in the cmdlineOptions?

modules/liquid.nix
```diff
@@ -179,7 +179,7 @@ in {
   };
 
   config = mkIf cfg.enable {
-    environment.systemPackages = [ pkgs.liquidd ];
+    environment.systemPackages = [ pkgs.elementsd ];
     systemd.services.liquidd = {
       description = "Liquid daemon";
```
nixbitcoin: Elements daemon

nixbitcoin:
If users would want to use the liquid/elements module for more than just the liquid instance of the Elements platform we should bite the bullet now and rename everything to the elements naming scheme. It will cause a lot of confusion down the line otherwise. We could migrate users by adding an `mv /var/lib/liquidd /var/lib/elemementsd` to the preStart.

If the module is only designed with liquid usage in mind we could stick to minimal changes in the liquid module approach, but we should at least add documentation in configuration.nix or the usage.md about the elements-cli liquid-cli switch.

Positive tests
* Verified pkgs/elementsd/default.nix sha256 against `git archive --format=tar --prefix=elements-elements-0.17.0.1/ elements-0.17.0.1 | gzip -n > elements-0.17.0.1.tar.gz` of gpg signed git tag
  * 0x6BE2CED14A9917BC key verified through WoT
* Checked for liquid.conf remnants
* Checked for liquidd binary remnants
* Tested with `nixops deploy` and `elements-cli -getinfo`

Nit
* Change liquidd mentions in options to elementsd
* Change Liquid daemon mentions in descriptions

jonasnick:
> users would want to use the liquid/elements module for more than just the liquid instance

I don't think so in the near to medium future.

> we should at least add documentation in configuration.nix or the usage.md about the elements-cli liquid-cli switch.

Added note to configuration.nix

> Change liquidd mentions in options to elementsd

Liquidd is not the worst name for the service because it refers to the daemon supporting Liquid. Just "liquid" would be confusing because liquid is a sidechain and not something you can run in a service. Calling it elementsd would be worse, especially if there were multiple chains for elements. Each should have their own service.

modules/liquid.nix
jonasnick: fixed

modules/liquid.nix
jonasnick: fixed

nixbitcoin:
As I see it the relationship is:
1. Liquid: an elements platform sidechain for an inter-exchange settlement network
2. Elements daemon: a daemon providing access to an elements platform sidechain
3. liquidd: an elements daemon providing access to the Liquid elements platform sidechain
4. elements-cli: a tool providing access to elements daemons
5. liquidd user: User of liquidd, an elements daemon providing access to Liquid, a sidechain for an inter-exchange settlement network

I've put the corresponding changes into https://github.com/fort-nix/nix-bitcoin/pull/108

jonasnick:
Fixed the `to ` typo in service description

jonasnick:
Oops forgot to push this

Issue #94, opened 2019-07-20
Improve iOS users' lives


stefan-mihaila:
Sadly, the experience on iOS is deplorable. I can run Spark Wallet with [Onion Browser](https://onionbrowser.com/), but even that is frustrating because WebRTC doesn't work. Even if Spark Wallet would work flawlessly, it seems to me that it doesn't actually allow users to transfer funds on-chain, only off-chain. In consequence, iOS users have no option to send funds on-chain from their phones using their nix-bitcoin node.

I think that this can be mitigated by adding support to run lnd on clearnet in nix-bitcoin, in parallel or as an alternative to clightning. Sadly, it must be in clearnet as there is not any alternative to Orbot on iOS and it might not ever be one because... Apple. iOS users would then be able to use [Zap](https://zap.jackmallers.com) for both on-chain and off-chain transfers with their nix-bitcoin node.

jonasnick:
Hm, if it's as painful on iOS as you're saying then that should be fixed. Would be easier if there was an iOS app for/alternative to spark, but if someone is motivated to add lnd I'd support it.

cypherpunk2140:
Just to be clear, Spark Wallet with Onion Browser is the PWA version of Spark Wallet. There is no native version of Spark Wallet for iOS atm, even though one is stated to come soon™. If there would be (or an alternative), leaving aside the little inconvenient that it can't be run over tor, I would be happy with that.

In the mean time, I looked a little into trying to integrate `lnd` into nix-bitcoin. [Building is trivial](https://github.com/NixOS/nixpkgs/pull/65263/files#diff-890b4fb9d231f348f899b142a4a56e23). I unfortunately got stuck on starting the service for now. I will take another look at it in a few days.

nixbitcoin:
I don't think adding clearnet functionality aligns with the nix-bitcoin networking model. If we do add this it should be an opt-in feature in configuration.nix. The inbound Tor HS model is crucial to the privacy and security of nix-bitcoin users. 

Would there by any way to use iOS's VPN API with Wireguard in a limited fashion to only tunnel zap/spark-wallet (soon™) from iOS to nix-bitcoin nodes?

cypherpunk2140:
In my view, that was the idea, to add an option to run (only) lnd on clearnet in configuration.nix.

> Would there by any way to use iOS's VPN API with Wireguard in a limited fashion to only tunnel zap/spark-wallet (soon™) from iOS to nix-bitcoin nodes?

[...]

PR #107, opened 2019-08-05
Add LND support

cypherpunk2140:


cypherpunk2140:
As it is, it _mostly_ works, with some caveats.

I tried to create the TLS certificates for LND in `generate_secrets.sh` using:
```
openssl ecparam -genkey -name prime256v1 -out secrets/tls.key
openssl req -new -sha256 -key secrets/tls.key -out secrets/csr.csr -subj '/CN=localhost/O=lnd'
openssl req -x509 -sha256 -days 1825 -key secrets/tls.key -in secrets/csr.csr -out secrets/tls.cert
```
The certificates created this way don't work correctly for me, trying to connect to LND using them results in: `authentication handshake failed: x509: cannot validate certificate for 127.0.0.1 because it doesn't contain any IP SANs`. I don't know what the problem is. Maybe it's my `openssl`? (I'm on macOS). As a workaround, I manually started LND and copied the certificates it creates on startup.

In lnd's service `ExecStartPost` I run a script that creates / unlocks the wallet.

For zmq messages, I used a linux socket. The other option would have been to use `tcp://` which doesn't play well with the hardening `RestrictAddressFamilies = = "AF_UNIX AF_INET AF_INET6";`

In this form, lnd connects to peers using Tor, but it doesn't use Tor for gRPC. [At the moment, sadly, not even Zap Desktop support gRPC over Tor.](https://github.com/LN-Zap/zap-desktop/issues/2209) I tested connecting to the lnd service using Zap Desktop with ssh tunneling and seems to work fine using the macaroon at `/var/lib/lnd/chain/bitcoin/mainnet/admin.macaroon`.

 I would have expected Zap connecting to the local ip (192.168.0.x) to work when adding `rpclisten=0.0.0.0:10009` to `lnd.conf`, but it didn't. I didn't figure why yet. I wanted to check that I can connect using the local IP because I assumed it's similar to how it should work using a VPN, as per @nixbitcoin's [suggestion](https://github.com/fort-nix/nix-bitcoin/issues/94#issuecomment-515421935) -- but I'm not sure, I don't know that much about network protocols.

secrets/openssl.cnf
```diff
@@ -0,0 +1,29 @@
...
+emailAddress			= Email Address
+emailAddress_max		= 64
+
+[ req_attributes ]
+challengePassword		= A challenge password
+challengePassword_min		= 4
+challengePassword_max		= 20
+
+[ v3_ca ]
+subjectAltName			= IP:127.0.0.1
```
cypherpunk2140: This is necessary to get rid of `authentication handshake failed: x509: cannot validate certificate for 127.0.0.1 because it doesn't contain any IP SANs` errors.

cypherpunk2140:
> Pretty sure that's because you only put localhost and not 127.0.0.1 in the common name.

Actually, it needs `subjectAltName` in `/etc/ssl/openssl.cnf`. I added an alternative `openssl.cnf` and used it in this PR.

> Yeah, we should lift the restriction and allow bitcoind to create zmq sockets.

I'm gonna take a look into it and try to find the address family.

> I presume that's because enforceTor is true, therefore hardening does only
allowTor (i.e. IPAddressAllow = "127.0.0.1/32 ::1/128";) and not ip
addresses of another subnet.

Actually, it looks to me that `IPAddressAllow` only influences outbound connections. ([See my electrs PR for proof](https://github.com/fort-nix/nix-bitcoin/pull/110)). It might have been filtered by the firewall, but if I remember correctly it didn't work with `networking.firewall.enable = true` either. I will try again.

modules/bitcoind.nix
```diff
@@ -222,6 +222,7 @@ in {
         User = "${cfg.user}";
         Group = "${cfg.group}";
         ExecStart = "${cfg.package}/bin/bitcoind ${cmdlineOptions}";
+        ExecStop = "/nix/store/0cvw3jxxncnm3whckl0qi72bs4g35833-coreutils-8.30/bin/rm -f ${cfg.dataDir}/zmqpubrawblock.sock ${cfg.dataDir}/zmqpubrawtx.sock";
```
nixbitcoin: You cannot assume that coreutils are always at this location. Use `${pkgs.coreutils}/bin/rm` instead.

modules/lnd.nix
```diff
@@ -0,0 +1,136 @@
...
+      wantedBy = [ "multi-user.target" ];
+      requires = [ "bitcoind.service" ];
+      after = [ "bitcoind.service" ];
+      preStart = ''
+        mkdir -m 0770 -p ${cfg.dataDir}
+        rm -f ${cfg.dataDir}/lnd.conf
+        chown 'lnd:lnd' '${cfg.dataDir}'
+        cp ${configFile} ${cfg.dataDir}/lnd.conf
+        chown 'lnd:lnd' '${cfg.dataDir}/lnd.conf'
+        chmod +w ${cfg.dataDir}/lnd.conf
```
jonasnick: Perhaps this is better expressed as `chmod u=rw,g=rw,o= lnd.conf`.

modules/lnd.nix
```diff
@@ -0,0 +1,136 @@
...
+
+  options.services.lnd = {
+    enable = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        If enabled, the LND service will be installed.
+      '';
+    };
+    bitcoin-rpcuser = mkOption {
```
jonasnick: Any particular reason why this is an option and `bitcoind.dataDir` isn't? Perhaps best to just use `config.services.bitcoind.dataDir` because then you can't forget to set the option, both in lnd and in bitcoind.

modules/nix-bitcoin.nix
```diff
@@ -107,7 +121,7 @@ in {
     };
     # Give operator access to onion hostnames
     services.onion-chef.enable = true;
-    services.onion-chef.access.operator = [ "bitcoind" "clightning" "nginx" "liquidd" "spark-wallet" "electrs" "sshd" ];
+    services.onion-chef.access.operator = [ "bitcoind" "clightning" "lnd" "nginx" "liquidd" "spark-wallet" "electrs" "sshd" ];
```
jonasnick: operator still doesn't seem to have access to lnd. For that operator would need to be in the same group and have an alias (see below) to use the correct data directory.

modules/lnd.nix
```diff
@@ -0,0 +1,136 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+  
+let
+  nix-bitcoin-services = pkgs.callPackage ./nix-bitcoin-services.nix { };
+  cfg = config.services.lnd;
+  bitcoindDataDir = config.services.bitcoind.dataDir;
+  configFile = pkgs.writeText "lnd.conf" ''
```
jonasnick: These are reasonable default options, but if  we should have extraoptions to allow people to easily add their own (see my module). I can also add this after this is merged.

cypherpunk2140:
> And yes, if you want to connect from anywhere else but localhost you'll need to adjust the firewall too

Yes, you're right, both firewall and IPAddressAllow were at fault. It's also necessary to add an additional `subjectAltName` IP to the certificate. Anyways, I won't include those changes in the PR, but I'll probably use them until we can get [Zap working over tor](https://github.com/fort-nix/nix-bitcoin/issues/94#issuecomment-522315635).

modules/lnd.nix
cypherpunk2140: Sorry, I don't understand this comment related to `bitcoin-rpcuser`. Lnd can use `bitcoin.conf` to get the rpc user, rpc password and zmq addresses, but I guess for this it would need read access to it, and that's what we want to remove, right? But you're probably talking about something else here and I'm completely missing the point?

jonasnick: Seems like a misunderstaning indeed. You're getting the path of the bitcoin data directory by doing `bitcoindDataDir = config.services.bitcoind.dataDir;` in the let block. Similarly you could do `bitcoindRpcUser = config.services.bitcoind.rpcuser` and then you wouldn't have to have that option which people can forget to set.

secrets/generate_secrets.sh
```diff
@@ -23,3 +24,10 @@ openssl genrsa -out secrets/ssl_certificate_key.key 2048
 openssl req -new -key secrets/ssl_certificate_key.key -out secrets/ssl_certificate.csr -subj "/C=KN"
 openssl x509 -req -days 1825 -in secrets/ssl_certificate.csr -signkey secrets/ssl_certificate_key.key -out secrets/ssl_certificate.crt
 echo Done
+
+echo Generate LND compatible TLS Cert
+openssl ecparam -genkey -name prime256v1 -out secrets/tls.key
```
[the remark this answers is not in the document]
cypherpunk2140: Tried it now, doesn't seem to work, I get `x509: unsupported elliptic curve` on starting lnd :-)

modules/lnd.nix
```diff
@@ -0,0 +1,136 @@
...
+      description = "The data directory for LND.";
+    };
+    enforceTor =  nix-bitcoin-services.enforceTor;
+  };
+
+  config = mkIf cfg.enable {
+    users.users.lnd = {
+        description = "LND User";
+        group = "lnd";
+        extraGroups = [ "bitcoinrpc" "keys" "bitcoin" ];
```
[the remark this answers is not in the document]
cypherpunk2140: There isn't a `nix-bitcoin-services.allowAnyProtocol` flag in `nix-bitcoin-services` :-)
How did you implement it?

Also, that needs to be added to bitcoind as well so it is allowed to create the zmq connections :-|

modules/lnd.nix
jonasnick: I guess `g=r` is sufficient.

modules/bitcoind.nix
```diff
@@ -222,6 +222,7 @@ in {
         User = "${cfg.user}";
         Group = "${cfg.group}";
         ExecStart = "${cfg.package}/bin/bitcoind ${cmdlineOptions}";
+        ExecStop = "${pkgs.coreutils}/bin/rm -f ${cfg.dataDir}/zmqpubrawblock.sock ${cfg.dataDir}/zmqpubrawtx.sock";
```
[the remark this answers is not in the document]
cypherpunk2140: Removed this line when switching to zmq over tcp

modules/lnd.nix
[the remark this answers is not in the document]
jonasnick: > There isn't a nix-bitcoin-services.allowAnyProtocol flag 

Oops, was only in my fork. Opened a PR to add that.

> Also, that needs to be added to bitcoind as well so it is allowed to create the zmq connections 

Yeah, right.

modules/lnd.nix
```diff
@@ -23,11 +22,13 @@ let
     tor.streamisolation=true
 
     [Bitcoind]
-    bitcoind.rpcuser=${cfg.bitcoin-rpcuser}
-    bitcoind.zmqpubrawblock=ipc://${bitcoindDataDir}/zmqpubrawblock.sock
-    bitcoind.zmqpubrawtx=ipc://${bitcoindDataDir}/zmqpubrawtx.sock
+    bitcoind.rpcuser=${config.services.bitcoind.rpcuser}
+    bitcoind.zmqpubrawblock=tcp://127.0.0.1:28332
+    bitcoind.zmqpubrawtx=tcp://127.0.0.1:28333
```
jonasnick: Nit, feel free to ignore. I think ideally we'd add `zmqpubrawblock` and `zmqpubrawtx` options to the bitcoin module with type `nullOr String` and default `null`. Then we can use them here with toString. As a result if you forget to set any of those options, the lnd module will throw errors at build time.

jonasnick:
New commits look good to me so far. Sorry for merging the allowAllAddressFamilies commit. I didn't see that you had already added that to this PR.

modules/lnd.nix
cypherpunk2140: makes sense, will add this as it's easy to do

modules/lnd.nix
```diff
@@ -0,0 +1,136 @@
...
+
+${pkgs.coreutils}/bin/sleep 5
+
+if [ ! -f ${cfg.dataDir}/chain/bitcoin/mainnet/wallet.db ]
+then
+  ${pkgs.coreutils}/bin/echo creating LND wallet
+
+  ${pkgs.curl}/bin/curl -s \
+  --cacert /secrets/tls_cert \
+  -X GET https://127.0.0.1:8080/v1/genseed | ${pkgs.jq}/bin/jq -c '.cipher_seed_mnemonic' > /secrets/lnd-seed-mnemonic
```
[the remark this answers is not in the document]
cypherpunk2140: Option 2, if I understand it correctly, means starting the "dummy" lnd (on the host computer, in nix-shell) and then calling the rest api to generate the seed. I think lnd can't be started without connecting to the bitcoin node.

lncli doesn't seem to have an option to create the seed "offline" either.

Maybe there's another way of generating the seed by using the [aezeed package from lnd](https://github.com/lightningnetwork/lnd/tree/master/aezeed) without starting lnd, but I think that unless there's an easy way to generate the seed, I will go with option 1 for now and option 2 would be left as a possible future improvement.

On the other side, I guess this is not more dangerous than not backing up the clightning seed, is it? Plus that channel state should be backed up also if you don't want to be #reckless :-)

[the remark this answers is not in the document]
cypherpunk2140: Added a comment in `configuration.nix` for now.

pkgs/nodeinfo/nodeinfo.sh
```diff
@@ -2,14 +2,21 @@ set -e
...
+    CLIGHTNING_ONION="$(cat /var/lib/onion-chef/operator/clightning)"
+    CLIGHTNING_ID="$CLIGHTNING_NODEID@$CLIGHTNING_ONION:9735"
+    echo CLIGHTNING_NODEID="$CLIGHTNING_NODEID"
+    echo CLIGHTNING_ONION="$CLIGHTNING_ONION"
+    echo CLIGHTNING_ID="$CLIGHTNING_ID"
+fi
+
+if [ -x "$(command -v lncli)" ]; then
+    LND_NODEID=$(sudo -u lnd lncli --tlscertpath /secrets/lnd_cert --macaroonpath /var/lib/lnd/chain/bitcoin/mainnet/admin.macaroon getinfo | jq -r '.uris[0]')
+    echo LND_NODEID="$LND_NODEID"
```
cypherpunk2140: The onion address reported by `lncli getinfo` is different by that found in `/var/lib/onion-chef/operator/lnd`. I didn't dig into it yet, I don't know how lnd uses `ControlPort` to create services, or what `tor.streamisolation=true` implies. For now I just included the node id reported by `lncli`.

modules/lnd.nix
```diff
@@ -0,0 +1,128 @@
...
+
+${pkgs.coreutils}/bin/sleep 5
+
+if [ ! -f ${cfg.dataDir}/chain/bitcoin/mainnet/wallet.db ]
+then
+  ${pkgs.coreutils}/bin/echo creating LND wallet
+
+  ${pkgs.curl}/bin/curl -s \
+  --cacert /secrets/lnd_cert \
+  -X GET https://127.0.0.1:8080/v1/genseed | ${pkgs.jq}/bin/jq -c '.cipher_seed_mnemonic' > /secrets/lnd-seed-mnemonic
```
cypherpunk2140: I need to add an umask to adjust the permissions of the `lnd-seed-mnemonic` file. I'm also thinking that it might be better to create the file just once and reuse it even if the `wallet.db` file is removed.

pkgs/nodeinfo/nodeinfo.sh
jonasnick: Good catch. It seems like in that case we could remove the explicit hidden service configuration and remove lnd from the onion chef to avoid confusion.

modules/lnd.nix
jonasnick: > it might be better to create the file just once and reuse it even if the wallet.db file is removed

Agreed, that's the safer thing to do.

jonasnick:
Hm, unfortunately the default config (with clightning) doesn't work anymore.
```
error: value is a list while a set was expected
```
I haven't figured out the issue yet.

jonasnick:
Rest seems to work fine!

modules/bitcoind.nix
```diff
@@ -239,7 +251,7 @@ in {
         // (if cfg.enforceTor
           then nix-bitcoin-services.allowTor
           else nix-bitcoin-services.allowAnyIP
-        );
+        ) // optionals config.services.lnd.enable nix-bitcoin-services.allowAnyProtocol;  # FOR ZMQ
```
jonasnick: Finally figured out that this line is at fault (`--show-trace` wasn't helpful at all). `optionals` is for lists, so if lnd is not enabled it returns an empty list. I think we can use `optionalAttrs` (https://github.com/NixOS/nixpkgs/blob/master/lib/attrsets.nix#L322).

cypherpunk2140: ty for testing and investigating this! 🙏🏻

I’ll take a another look at the whole PR later today.

modules/lnd.nix
```diff
@@ -0,0 +1,132 @@
...
+  ${pkgs.curl}/bin/curl -s \
+  --cacert /secrets/lnd_cert \
+  -X POST -d "{\"wallet_password\": \"$(${pkgs.coreutils}/bin/cat /secrets/lnd-wallet-password | ${pkgs.coreutils}/bin/tr -d '\n' | ${pkgs.coreutils}/bin/base64 -w0)\", \
+  \"cipher_seed_mnemonic\": $(${pkgs.coreutils}/bin/cat /secrets/lnd-seed-mnemonic | ${pkgs.coreutils}/bin/tr -d '\n')}" \
+  https://127.0.0.1:8080/v1/initwallet
+else
+  ${pkgs.coreutils}/bin/echo Unlocking lnd wallet
+
+  ${pkgs.curl}/bin/curl -s \
+      -H "Grpc-Metadata-macaroon: $(${pkgs.xxd}/bin/xxd -ps -u -c 1000 ${cfg.dataDir}/chain/bitcoin/mainnet/admin.macaroon)" \
```
guggero: The unlock command is part of the `WalletUnlocker` service that doesn't know about macaroons yet. So this is not really needed. It doesn't hurt either but if you leave it in, I suggest to change the number of columns to 99999 in the `xxd` command, just in case there are a lot of permissions added to the admin macaroon in the future.

configuration.nix
```diff
@@ -30,6 +30,15 @@
   # default nix-bitcoin nodes offer outgoing connectivity.
   # services.clightning.autolisten = true;
 
+  ### LND
+  # Disable clightning and uncomment the following line in order to enable lnd,
+  # a lightning implementation written in Go.
+  # services.lnd.enable = assert (!config.services.clightning.enable); true;
+  # WARNING: If you use lnd, you should manually backup your wallet mnemonic
+  # seed. In order to do so, you can run the following command after the
```
guggero: If you have this warning in here, it would be nice to also add the warning about the `channel.backup` file that is updated with every new channel that is created. Or is this communicated to the user at another point?

jonasnick: Thanks for your review! The sole reason why we added this is because due to our unlocker/walletcreate script users do not see the warning about backing up their seed that they normally see. But we could optionally add a sentence about channel.backup too.

cypherpunk2140: Right, it does make sense to add a comment about channel.backup

modules/lnd.nix
cypherpunk2140: ty for reviewing!

PR #108, opened 2019-08-05
Further clarify liquid/elements relation

nixbitcoin:


nixbitcoin:
Seems a bit wordy on second read.

PR #110, opened 2019-08-07
optionally connect to electrs w/o tor; minor improvements

cypherpunk2140:
- Add config files to .gitignore 

- Move electrs configuration options to configuration.nix

- Make `enforceTor` default to `true` & other small refactors

- Add option to connect to electrs via clearnet


docs/install.md
```diff
@@ -108,7 +108,11 @@ You can also build Nix from source by following the instructions at https://nixo
 	nixops create network/network.nix network/network-vbox.nix -d bitcoin-node
 	```
 
-4. Adjust configuration by opening `configuration.nix` and removing FIXMEs. Enable/disable the modules you want in `configuration.nix`.
+4. Create `configuration.nix` using `sample-configuration.nix`.
```
nixbitcoin: Having a sample-configuration.nix would make it more difficult for us to push updates to configuration.nix like this one. Wouldn't a user have to manually update his configuration.nix from his sample-configuration.nix that isn't .gitignore'd?

modules/nix-bitcoin.nix
```diff
@@ -4,6 +4,7 @@ with lib;
 
 let
   cfg = config.services.nix-bitcoin;
+  electrs-cfg = config.services.electrs;
```
nixbitcoin: What is the point of electrs-cfg?

modules/electrs.nix
```diff
@@ -41,19 +41,26 @@ in {
     port = mkOption {
         type = types.ints.u16;
         default = 50001;
-        description = "Override the default port on which to listen for connections.";
+        description = "RPC port.";
```
nixbitcoin: ACK

modules/electrs.nix
```diff
@@ -41,19 +41,26 @@ in {
...
     };
     onionport = mkOption {
         type = types.ints.u16;
         default = 50002;
-        description = "Override the default port on which to listen for connections.";
+        description = "Port on which to listen for tor client connections.";
```
nixbitcoin: ACK

modules/nix-bitcoin-services.nix
```diff
@@ -32,14 +32,14 @@ in
...
-     type = types.bool;
-     default = false;
-     description = ''
-       "Whether to force Tor on a service by only allowing connections from and
-       to 127.0.0.1;";
-     '';
-   };
+  enforceTor = mkOption {
+    type = types.bool;
+    default = true;
```
nixbitcoin: I'm unsure what consequences this could have. @jonasnick

sample-configuration.nix
```diff
@@ -36,12 +36,32 @@
...
   # Enable this module to use electrs, an efficient re-implementation of
   # Electrum Server in Rust. Only enable this if hardware wallets are
   # disabled.
   # services.electrs.enable = true;
 
+  # If enabled, electrs will sync faster on high-memory systems (≥ 8GB).
+  # services.electrs.high-memory = true;
+
+  # Uncomment if you want to allow peer connections without tor
+  # (not recommended)
```
nixbitcoin: Good

modules/electrs.nix
```diff
@@ -41,19 +41,26 @@ in {
...
     };
-    enforceTor =  nix-bitcoin-services.enforceTor;
+    expose-clearnet-nginxport = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        If enabled, client connections are allowed through clearnet on port nginxport.
+      '';
+    };
+    enforceTor = nix-bitcoin-services.enforceTor;
```
nixbitcoin: utACK

nixbitcoin:
### configuration.nix
I'm open to adding more functionality to `configuration.nix`. However, `configuration.nix` _must_ never allow a novice user to break his node or weaken his security/privacy inadvertently. `configuration.nix` also _should not_ overwhelm the user with FIXMEs. So far we have tried to keep `configuration.nix` to on-off settings, with more granular customization in the modules. Both the must and the should are still major areas for improvement in nix-bitcoin with configuration checks, simpler options, and better documentation.

### secrets/generate_secrets.sh
While you're at it maybe give `secrets/ssl_certificate.crt`, `secrets/ssl_certificate.csr`, `secrets/ssl_certificate_key.key`, and `secrets/tls.key`, `secrets/tls.cert` in #107 more descriptive names. You could also give the electrs and lnd certificate generation the same structure.

docs/install.md
cypherpunk2140: Yes, that is true.

But, in the case of this PR, an old `configuration.nix` should work without breaking anything or changing any functionality. There are just more options exposed in `configuration.nix`, but all those options have sane defaults and can be ignored, and everything would work as before if they are ignored. I would assume we want to preserve this kind of compatibility with future changes also.

So, without this change, in case of non-breaking changes to `configuration.nix`, users would need to read `configuration.nix` to understand what new options have been added and enable them if they want to. With this change, they would need to read `sample-configuration.nix` and add the options they want to their `configuration.nix`.

(In the case of breaking changes, which I assume we want to avoid, without this change users would probably need to do a merge, and with this change they would get an error on trying to deploy.)

modules/nix-bitcoin.nix
cypherpunk2140: This is just a matter of style, I prefer using `electrs-cfg` instead of `config.services.electrs` as it is shorter and keeps the line length shorter also.

modules/nix-bitcoin-services.nix
cypherpunk2140: This is intended to be just a [refactor](0a0254c840730757548eb5343543ccadb30e09b0) and should not change any functionality (unless I overlooked something).

I prefer having sane defaults and not override them unless necessary, but of course a point can also be made that explicit is better than implicit.

cypherpunk2140:
> I'm open to adding more functionality to configuration.nix. However, configuration.nix must never allow a novice user to break his node or weaken his security/privacy inadvertently. configuration.nix also should not overwhelm the user with FIXMEs.

I guess novice user is a little hard to define in the case of a project like this, which in its current form seems to me to target mostly competent users. I'm more of an approach of "we're all consenting adults here". I think that by providing reasonable defaults and good documentation, the user has a choice. If they don't understand an option and they don't want to take the time to understand it, that's why the reasonable default is there, so they can just omit it. But if they want it or need it, the option is there and can be changed without the user needing to dive in the code base. (I was personally surprised when I found the electrs `high-memory` option in the file `nix-bitcoin.nix` _after_ syncing my 16GB machine -- probably it wouldn't have made a big difference, but still...)

Of course, I do agree that we shouldn't overwhelm an user who just wants to get a node running fast with options. I just think that "hiding" configurable options in the code base is not a good way to achieve that.

Hmmm... What about splitting the configurable options in 2 files somehow? Having more advanced options in a different file for more advanced users, but still not require them to go through the whole code base in order for them to change them?

cypherpunk2140:
> While you're at it maybe give secrets/ssl_certificate.crt, secrets/ssl_certificate.csr, secrets/ssl_certificate_key.key, and secrets/tls.key, secrets/tls.cert in #107 more descriptive names. You could also give the electrs and lnd certificate generation the same structure.

Sounds good.

cypherpunk2140:
> Hmmm... What about splitting the configurable options in 2 files somehow? Having more advanced options in a different file for more advanced users, but still not require them to go through the whole code base in order for them to change them?

Another option would have to not add _advanced_ options to `sample-configuration.nix` or `configuration.nix`, but have a doc file with all configurable options. An advanced user will check that doc file and add options he wants. Note that we still need to provide reasonable defaults and remove assignments from `nix-bitcoin.nix` for this to work, as otherwise a redefinition error would be raised on deploy.

docs/install.md
cypherpunk2140: I do agree that it might be nice to preserve the option of pushing updates to `configuration.nix` for some things without bothering the user though. My biggest problem with it being tracked by git is it makes switching between branches more tedious, and that I commit it by mistake sometimes. Commiting by mistake is not such a big deal. I could add it to `.git/info/exclude`, but that won't make switching branches easier. I guess `git stash` is the best way to do it, stashing the configuration file separately so I can pop it separately.

modules/nix-bitcoin-services.nix
jonasnick: I agree with sane defaults. But sane defaults should be set in nix-bitcoin.nix. People using affected modules outside of nix-bitcoin will become _insane_ if the module behaves in such an unexpected way.

cypherpunk2140:
Created https://github.com/fort-nix/nix-bitcoin/pull/112 with the non-controversial changes. Removed the `configuration.nix` rename for know, as it is easily do add if we want to.

PR #114, opened 2019-08-19
Add allowAnyProtocol option to nix-bitcoin-services

jonasnick:


cypherpunk2140:
lgtm, I came up with the same way of doing it in the lnd branch

PR #118, opened 2019-09-03
Move zmq options from nix-bitcoin.nix to bitcoind module

jonasnick:
... which is a better place for this. CC @cypherpunk2140

### DIFF
--- a/modules/bitcoind.nix
+++ b/modules/bitcoind.nix
@@ -253,9 +253,9 @@ in {
         PermissionsStartOnly = "true";
       } // nix-bitcoin-services.defaultHardening
         // (if cfg.enforceTor
-          then nix-bitcoin-services.allowTor
-          else nix-bitcoin-services.allowAnyIP
-        ) // optionalAttrs config.services.lnd.enable nix-bitcoin-services.allowAnyProtocol;  # FOR ZMQ
+            then nix-bitcoin-services.allowTor
+            else nix-bitcoin-services.allowAnyIP)
+        // optionalAttrs (cfg.zmqpubrawblock != null || cfg.zmqpubrawtx != null) nix-bitcoin-services.allowAnyProtocol;
     };
     systemd.services.bitcoind-import-banlist = {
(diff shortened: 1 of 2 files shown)


cypherpunk2140:
ACK 2937f18c8f16ec530f82cafaf91768b26caeb5be

jonasnick:
Putting dependency on `config.services.lnd` in bitcoind.nix? I must have missed this. Replaced that by checking whether zmq options are set.

erikarvstedt:
Tested ACK cf39d88c63303a23e2040cc16807c04f39e6c6f4

Additionally, I think we should remove the dependency on lnd in the bitcoin module (`optionalAttrs config.services.lnd.enable`) and instead set the option in the lnd module, like so:
```
# lnd.nix
systemd.services.bitcoin = nix-bitcoin-services.allowAnyProtocol; # for ZMQ
```

jonasnick:
Or simply set `allowAnyProtocol` when one of the zmq options is set. Added commit to do that. Good catch @erikarvstedt

erikarvstedt:
Ah, that's even better. Good catch, @jonasnick. :smiley_cat:

modules/bitcoind.nix
```diff
@@ -253,9 +253,9 @@ in {
         PermissionsStartOnly = "true";
       } // nix-bitcoin-services.defaultHardening
         // (if cfg.enforceTor
-          then nix-bitcoin-services.allowTor
-          else nix-bitcoin-services.allowAnyIP
-        ) // optionalAttrs config.services.lnd.enable nix-bitcoin-services.allowAnyProtocol;  # FOR ZMQ
+            then nix-bitcoin-services.allowTor
+            else nix-bitcoin-services.allowAnyIP)
+        // optionals (cfg.zmqpubrawblock != null || cfg.zmqpubrawtx != null) nix-bitcoin-services.allowAnyProtocol;
```
erikarvstedt: Should be `optionalAttrs` instead of `optionals`.

jonasnick:
fixed and merged

PR #119, opened 2019-10-02
Update elementsd to 0.18.1.1

jonasnick:
Because `LC_ALL` must be set to make the tests work (https://github.com/bitcoin/bitcoin/issues/14948#issuecomment-447049325) I also copied the rest from the current nixpkgs bitcoind package.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -10,7 +10,7 @@
   nanopos = pkgs.callPackage ./pkgs/nanopos { };
   spark-wallet = pkgs.callPackage ./pkgs/spark-wallet { };
   electrs = (pkgs.callPackage ./pkgs/electrs { }).rootCrate.build;
-  elementsd = pkgs.callPackage ./pkgs/elementsd { };
+  elementsd = pkgs.callPackage ./pkgs/elementsd { withGui = false; };
   hwi = pkgs.callPackage ./pkgs/hwi { };
   pylightning = pkgs.python3Packages.callPackage ./pkgs/pylightning { };
   liquid-swap = pkgs.python3Packages.callPackage ./pkgs/liquid-swap { };
--- a/pkgs/elementsd/default.nix
+++ b/pkgs/elementsd/default.nix
@@ -1,32 +1,46 @@
-{ stdenv, fetchurl, pkgconfig, autoreconfHook, openssl, db48, boost, zeromq
-, zlib, qtbase ? null, qttools ? null, utillinux, protobuf, python3, qrencode, libevent
-}:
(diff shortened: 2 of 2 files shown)


default.nix
```diff
@@ -10,7 +10,7 @@
   nanopos = pkgs.callPackage ./pkgs/nanopos { };
   spark-wallet = pkgs.callPackage ./pkgs/spark-wallet { };
   electrs = (pkgs.callPackage ./pkgs/electrs { }).rootCrate.build;
-  elementsd = pkgs.callPackage ./pkgs/elementsd { };
+  elementsd = pkgs.callPackage ./pkgs/elementsd { withGui = false; };
```
jb55: what do you think about having an attribute for the gui version?

```suggestion
  elementsd = pkgs.callPackage ./pkgs/elementsd { withGui = false; };
  elements = pkgs.callPackage ./pkgs/elementsd { withGui = true; };
```

pkgs/elementsd/default.nix
```diff
@@ -1,33 +1,47 @@
-{ stdenv, fetchurl, pkgconfig, autoreconfHook, openssl, db48, boost, zeromq
-, zlib, qtbase ? null, qttools ? null, utillinux, protobuf, python3, qrencode, libevent
-}:
+{ stdenv, fetchurl, pkgconfig, autoreconfHook, openssl, db48, boost, zeromq, rapidcheck
+, zlib, miniupnpc, qtbase ? null, qttools ? null, wrapQtAppsHook ? null, utillinux, protobuf, python3, qrencode, libevent
```
jb55: you shouldn't need `? null` on qtbase,qttools, etc.

I'm getting an evaulation error on these when withGui is true

jonasnick: The function args just copied from the bitcoin nixpkg. The way this works is that instead of `callPackage` they use `libsForQt5.callPackage`. However, I'm getting a weird error after a bit of buidling

```
  GEN      qt/paymentrequest.pb.h
/nix/store/x1khw8x0465xhkv6w31af75syyyxc65j-gnused-4.7/bin/sed -e 's:\(res\/\(icons\|movies\)/\):/tmp/nix-build-elements-0.18.1.1.drv-0/elements-elements-0.18.1.1/src/qt/\1:' <"qt/bitcoin.qrc" >"qt/qrc_bitcoin_build.qrc"
make[2]: *** No rule to make target 'qt/res/rendered_icons/about.png', needed by 'qt/qrc_bitcoin.cpp'.  Stop.
```

For now, I removed the withGui = true commit again.

PR #124, opened 2019-10-28
Update to NixOS 19.09

jonasnick:


jonasnick:
thanks for reviewing @erikarvstedt

PR #125, opened 2019-11-08
Fix virtualbox deployment for 19.09

erikarvstedt:
Without this, starting the virtualbox guest service fails during machine activation.
This is due to an incompatible NixOS machine base image.
Fix this by using an updated version of nixops.

I've already reported [this issue](https://github.com/NixOS/nixops/issues/1207) upstream.

@jonasnick, could you check if this new nixops version works
correctly for your non-virtualbox deployments?

### DIFF
--- a/pkgs/nixops/default.nix
+++ b/pkgs/nixops/default.nix
@@ -1,0 +1,53 @@
+# An updated version of nixops that's compatible with machines running NixOS 19.09.
+# 19.09 demands a suitable base image (defined in nixops-vbox/nix/virtualbox.nix) to
+# start the virtualbox guest service during system activation.
+
+{ stdenv, nixops, runCommand, fetchFromGitHub }:
+
+let
+  pluginData = {
+    aws = {
+      owner = "nixos";
+      repo = "nixops-aws";
+      rev= "v1.0.0";
+      sha256 = "1if6spscsgd6ckivgvbqza5fvvn5hbafi1n8q0fw98s3xpz2hjfm";
+    };
+    hetzner = {
+      owner = "nixos";
+      repo = "nixops-hetzner";
+      rev = "v1.0.0";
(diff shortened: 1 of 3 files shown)


pkgs/nixops/default.nix
```diff
@@ -0,0 +1,53 @@
...
+    owner = "NixOS";
+    repo = "nixops";
+    rev = "2434bf26e0bba49441041ffce36dc324f049bc00";
+    sha256 = "0ag05pjwwqdw8in49hr8m8bdg31xsgqs1cawcqyh6a5lsys7f6zg";
+  };
+
+  src = runCommand "src" {} ''
+    cp --no-preserve=mode -r ${origSrc} $out
+    cd $out
+    patch -p1 < ${./release.nix.patch}
```
jonasnick: Not sure why we need to patch. Can't we just provide additional plugins (vbox, libvirtd) through the `p` argument as mentioned in the nixops README?

erikarvstedt:
Yes, that's possible, like so: https://github.com/erikarvstedt/nix-bitcoin/commit/baeeea98a56a443c4917ea68a7ec9b138bfaefad.

But I thought the patched version is a bit easier to review. I've added a fixup
commit (to be squashed before merging) that makes the patch as simple and minimal as
possible.

In another fixup commit, I've amended the patch to silence a nixops plugin debug message during
evaluation that clutters nix-shell's output. This would also be possible, but
more convoluted, without the patching approach.

jonasnick:
One of the points of using a not using patching was to remove the redundant definitions of `aws` and `hetzner`. Also, the `p` route seems to be the intended way to add plugins. But anyway, the hashes look right and removing the debug message is a fun bonus. I still don't like that we have to switch to nixops master but that's unovoidable for now. ACK. Feel free to squash.

erikarvstedt:
Done. Hopefully we can remove this again soon.

jonasnick:
Thanks for your help!

PR #128, opened 2019-11-11
Add content hashes for pinned channels

erikarvstedt:
Unhashed external content is bad for security and performance (due to
re-fetches when the cache times out).
Use this simple fix until https://github.com/fort-nix/nix-bitcoin/pull/78 is fleshed out.

For testing, run this in the repo root dir:
```bash
nix eval '(import ./pkgs/nixpkgs-pinned.nix)'
```


pkgs/fetch-channel
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
```
jonasnick: How about putting this in `contrib/`?

erikarvstedt:
Yeah, I thought about that for a second, but `contrib` has a [slightly different scope](https://softwareengineering.stackexchange.com/a/339925/310061).
I'll let you decide what to do:
1. Put it in contrib/
2. Rename contrib/ to helper/ and put it in helper/

jonasnick:
You're right. I prefer 2. too.

PR #130, opened 2019-11-12
Improve banlist importer

erikarvstedt:
Please see the individual commit messages for more infos.

This time I've used commit msg titles starting with lower-case letters.
I think this style is the simplest and cleanest, especially with
`topic:` prefixes. Let me know if I should reformat this.

Regarding commit `banlist: simplify script, remove package`:
I've added the current version of Greg's banlist, but it has 5600 entries instead of the previous 739. Is this ok?

Rationale for commit `bitcoind: add cli option`:
For some cases it can be useful when bitcoin-cli is an executable instead of a shell alias.
  - non-interactive ssh commands like `ssh mynode 'bitcoin-cli getnetworkinfo'`.
    This can also be achieved with shell aliases in `environment.shellInit`, but it's good
    practice to avoid aliases in noninteractive shells.
  - Scripts that call `bitcoin-cli`. The banlist importer which uses this is a good example.


modules/bitcoind.nix
```diff
@@ -209,6 +209,14 @@ in {
         example = "tcp://127.0.0.1:28333";
         description = "ZMQ address for zmqpubrawtx notifications";
       };
+      cli = mkOption {
```
jonasnick: cool idea

jonasnick:
Fwiw Greg said his methodology hasn't changed. Apparently the new spy nodes are mostly from china.

PR #137, opened 2020-02-19
Add nixops19_09 to default pkgs.

jonasnick:
It's helpful to be able to use these packages when importing nix-bitcoin. Opened a PR separately from #136 because that allows experimentation with the idea before adjusting the docs and everything.

### DIFF
--- a/pkgs/default.nix
+++ b/pkgs/default.nix
@@ -10,6 +10,7 @@
   pylightning = pkgs.python3Packages.callPackage ./pylightning { };
   liquid-swap = pkgs.python3Packages.callPackage ./liquid-swap { };
   generate-secrets = pkgs.callPackage ./generate-secrets { };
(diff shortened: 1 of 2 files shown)


pkgs/default.nix
```diff
@@ -10,6 +10,6 @@
   pylightning = pkgs.python3Packages.callPackage ./pylightning { };
   liquid-swap = pkgs.python3Packages.callPackage ./liquid-swap { };
   generate-secrets = pkgs.callPackage ./generate-secrets { };
-
   pinned = import ./pinned.nix;
+  nixops19_09 = pkgs.callPackage ./nixops { };
```
erikarvstedt: Nit: I'd keep that visual separation:  Normal packages, blank line, pinned.

jonasnick: Sure, fixed

erikarvstedt:
ACK

PR #138, opened 2020-02-23
Update nixpkgs (lnd 0.8.1 -> lnd 0.9.0, clightning 0.8.0 -> 0.8.1)

jonasnick:


jonasnick:
Hmm, I had forgotten that locally you'll only build the same derivation for the module test if you run with the same `numCPUs` and `memoryMiB` as travis: `memoryMiB=3072 numCPUs=2 ./test/run-tests.sh`.

erikarvstedt:
Why did you want to re-run the test locally?
You can also use `run-tests.sh run` which always executes the test.

jonasnick:
Travis may time out in some of the tests, but we can make it pass nonetheless by running the tests locally and putting them on cachix.

erikarvstedt:
Ah, got it. 
Yes, the integration test times out on channel updates. But one manual build restart is generally enough to make it pass.

erikarvstedt:
ACK f4ef99e4e0db7f9217d87a03270fde031b01e921

.travis.yml
```diff
@@ -49,6 +49,7 @@ script:
     }
   - buildExpr=$(getBuildExpr)
   - time nix-instantiate -E "$buildExpr" --add-root ./drv --indirect
+  - nix show-derivation ./drv
```
erikarvstedt: I think this is too noisy, e.g. look at this build output: https://travis-ci.org/fort-nix/nix-bitcoin/jobs/654613360
What was your reason to add this?

jonasnick: Hm, that's indeed too noisy. Reason was to figure out how my local derivation differed from what is built on travis (which turned out to be the wrong `numCPUs` and `memoryMiB` as mentioned above). IIRC if you cat on travis it doesn't print the output unless you expand that section.

jonasnick:
> But one manual build restart is generally enough to make it pass.

I didn't know that, but I restarted the build a couple of times and it never passed.

.travis.yml
jonasnick: Hmm perhaps that only happens in `after_script` and not `script`.

erikarvstedt: Yeah, Travis doesn't show the `cat` output.
`nproc` and memory are the only build host variables that influence the test derivation.
I've added https://github.com/fort-nix/nix-bitcoin/pull/140/files to show these in full detail.
I think that's good enough.

erikarvstedt:
A new channel update build takes 60-70 min on Travis' GCE nodes, so it it normally takes one rebuild when cachix uploading is enabled.

erikarvstedt:
ACK b6d7e81c31d52756b3b1054322b41e4c3901c14d

Issue #66, opened 2019-05-11
check out crate2nix for electrs builds


jb55:
I'm slowly in the process of upstreaming some of this stuff to nixos/nixpkgs. I noticed electrs was crashing on some blocks for some reason so I updated to the latest version, only to find that the rust packages in whatever rust channel nixpkgs was using was out of date.

there's a handy tool called [crate2nix](https://github.com/kolloch/crate2nix) which does fully self contained packaging of cargo projects. I'm giving it a spin now, just thought I'd mention it incase we want to switch to it in here for more reliable builds.

nixbitcoin:
This is definitely something we're interested in. Could you share the resulting crate2nix expression so we can start using fully self contained packaging for electrs?

jb55:
here's the branch I've been working on:

https://github.com/jb55/electrs/commits/master

to install crate2nix I cloned it and ran this because I didn't want to use their pinned nixpkgs:

```bash
nix-build --cores 14 -E 'with import <nixpkgs> {}; callPackage ./default.nix { inherit pkgs; }'
nix-env -i ./result
```

then I just ran

```bash
$ crate2nix generate
```
in the electrs path. I had to fix up the generated rocksdb derivation since it seemed broken, I left that as a separate commit so you can check it out.

PR #144, opened 2020-02-28
Electrs fixes

erikarvstedt:
The electrs package update is needed for fixing the exposed secret.


modules/electrs.nix
```diff
@@ -5,11 +5,13 @@ let
   cfg = config.services.electrs;
   inherit (config) nix-bitcoin-services;
   secretsDir = config.nix-bitcoin.secretsDir;
-  index-batch-size = "${if cfg.high-memory then "" else "--index-batch-size=10"}";
-  jsonrpc-import = "${if cfg.high-memory then "" else "--jsonrpc-import"}";
 in {
   imports = [
     (mkRenamedOptionModule [ "services" "electrs" "nginxport" ] [ "services" "electrs" "TLSProxy" "port" ])
+    (mkRemovedOptionModule [ "services" "electrs" "high-memory" ] ''
+      This option has been removed because its implementation was defective, leading
```
jonasnick: I have never used this module outside of testing but I don't get what youre' saying here. If `high-memory` was set, then jsonrpc-import was _not_ set. The [electrs docs](https://github.com/romanz/electrs/blob/master/doc/usage.md#usage) state that if these options are not set in a 2GB memory system, then there will be `memory allocation of xxxxxxxx bytes failedAborted`. In @nixbitcoin's case this happened with 4GB memory even.

pkgs/electrs/default.nix
```diff
@@ -1,2933 +1,25 @@
...
-      };
-    "rand_core 0.3.1 (registry+https://github.com/rust-lang/crates.io-index)"
-      = rec {
-        crateName = "rand_core";
-        version = "0.3.1";
-        edition = "2015";
-        sha256 = "0q0ssgpj9x5a6fda83nhmfydy7a6c0wvxm0jhncsmjx8qp8gw91m";
-        authors = [
-          "The Rand Project Developers"
-          "The Rust Project Devel
```
jonasnick: Do you know what exactly is the input to the hash? There seemed to have been problems with fully pinning the dependencies before we used crate2nix (https://github.com/fort-nix/nix-bitcoin/issues/66 @jb55). Other than that I'd be very happy to simplify this by going back to buildRustPackage.

modules/electrs.nix
erikarvstedt: I completely misread that if-statement. Sorry, that's embarrassing. Reverted.

modules/electrs.nix
```diff
@@ -66,16 +79,18 @@ in {
       preStart = ''
         mkdir -m 0770 -p ${cfg.dataDir}
         chown -R '${cfg.user}:${cfg.group}' ${cfg.dataDir}
-        echo "cookie = \"${config.services.bitcoind.rpcuser}:$(cat ${secretsDir}/bitcoin-rpcpassword)\"" \
-          > electrs.toml
+        {
+          echo "cookie = \"${config.services.bitcoind.rpcuser}:$(cat ${secretsDir}/bitcoin-rpcpassword)\"";
+          echo "verbose = ${toString logLevels.${cfg.logLevel}}";
+        } >> electrs.toml
```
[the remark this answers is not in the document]
erikarvstedt: Fixed. That was a leftover from testing. (Although it wouldn't matter here because it's writing to a temporary runtime dir.)

pkgs/electrs/default.nix
erikarvstedt: You mean `cargoSha256`? It's the fixed output hash on the result of running `cargo vendor`. (Here's the [derivation source](https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/rust/fetchCargoTarball.nix).)
 With the current state of `buildRustPackage` the crate2nix approach has no advantages when it comes to pinning. Both fetch the exact dependencies defined in cargo.lock. 

To quote @jb55, the reason for switching to crate2nix was

> only to find that the rust packages in whatever rust channel nixpkgs was using was
> out of date

crate2nix doesn't pin `rustPackages` (that is rustc, cargo, ...), so in what way did it help here?

modules/electrs.nix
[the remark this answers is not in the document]
jonasnick: Oh I see, it's not writing to the datadir.

erikarvstedt:
I've removed option `logLevel` because I don't like the impl complexity for such a trivial setting.
I've added `extraArgs` instead, and I'll add an upstream patch to electrs to allow overriding the verbosity via cmdline args.
ACK? I'll squash then.

jonasnick:
reACK

jb55:
ACK, I believe I just misunderstood how nixpkgs did it. The only benefit of my PR was to have build-cached dependencies, but using crate2nix is probably more brittle than just letting cargo do the build.

PR #152, opened 2020-04-07
Rename nix-bitcoin.nix to presets/secure-node.nix

erikarvstedt:
Note that all commits up to and including `extract variable 'cfg'` are pure refactorings without any effects on evaluation results.
You can verify this with [this helper script](https://gist.github.com/5b0b1a7787ebc1342fcc3f237da5afa7), which compares the VM test derivation of the current worktree to the one in master.
Copy it to the repo root and use it like this:
```
git checkout ab61794 # commit 'extract variable 'cfg'
./diff-vm-drv.sh
# => No changes
git checkout 6c22e13 # commit 'copy-root-authorized-keys: use inline script definition'
./diff-vm-drv.sh
# => diff of changes
```

### DIFF
--- a/examples/configuration.nix
+++ b/examples/configuration.nix
@@ -18,10 +18,7 @@
     #./hardware-configuration.nix
   ];
   # FIXME: Enable modules by uncommenting their respective line. Disable
-  # modules by commenting out their respective line.  Enable this module to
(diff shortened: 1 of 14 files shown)


erikarvstedt:
Rebased to master.

PR #153, opened 2020-04-07
Improve bitcoind

erikarvstedt:


jonasnick:
Looks good! Needs rebase.

modules/presets/secure-node.nix
```diff
@@ -46,8 +46,7 @@ in {
     services.bitcoind = {
       enable = true;
       listen = true;
-      sysperms = if cfg.electrs.enable then true else null;
-      disablewallet = if cfg.electrs.enable then true else null;
+      services.bitcoind.dataDirReadableByGroup = mkIf config.services.electrs.enable true;
```
nixbitcoin, line 49: Should be `dataDirReadableByGroup = mkIf cfg.electrs.enable true;`

`services.bitcoind.`: when deploying causes 
```
error: The option `services.bitcoind.services' defined in `/home/nix-bitcoin/nix-bitcoin/modules/presets/secure-node.nix does not exist.
```

`config.services`:  Other options use `cfg` so I think this one should too

erikarvstedt: Thanks for catching this, this slipped in through the rebase.

PR #154, opened 2020-04-08
nixops: fix format exception from upstream nixops

jonasnick:
Without this our nixops doesn't build with unstable nixpkgs.

I still don't understand why the nixpkgs are not properly pinned. Our `nixops` should be built with nothing but `(import ./pkgs/nixpkgs-pinned.nix).nixpkgs;` but apparently that's not the case.

### DIFF
--- a/pkgs/nixops/release.nix.patch
+++ b/pkgs/nixops/release.nix.patch
@@ -43,3 +43,15 @@
                                     ('destDir',     'string'),
                                     ('user',        'string'),
                                     ('group',       'string'),
+
+--- a/nixops/deployment.py
++++ b/nixops/deployment.py
+@@ -748,7 +748,7 @@
+                 if res == 100 or force_reboot or m.state == m.RESCUE:
+                     if not allow_reboot and not force_reboot:
+                         raise Exception("the new configuration requires a "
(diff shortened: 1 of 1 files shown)


jonasnick:
To reproduce run the following:
```
$ NIX_PATH=nixpkgs=channel:nixpkgs-unstable nix-shell
```

That's result in 
```
/nix/store/mjnjm1zzdl8q5k5r0mcmqcnqxwh6xdcy-python2.7-setuptools-44.0.0/lib/python2.7/site-packages/setuptools/dist.py:485: UserWarning: The version specified ('1.8pre0_abcdef') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
  "details." % self.metadata.version
running test
......................................
----------------------------------------------------------------------
Ran 38 tests in 0.010s

OK
nixops/deployment.py:750: error:(B Not all arguments converted during string formatting(B
Found 1 error in 1 file (checked 20 source files)(B
```

erikarvstedt:
Fixed in https://github.com/fort-nix/nix-bitcoin/pull/155.

jonasnick:
#155 fixed this

Issue #53, opened 2019-04-28
Restrict directory access for services


jonasnick:
We're already restricting access relatively okay by giving every service its own user. However there's at least one problem: the classic way to gain persistence as an unprivileged user is to add something similar to a `.bashrc` file into the users home which is then executed every time the user logs in. We should make that impossible by not giving access to the users home directory. This would in principle work right now already through systemd's `ProtectHomes`, except that the users home directories are often set to the services data directory, instead of `/home`.

nixbitcoin:
I'll take care of this

PR #159, opened 2020-04-19
Remove bitcoin, clightning, electrs, liquid user home directory

nixbitcoin:
Closes https://github.com/fort-nix/nix-bitcoin/issues/53

Combined with the fact that AFAICT you can't login with `bitcoin`, `clightning`, `electrs`, `liquid` users, this should completely eliminate the possibility of an attacker placing something similar to a `.bashrc` file into the users home which is then executed every time the user logs in.

For example:
```console
[root@nix-bitcoin:/var/lib]# sudo -u bitcoin printenv
LANG=en_US.UTF-8
TERMINFO_DIRS=/root/.nix-profile/share/terminfo:/etc/profiles/per-user/root/share/terminfo:/nix/var/nix/profiles/default/share/terminfo:/run/current-system/sw/share/terminfo
TERM=xterm-256color
PATH=/root/bin:/run/wrappers/bin:/root/.nix-profile/bin:/etc/profiles/per-user/root/bin:/nix/var/nix/profiles/default/bin:/run/current-system/sw/bin
MAIL=/var/mail/bitcoin
LOGNAME=bitcoin
USER=bitcoin
HOME=/var/empty
SHELL=/run/current-system/sw/bin/nologin
SUDO_COMMAND=/run/current-system/sw/bin/printenv
SUDO_USER=root
SUDO_UID=0
SUDO_GID=0
LOCALE_ARCHIVE=/run/current-system/sw/lib/locale/locale-archive
NIX_PATH=nixpkgs=/nix/var/nix/profiles/per-user/root/channels/nixos:nixos-config=/etc/nixos/configuration.nix:/nix/var/nix/profiles/per-user/root/channels
TZDIR=/etc/zoneinfo
```
The rest of the system should be protected with `ProtectHome=true` and `ProtectSystem=full`

Tested with `nixops deploy` on NixOS nix-bitcoin node

### DIFF
--- a/modules/bitcoind.nix
+++ b/modules/bitcoind.nix
@@ -330,7 +330,6 @@ in {
     users.users.${cfg.user} = {
       group = cfg.group;
       description = "Bitcoin daemon user";
-      home = cfg.dataDir;
     };
     users.groups.${cfg.group} = {};
     users.groups.bitcoinrpc = {};
--- a/modules/clightning.nix
+++ b/modules/clightning.nix
@@ -8,6 +8,7 @@ let
   configFile = pkgs.writeText "config" ''
     autolisten=${if cfg.autolisten then "true" else "false"}
     network=bitcoin
+    bitcoin-datadir=${config.services.bitcoind.dataDir}
     ${optionalString (cfg.proxy != null) "proxy=${cfg.proxy}"}
     always-use-proxy=${if cfg.always-use-proxy then "true" else "false"}
     ${optionalString (cfg.bind-addr != null) "bind-addr=${cfg.bind-addr}"}
@@ -76,7 +77,6 @@ in {
         description = "clightning User";
         group = "clightning";
         extraGroups = [ "bitcoinrpc" ];
-        home = cfg.dataDir;
     };
     users.groups.clightning = {};
 
--- a/modules/electrs.nix
+++ b/modules/electrs.nix
@@ -107,7 +107,6 @@ in {
       description = "electrs User";
       group = cfg.group;
       extraGroups = [ "bitcoinrpc" "bitcoin"];
-      home = cfg.dataDir;
     };
     users.groups.${cfg.group} = {};
   }
--- a/modules/liquid.nix
+++ b/modules/liquid.nix
@@ -235,7 +235,6 @@ in {
     users.users.${cfg.user} = {
       group = cfg.group;
       description = "Liquid sidechain user";
-      home = cfg.dataDir;
     };
     users.groups.${cfg.group} = {};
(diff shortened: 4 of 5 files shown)


erikarvstedt:
This breaks [the test](https://travis-ci.org/github/fort-nix/nix-bitcoin/builds/676863387). 
Can be fixed by re-adding homes for lnd and clightning.
I've tried to diagnose the clightning failure: Here's [relevant log output](https://gist.github.com/45f1a8451a42e0493ce54a682095c5f8), which shows that clightning can't connect to bitcoind RPC without HOME. I couldn't find out in a reasonable amount of time why that's the case.

I don't see any attack vector here at all (could you explain it in more detail?), but I'm glad to see uneeded options removed, so ACK for the remaining changes.

nixbitcoin:
> I don't think this attack vector is serious, but I'm glad to see uneeded options removed, so ACK for the remaining changes.

I agree, but no home directory means no home directory level persistence. We should also continue to remove unneeded options, and simplify nix-bitcoin as much as possible.

> This breaks the test.

Upon restarting my clightning service, it also broke my installation.

I was able to fix clightning by specifying `bitcoin-datadir` in the clightning config file.

I applied the same measure to lnd, let's see if that also fixes it.

erikarvstedt:
Lnd is still failing:
```
machine # [   52.406372] lnd[1954]: loadConfig: Failed to create lnd directory: mkdir /var/empty/.lnd: operation not permitted
```
Could you try to run the tests before committing?
```
nix-bitcoin/test/run-tests.sh
```
To avoid GC'ing the result, run them like this:
```
nix-bitcoin/test/run-tests.sh build --out-link ~/.cache/nix-builds/nix-bitcoin-test
```

nixbitcoin:
I'm getting error
```
error: a 'x86_64-linux' with features {kvm, nixos-test} is required to build '/nix/store/ibb137rfhryllbp76657i36ib941wmma-vm-test-run-nix-bitcoin.drv', but I am a 'x86_64-linux' with features {benchmark, big-parallel, nixos-test}
```
Which dependencies do I need for your tests?

erikarvstedt:
You need [KVM](https://wiki.archlinux.org/index.php/KVM).

nixbitcoin:
LND seems to not work without a home directory. I exempted it from this PR.

erikarvstedt:
Could you add a comment in `lnd.nix` to explain why `home` must be set? Something like `# lnd creates .lnd dir in HOME`.

nixbitcoin:
Comment added to `lnd.nix`. Ready to merge in my opinion.

PR #160, opened 2020-04-19
Document low-entropy problem

nixbitcoin:
I had a problem with low system entropy. I want to document this to spare people the troubleshooting.


nixbitcoin:
Please still merge, it's a valuable pointer for people troubleshooting. What would be a saner alternativ to apg?

PR #163, opened 2020-04-22
Make it easy to use pinned packages

jb55:
It appears the pinned attrset is a bit adhoc. This generalizes
pinned.nix so that you can do:

    $ nix build -f . pinned.stable.hwi
    $ nix build -f . pinned.unstable.hwi

### DIFF
--- a/pkgs/pinned.nix
+++ b/pkgs/pinned.nix
@@ -1,13 +1,18 @@
 let
   nixpkgsPinned = import ./nixpkgs-pinned.nix;
-  unstable = import nixpkgsPinned.nixpkgs-unstable { config = {}; overlays = []; };
-  nixBitcoinPkgsUnstable = import ./. { pkgs = unstable; };
+  nixpkgsStable = import nixpkgsPinned.nixpkgs { config = {}; overlays = []; };
+  nixpkgsUnstable = import nixpkgsPinned.nixpkgs-unstable { config = {}; overlays = []; };
+  nixBitcoinPkgsStable = import ./. { pkgs = nixpkgsStable; };
+  nixBitcoinPkgsUnstable = import ./. { pkgs = nixpkgsUnstable; };
 in
 {
-  inherit (unstable)
+  inherit (nixpkgsUnstable)
     bitcoin
     bitcoind
     clightning
     lnd;
   inherit (nixBitcoinPkgsUnstable) electrs;
+
+  stable = nixBitcoinPkgsStable;
+  unstable = nixBitcoinPkgsUnstable;
 }


pkgs/pinned.nix
```diff
@@ -1,13 +1,16 @@
 let
   nixpkgsPinned = import ./nixpkgs-pinned.nix;
-  unstable = import nixpkgsPinned.nixpkgs-unstable { config = {}; overlays = []; };
-  nixBitcoinPkgsUnstable = import ./. { pkgs = unstable; };
+  nixpkgs-unstable = import nixpkgsPinned.nixpkgs-unstable { config = {}; overlays = []; };
+  nixpkgs-stable = import nixpkgsPinned.nixpkgs { config = {}; overlays = []; };
 in
-{
+rec {
   inherit (unstable)
```
erikarvstedt, line 7: These must be inherited from unstable nixpkgs, not the unstable nix-bitcoin packages, which don't include, e.g., bitcoind.

pkgs/pinned.nix
```diff
@@ -1,13 +1,16 @@
...
     bitcoin
     bitcoind
     clightning
     lnd;
-  inherit (nixBitcoinPkgsUnstable) electrs;
+  inherit (stable) electrs;
```
erikarvstedt, line 12: This must be inherited the unstable nix-bitcoin packages, as before.

pkgs/pinned.nix
```diff
@@ -1,13 +1,16 @@
 let
   nixpkgsPinned = import ./nixpkgs-pinned.nix;
-  unstable = import nixpkgsPinned.nixpkgs-unstable { config = {}; overlays = []; };
-  nixBitcoinPkgsUnstable = import ./. { pkgs = unstable; };
+  nixpkgs-unstable = import nixpkgsPinned.nixpkgs-unstable { config = {}; overlays = []; };
```
erikarvstedt, line 3: I think we should use Nix' camelCase naming conventions.

pkgs/pinned.nix
jb55: oh right this is a typo, thanks

pkgs/pinned.nix
jb55: you are correct, not sure what I was smoking when I did this PR

pkgs/pinned.nix
```diff
@@ -1,13 +1,18 @@
 let
   nixpkgsPinned = import ./nixpkgs-pinned.nix;
-  unstable = import nixpkgsPinned.nixpkgs-unstable { config = {}; overlays = []; };
-  nixBitcoinPkgsUnstable = import ./. { pkgs = unstable; };
+  nixpkgsUnstable = import nixpkgsPinned.nixpkgs-unstable { config = {}; overlays = []; };
+  nixpkgsStable = import nixpkgsPinned.nixpkgs { config = {}; overlays = []; };
+  nixBitcoinPkgsUnstable = import ./. { pkgs = nixpkgsUnstable; };
+  nixBitcoinPkgsStable = import ./. { pkgs = nixpkgsStable; };
```
erikarvstedt, line 6: Nit: Put stable before unstable.
```suggestion
  nixpkgsStable = import nixpkgsPinned.nixpkgs { config = {}; overlays = []; };
  nixpkgsUnstable = import nixpkgsPinned.nixpkgs-unstable { config = {}; overlays = []; };
  nixBitcoinPkgsStable = import ./. { pkgs = nixpkgsStable; };
  nixBitcoinPkgsUnstable = import ./. { pkgs = nixpkgsUnstable; };
```

Issue #32, opened 2019-04-13
CoinJoin Integration


nixbitcoin:
**Time Horizon:** Medium Term

**Description:** Integrate a CoinJoin service into nix-bitcoin.

**Motivation:** Running a CoinJoin service 24/7 is desirable to maximize probability of success. Users may not wish to keep their laptops on 24/7, but nix-bitcoin is designed to be on 24/7. So it makes sense to run a CoinJoin service on a nix-bitcoin node.

**Possible Implementation:**

Option 1: Integrate [JoinMarket](https://github.com/JoinMarket-Org/joinmarket-clientserver) maker as well as taker, using its command-line UI over ssh.

Option 2: Integrate [Wasabi](https://github.com/zkSNACKs/WalletWasabi) Daemon as-is by pushing users walletfile to nix-bitcoin node as a temporary file with `nixops deploy` and starting a mix on the nix-bitcoin node (`wassabee --mix --wallet <WalletName>`). When the user wishes to stop the mix, nix-bitcoin stops the `wassabee --mix` and automatically deletes the walletfile. When the user starts up his Wasabi GUI, it should display the now post-mix coins.

Possible problems
* User going online while mix is ongoing on nix-bitcoin node
* User stopping `wassabee --mix` in the middle of a round, possibly causing DOS punishments
* Wasabi GUI not being able to accurately display new state after `wassabee --mix` on the nix-bitcoin node. This could possibly be fixed by copying the walletfile back to the users computer after he stops the mix.

Option 3: Wait for a fully-featured Wasabi headless daemon (receive, mix and send) and integrate that

Possible problems
* Both Option 2 and Option 3 (ab)use Wasabi as an intermediary wallet for CoinJoining

Option 4: Wait for and integrate Samourai's whirlpool daemon

SamouraiDev:
Samourai whirlpool daemon and GUI are ready for release. Ping us.

nopara73:
Hi there, I'll share some of my thoughts on the specific issues you brought up and I'll bring up other ones, too.

I'd say let me know if there's something we can implement, but I won't. Currently we ended up adding too many features, which makes the wallet less reliable and more cluttery. Because of this we had to commit to a "feature freeze" to fix up all the instabilities for the time being: https://github.com/zkSNACKs/WalletWasabi/issues/1412

[...]

PR #164, opened 2020-04-24
Add JoinMarket Clientserver

nixbitcoin:
This PR integrates JoinMarket into nix-bitcoin. All user scripts are created with `pkgs.writeScriptBin` and are executed in the `cfg.dataDir` with the right user and options. `tumbler.py` can be started in the background using GNU Screen. Yieldgenerator can be run as a systemd service.

How could I firewall the entire joinmarket user to only use Tor?

Thank you @erikarvstedt for your essential help throughout packaging this complex python program.

Closes https://github.com/fort-nix/nix-bitcoin/issues/32

### DIFF
--- a/examples/configuration.nix
+++ b/examples/configuration.nix
@@ -172,6 +172,16 @@
   # and electrs data directory, enable
   # services.backups.with-bulk-data = true;
 
+  ### JOINMARKET
+  # Enable this module to allow using JoinMarket's user interactive scripts (including
+  # tumbler.py).
+  # Note: JoinMarket has full access to bitcoind, including its wallet functionality.
+  # services.joinmarket.enable = true;
+  # Enable this option to enable the JoinMarket Yield Generator Bot. You will be able to
+  # earn sats by providing CoinJoin liquidity. This makes it impossible to use other
+  # scripts that access your wallet.
+  # services.joinmarket.yieldgenerator.enable = true;
+
   # FIXME: Define your hostname.
   networking.hostName = "nix-bitcoin";
   time.timeZone = "UTC";
--- a/modules/joinmarket.nix
+++ b/modules/joinmarket.nix
@@ -1,0 +1,202 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.services.joinmarket;
+  inherit (config) nix-bitcoin-services;
+  secretsDir = config.nix-bitcoin.secretsDir;
+
+  torAddress = builtins.head (builtins.split ":" config.services.tor.client.socksListenAddress);
+  configFile = builtins.toFile "config" ''
+    # Based on https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/jmclient/jmclient/configure.py
+    [DAEMON]
+    no_daemon = 0
+    daemon_port = 27183
+    daemon_host = localhost
+    use_ssl = false
+
+    [BLOCKCHAIN]
+    blockchain_source = bitcoin-rpc
+    network = mainnet
+    rpc_host = ${builtins.elemAt config.services.bitcoind.rpcbind 0}
+    rpc_port = 8332
+    rpc_user = ${config.services.bitcoind.rpc.users.privileged.name}
+    @@RPC_PASSWORD@@
+
+    [MESSAGING:server1]
+    host = darksci3bfoka7tw.onion
+    channel = joinmarket-pit
+    port = 6697
+    usessl = true
+    socks5 = true
+    socks5_host = ${torAddress}
+    socks5_port = 9050
+
+    [MESSAGING:server2]
+    host = ncwkrwxpq2ikcngxq3dy2xctuheniggtqeibvgofixpzvrwpa77tozqd.onion
+    channel = joinmarket-pit
+    port = 6667
+    usessl = false
+    socks5 = true
+    socks5_host = ${torAddress}
+    socks5_port = 9050
+
+    [LOGGING]
+    console_log_level = INFO
+    color = false
+
+    [POLICY]
+    segwit = true
+    native = false
+    merge_algorithm = default
+    tx_fees = 3
+    absurd_fee_per_kb = 350000
+    tx_broadcast = self
+    minimum_makers = 4
+    max_sats_freeze_reuse = -1
+    taker_utxo_retries = 3
+    taker_utxo_age = 5
+    taker_utxo_amtpercent = 20
+    accept_commitment_broadcasts = 1
+    commit_file_location = cmtdata/commitments.json
+  '';
+
+   # The jm scripts create a 'logs' dir in the working dir,
+   # so run them inside dataDir.
+   cli = pkgs.runCommand "joinmarket-cli" {} ''
+     mkdir -p $out/bin
+     jm=${pkgs.nix-bitcoin.joinmarket}/bin
+     cd $jm
+     for bin in jm-*; do
+       {
+         echo "#!${pkgs.bash}/bin/bash";
+         echo "cd '${cfg.dataDir}' && ${cfg.cliExec} sudo -u ${cfg.user} $jm/$bin --datadir='${cfg.dataDir}' \"\$@\"";
+       } > $out/bin/$bin
+     done
+     chmod -R +x $out/bin
+   '';
+in {
+  options.services.joinmarket = {
+    enable = mkEnableOption "JoinMarket";
+    yieldgenerator = {
+      enable = mkEnableOption "yield generator bot";
+      customParameters = mkOption {
+        type = types.str;
+        default = "";
+        example = ''
+          txfee = 200
+          cjfee_a = 300
+        '';
+        description = ''
+          Python code to define custom yield generator parameters, as described in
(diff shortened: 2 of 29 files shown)


nixbitcoin:
Rebased

nixbitcoin:
Rebased and ready for review

nixbitcoin:
Updated to JoinMarket 0.7.0 & rebased. Integrated `netns-isolation` and `cliExec`.

I'm relying on manual wallet generation at the moment as described in `/docs/usage.md`. This has advantages if the user wants to customize his wallet experience, with/without 2FA or Fidelity Bonds for example. However, it also creates friction because the user must manually enter his passwords in the local secrets to use the `joinmarket-yieldgenerator` bot.

We could implement a low-level python script to automatically create a wallet and save the seed, as in our LND module. An example of such a python script can be found in https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/scripts/joinmarket-qt.py#L2010-L2018.

erikarvstedt:
[Fixups](https://github.com/erikarvstedt/nix-bitcoin/commits/joinmarket).

Regarding `python-bitcointx: fix provision of openssl dependency`: This is a good example on how nix differs from traditional packaging by making all runtime deps explicit. It's really helpful to understand why the previous approach didn't work.

I'm not very happy about `pkg: make bin script names more cli friendly`, but I think that's the best way to go for now. I've suggested an improved CLI [at the jm repo](https://github.com/JoinMarket-Org/joinmarket-clientserver/issues/681).

A script to automatically create a wallet would be a great idea, if only for automated testing.

Why does joinmarketd need privileged bitcoin RPC access?

erikarvstedt:
How would a user configure the yieldgenerator? Edit `joinmarket.cfg`? We should then add a `extraOptions` option.

erikarvstedt:
So joinmarket uses bitcoind wallet functionality? We should then set `bitcoind.disablewallet = false` and add a comment why have to use the privileged RPC user.

nixbitcoin:
`joinmarket: automatically generate wallet` now no longer WIP.

I've also added 0535da4. I was getting
```
Sep 12 18:18:50 nix-bitcoin bash[15886]: Amp server or network failure unhandled by client application.  Dropping connection!  To avoid, add errbacks to ALL remote commands!
Sep 12 18:18:50 nix-bitcoin bash[15886]: Traceback (most recent call last):
Sep 12 18:18:50 nix-bitcoin bash[15886]: --- <exception caught here> ---
Sep 12 18:18:50 nix-bitcoin bash[15886]:   File "/nix/store/6li72ddcdwcdg1m4zdghv2xjw6a72sf7-python3-3.7.6-env/lib/python3.7/site-packages/twisted/internet/defer.py", line 654, in _runCallbacks
Sep 12 18:18:50 nix-bitcoin bash[15886]:     current.result = callback(current.result, *args, **kw)
Sep 12 18:18:50 nix-bitcoin bash[15886]:   File "/nix/store/6li72ddcdwcdg1m4zdghv2xjw6a72sf7-python3-3.7.6-env/lib/python3.7/site-packages/jmclient/client_protocol.py", line 51, in defaultErrback
Sep 12 18:18:50 nix-bitcoin bash[15886]:     ConnectionLost) #pragma: no cover
Sep 12 18:18:50 nix-bitcoin bash[15886]:   File "/nix/store/6li72ddcdwcdg1m4zdghv2xjw6a72sf7-python3-3.7.6-env/lib/python3.7/site-packages/twisted/python/failure.py", line 460, in trap
Sep 12 18:18:50 nix-bitcoin bash[15886]:     self.raiseException()
Sep 12 18:18:50 nix-bitcoin bash[15886]:   File "/nix/store/6li72ddcdwcdg1m4zdghv2xjw6a72sf7-python3-3.7.6-env/lib/python3.7/site-packages/twisted/python/failure.py", line 488, in raiseException
Sep 12 18:18:50 nix-bitcoin bash[15886]:     raise self.value.with_traceback(self.tb)
Sep 12 18:18:50 nix-bitcoin bash[15886]: twisted.protocols.amp.UnknownRemoteError: Code<UNKNOWN>: Unknown Error
```
when a peer tried to take a yieldgenerator offer. I figure a missing `txtorcon` runtime dependency could have been the issue.

nixbitcoin:
The missing `txtorcon` runtime dependency was not the issue. Instead, joinmarket daemon was trying to create `commitmentlist` in the `scripts` directory which was causing a `builtins.OSError: [Errno 30] Read-only file system: 'commitmentlist'`. Fixed in 1e7f845f1df.

I also changed the output of `genwallet.py` and grepped the specific string `recovery_seed` to make sure that seed storage only succeeds when `genwallet.py` really succeeded.

modules/joinmarket.nix
```diff
@@ -0,0 +1,222 @@
...
+
+with lib;
+
+let
+  cfg = config.services.joinmarket;
+  inherit (config) nix-bitcoin-services;
+  secretsDir = config.nix-bitcoin.secretsDir;
+
+  torAddress = builtins.head (builtins.split ":" config.services.tor.client.socksListenAddress);
+  configFile = builtins.toFile "config" ''
```
jonasnick, line 11: Where's that config from? Perhaps add comment. Do we need to state all these options, or does joinmarket pick the defaults if we don't? In the latter case we should remove these options because then we always have the most up to date defaults.

modules/joinmarket.nix
```diff
@@ -0,0 +1,222 @@
...
+    cli = mkOption {
+      default = cli;
+    };
+    inherit (nix-bitcoin-services) cliExec;
+  };
+
+  config = mkIf cfg.enable (mkMerge [{
+    environment.systemPackages = [
+      (hiPrio cfg.cli)
+      pkgs.screen
```
jonasnick, line 120: That's up to the user's preferences. I use tmux instead of screen, and I don't know anymore how to use screen, so I wouldn't want it installed.

jonasnick (commented):
I skimmed through the changes and they look good mostly. My issue with this PR is that we should avoid working around things that should be fixed in joinmarket itself and we shouldn't deviate too much from normal usage of joinmarket. In particular,

- I don't really see what the point of the genwallet script is. It means that we're not consistent with joinmarket's usage guide (https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/docs/USAGE.md#using-the-wallet-toolpy-script), requires custom usage instructions and keeping them up-to-date.
- The usage documentation should not do much more than describing the diff to normal joinmarket usage, i.e. the script renamings, how to use the yield generator, etc.
- We can not automatically remove the wallet lock. That's potentially dangerous. If that's a joinmarket problem, it should be fixed upstream

modules/joinmarket.nix
nixbitcoin: > Where's that config from? Perhaps add comment.

Fixed in `fixup! module: explain where defaultConfig comes from`

> Do we need to state all these options, or does joinmarket pick the defaults if we don't?

AFAICT from https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/jmclient/jmclient/configure.py, we do need to state all these options.

modules/joinmarket.nix
nixbitcoin: Fixed in `fixup! module: remove screen from systemPackages`

nixbitcoin:
> I don't really see what the point of the genwallet script is. It means that we're not consistent with joinmarket's usage guide (https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/docs/USAGE.md#using-the-wallet-toolpy-script), requires custom usage instructions and keeping them up-to-date.

The point of the genwallet script is to automatically create a JoinMarket wallet for ease of use and testing. This allows us to easily use yieldgenerator with a `generate-secrets` generated password, and store & backup the wallet's seed. It also allows us to test an essential function of JoinMarket, wallet generation, easily in the CI.

> The usage documentation should not do much more than describing the diff to normal joinmarket usage, i.e. the script renamings, how to use the yield generator, etc.

Fixed in `fixup! docs: describe diff to normal joinmarket usage`

> We can not automatically remove the wallet lock. That's potentially dangerous. If that's a joinmarket problem, it should be fixed upstream

Fixed in `fixup! module: don't automatically remove wallet lock`. @erikarvstedt Do you ACK?

modules/joinmarket.nix
jonasnick: Ok

modules/joinmarket.nix
jonasnick: cool

examples/configuration.nix
```diff
@@ -172,6 +172,15 @@
   # and electrs data directory, enable
   # services.backups.with-bulk-data = true;
 
+  ### JOINMARKET
+  # Enable this module to allow using JoinMarket's user interactive scripts (including
```
jonasnick, line 176: I'd mention here that joinmarket has priviledged bitcoind access.

jonasnick (commented):
Looks good mod nit. Feel free to squash.

examples/configuration.nix
nixbitcoin: Implemented in `fixup! module: note joinmarket's full access to bitcoind`

nixbitcoin:
Squashed all fixups

jonasnick (approved):
ACK dd882753e6e7782372980c75e9f537904eab6c4e

PR #169, opened 2020-04-30
Update spark-wallet 0.2.13 -> 0.2.14

nixbitcoin:
I updated spark-wallet to 0.2.14 and changed it to fetch spark-wallet sources from GitHub, so we can verify @shesek's signature. To replicate run `generate.sh`.

### DIFF
--- a/pkgs/spark-wallet/generate.sh
+++ b/pkgs/spark-wallet/generate.sh
@@ -1,3 +1,27 @@
 #!/usr/bin/env nix-shell
-#! nix-shell -i bash -p nodePackages.node2nix
-exec node2nix --nodejs-10 -i pkg.json -c composition.nix --no-copy-node-env --supplement-input supplement.json
+#! nix-shell -i bash -p nodePackages.node2nix gnupg wget jq moreutils
+set -euo pipefail
+
+TMPDIR="$(mktemp -d -p /tmp)"
+trap "rm -rf $TMPDIR" EXIT
+
+# Get/verify spark-wallet-npm.tgz
+version="0.2.14"
+export GNUPGHOME=$TMPDIR
+gpg --keyserver hkps://hkps.pool.sks-keyservers.net --recv-key FCF19B67866562F08A43AAD681F6104CD0F150FC
+wget -P $TMPDIR https://github.com/shesek/spark-wallet/releases/download/v${version}/SHA256SUMS.asc
+wget -P $TMPDIR https://github.com/shesek/spark-wallet/releases/download/v${version}/spark-wallet-${version}-npm.tgz
+(cd $TMPDIR; gpg --verify $TMPDIR/SHA256SUMS.asc; sha256sum -c --ignore-missing $TMPDIR/SHA256SUMS.asc)
+shasum=$(sha256sum $TMPDIR/spark-wallet-${version}-npm.tgz | cut -d\  -f1)
+
+# Make qrcode-terminal a strict dependency so that node2nix includes it in the package derivation.
+tar xvf $TMPDIR/spark-wallet-*-npm.tgz -C $TMPDIR
+jq '.dependencies["qrcode-terminal"] = .optionalDependencies["qrcode-terminal"]' $TMPDIR/package/package.json | sponge $TMPDIR/package/package.json
+
+# Run node2nix
+cp pkg.json $TMPDIR/pkg.json
+node2nix --nodejs-10 -i $TMPDIR/pkg.json -c composition.nix --no-copy-node-env --supplement-input supplement.json
+
+# Use verified source in node-packages.nix
+url="https://github.com/shesek/spark-wallet/releases/download/v$version/spark-wallet-$version-npm.tgz"
+sed -i '/packageName = "spark-wallet";/!b;n;n;c\    src = fetchurl {\n      url = "'$url'";\n      sha256 = "'$shasum'";\n    };' node-packages.nix
--- a/pkgs/spark-wallet/node-packages.nix
+++ b/pkgs/spark-wallet/node-packages.nix
@@ -13,24 +13,6 @@ let
         sha512 = "LeSfP9bNZH2UOZgcGcZ0PIHUt1ZuHub1L3CVmEyqLxCeDLm4C5Gi8jRH8ZX2PNpDhQCo0z6y/+DIs2JlliXW8w==";
       };
     };
-    "@root/mkdirp-1.0.0" = {
-      name = "_at_root_slash_mkdirp";
-      packageName = "@root/mkdirp";
-      version = "1.0.0";
-      src = fetchurl {
-        url = "https://registry.npmjs.org/@root/mkdirp/-/mkdirp-1.0.0.tgz";
-        sha512 = "hxGAYUx5029VggfG+U9naAhQkoMSXtOeXtbql97m3Hi6/sQSRL/4khKZPyOF6w11glyCOU38WCNLu9nUcSjOfA==";
-      };
-    };
-    "@root/request-1.5.0" = {
-      name = "_at_root_slash_request";
-      packageName = "@root/request";
-      version = "1.5.0";
-      src = fetchurl {
-        url = "https://registry.npmjs.org/@root/request/-/request-1.5.0.tgz";
-        sha512 = "J9RUIwVU99/cOVuDVYlNpr4G0A1/3ZxhCXIRiTZzu8RntOnb0lmDBMckhaus5ry9x/dBqJKDplFIgwHbLi6rLA==";
-      };
-    };
     "accepts-1.3.7" = {
       name = "accepts";
       packageName = "accepts";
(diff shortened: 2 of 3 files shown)


pkgs/spark-wallet/verify.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env nix-shell
+#! nix-shell -i bash -p gnupg
+set -euo pipefail
+
+TMPDIR="$(mktemp -d -p /tmp)"
+trap "rm -rf $TMPDIR" EXIT
+
+version=$(grep spark pkg.json | cut -d^  -f2 | cut -d '"' -f1) # Get only spark-wallet version number from pkg.json
+export GNUPGHOME=$TMPDIR
+gpg --keyserver hkps://keyserver.ubuntu.com --recv-key FCF19B67866562F08A43AAD681F6104CD0F150FC
```
erikarvstedt, line 10: `electrs/get-sha256.sh` uses `hkps://hkps.pool.sks-keyservers.net`. We should choose one that we use consistently.

nixbitcoin: This was based on a mistaken belief that shesek's updated key was only available from keyserver.ubuntu.com, fixed in 79785f6

nixbitcoin:
> Hm it's not super great that we fetch the dependencies from the npm version of spark and only then replace spark with shesek's version. If the npm version would have a malicious dependency then the verify.sh script wouldn't help. But it's better than what we have now.

Can we get the whole package with dependencies from shesek's version?

> How about moving the content of verify.sh into generate.sh? You probably always want to run these 2 commands together so having them separate just makes it more difficult to remember how to update spark.

Fixed in 1343d2e

shesek:
> Can we get the whole package with dependencies from shesek's version?

It looks like `node2nix` [supports](https://github.com/svanderburg/node2nix#generating-a-tarball-from-a-nodejs-development-project) "Any version specification that NPM supports can be used, such as version numbers, version ranges, HTTP(S) URLs, Git URLs, and GitHub identifiers."

So this can be done by changing the version number to the tarball url in [pkgs/spark-wallet/pkg.json](https://github.com/fort-nix/nix-bitcoin/blob/master/pkgs/spark-wallet/pkg.json), like so:

```diff
diff --git a/pkgs/spark-wallet/pkg.json b/pkgs/spark-wallet/pkg.json
index 0c1f11b..5901b81 100644
--- a/pkgs/spark-wallet/pkg.json
+++ b/pkgs/spark-wallet/pkg.json
@@ -1,3 +1,3 @@
 [
-  { "spark-wallet": "^0.2.13" }
+  { "spark-wallet": "https://github.com/shesek/spark-wallet/releases/download/v0.2.14/spark-wallet-0.2.14-npm.tgz" }
 ]
```

nixbitcoin:
Looks like @clefru tried that before and it failed because optionalDependencies are [handled differently for non-registry sources](https://github.com/svanderburg/node2nix/issues/137).

I actually came all the way to that point with a working signature verification, but optionalDepencies are fucking it all up.

Apart from patching the tar.gz or node2nix, I don't see a way how it's going to get better than 1343d2e.

shesek:
Ugh. Well, that's annoying. :(

How about you verify the signature first, then unpack it and patch the `package.json` to copy `optionalDependencies` into `dependencies`, then let `node2nix` run with it?

Something like this with `jq`:
```
jq '.dependencies = .dependencies + .optionalDependencies' package.json | sponge package.json
```

shesek:
Or actually, if you don't need all the tls/letsencrypt stuff and just want `--print-qr`/`--pairing-qr` to work,

```
jq '.dependencies["qrcode-terminal"] = .optionalDependencies["qrcode-terminal"]' package.json | sponge package.json
```

(this will obviously need updating if I add new optional dependencies that you want to have installed, but I don't anticipate adding any additional optional dependencies soon)

nixbitcoin:
> How about you verify the signature first, then unpack it and patch the `package.json` to copy `optionalDependencies` into `dependencies`, then let `node2nix` run with it?

I had that idea too, but I don't see a way to make `node2nix` run with the patched `package.json` seeing as it needs a remote URL. I tried adding `qrcode-terminal` as a supplemental package but spark-wallet couldn't find the module. I asked in https://github.com/svanderburg/node2nix/issues/137 because I have no idea how to fix that, if you have any ideas let me know.

Last option would be manually patching `node-packages.nix`

shesek:
`node2nix` claims that "Any version specification that NPM supports can be used", this should (hopefully?) include [local paths](https://docs.npmjs.com/files/package.json#local-paths) too.

nixbitcoin:
I updated `pkg.json` to pull in the sources from a local folder created by `generate.sh`. Is there a better place to put the sources other than in the pkgs/spark-wallet directory itself?

nixbitcoin:
I updated the commit again to remove the spark-wallet source from the directory. It is now used in a TMPDIR for node2nix generation and references to it are edited out of `node-packages.nix`.

I think it's more elegant to just fix node2nix, but I wasn't able to fix it when I tried.

pkgs/spark-wallet/generate.sh
```diff
@@ -1,3 +1,26 @@
...
+# Make qrcode-terminal dependency
+tar xvf $TMPDIR/spark-wallet-*-npm.tgz -C $TMPDIR
+jq '.dependencies["qrcode-terminal"] = .optionalDependencies["qrcode-terminal"]' $TMPDIR/package/package.json | sponge $TMPDIR/package/package.json
+
+# Run node2nix
+cp pkg.json $TMPDIR/pkg.json
+node2nix --nodejs-10 -i $TMPDIR/pkg.json -c composition.nix --no-copy-node-env --supplement-input supplement.json
+
+# Clean up node-packages.nix
+sed -i '/packageName = "spark-wallet";/!b;n;n;c\    src = fetchurl {\n      url = "https://github.com/shesek/spark-wallet/releases/download/v'$version'/spark-wallet-'$version'-npm.tgz";\n      sha256 = "'$shasum'";\n    };' node-packages.nix
```
erikarvstedt, line 26:
```suggestion
# Use verified source in node-packages.nix
url="https://github.com/shesek/spark-wallet/releases/download/v$version/spark-wallet-$version-npm.tgz"
sed -i '/packageName = "spark-wallet";/!b;n;n;c\    src = fetchurl {\n      url = "'$url'";\n      sha256 = "'$shasum'";\n    };' node-packages.nix
```

pkgs/spark-wallet/generate.sh
```diff
@@ -1,3 +1,26 @@
...
-#! nix-shell -i bash -p nodePackages.node2nix
-exec node2nix --nodejs-10 -i pkg.json -c composition.nix --no-copy-node-env --supplement-input supplement.json
+#! nix-shell -i bash -p nodePackages.node2nix gnupg wget jq moreutils
+set -euo pipefail
+
+TMPDIR="$(mktemp -d -p /tmp)"
+trap "rm -rf $TMPDIR" EXIT
+
+# Get/verify spark-wallet-npm.tgz
+version=$(cat version)
```
erikarvstedt, line 9:
```suggestion
version="0.2.14"

# Get/verify spark-wallet-npm.tgz
```
and remove file `version`. I think an extra file for just the version is unnecessary.

pkgs/spark-wallet/generate.sh
```diff
@@ -1,3 +1,26 @@
...
+export GNUPGHOME=$TMPDIR
+gpg --keyserver hkps://hkps.pool.sks-keyservers.net --recv-key FCF19B67866562F08A43AAD681F6104CD0F150FC
+wget -P $TMPDIR https://github.com/shesek/spark-wallet/releases/download/v${version}/SHA256SUMS.asc
+wget -P $TMPDIR https://github.com/shesek/spark-wallet/releases/download/v${version}/spark-wallet-${version}-npm.tgz
+(cd $TMPDIR; gpg --verify $TMPDIR/SHA256SUMS.asc; sha256sum -c --ignore-missing $TMPDIR/SHA256SUMS.asc)
+shasum=$(sha256sum $TMPDIR/spark-wallet-${version}-npm.tgz | cut -d\  -f1)
+
+# Make qrcode-terminal dependency
```
erikarvstedt, line 17: Please add a comment why this is needed.

nixbitcoin: Resolved in e1527c7

pkgs/spark-wallet/generate.sh
nixbitcoin: Resolved in e1527c7

pkgs/spark-wallet/generate.sh
nixbitcoin: Resolved in e1527c7

pkgs/spark-wallet/generate.sh
erikarvstedt: Sorry, I meant that it wasn't clear to me why the hacking packing.json hacking was needed in the first place.
But I think I get it now.
Here's a comment that would clear it up:
`Make qrcode-terminal a strict dependency so that node2nix includes it in the package derivation.`

nixbitcoin: Resolved in 9049f8d

I get the confusion. I would love to fix it upstream in node2nix but I couldn't do it when I tried.

erikarvstedt:
Thanks for the fixups. You can squash now.

nixbitcoin:
Done

nixbitcoin:
FWIW `npm-shrinkwrap.json` in @shesek's `spark-npm.tgz` does pin the dependencies by hash. I don't think node2nix uses them though...

Issue #170, opened 2020-05-03
Run services using clightning under different users


jonasnick:
Currently these services run under the clightning user because clightnings RPC socket is only readable by the owner. However, there's a flag now to set the permissions on the RPC socket, which would allow making it group readable:
```
lightning --rpc-file-mode <arg>                Set the file mode (permissions) for the JSON-RPC socket (default: "0600")
```

nixbitcoin:
Already updated in an unpushed branch.

I updated lightning-charge to use the group writeable and readable socket. I left the lightning-cli option because I think it isn't a good idea to run lightning-cli as root under nodeinfo or as any other user than clightning in general.

Issue #145, opened 2020-03-01
Support having another username for "operator"


stevenroose:

nixbitcoin:
I think this adds an unneeded degree of complexity. For me it also creates an unnecessary opportunity to doxx myself when sharing screenshots/code snippets. Open to other opinions though @jonasnick .

erikarvstedt:
It would be rather simple to implement, the only prerequisite is to convert `pkgs/nodeinfo` to a module.
Also, the feature can be helpful when nix-bitcoin is used as part of a larger Nixos config that already has a operator-like user.

nixbitcoin:
Ok, I'll take a look at it

PR #171, opened 2020-05-04
Hardening DAC

nixbitcoin:
This PR all-around hardens our existing discretionary access control.

All lightning modules (spark-wallet, recurring-donations, lightning-charge) now run under their own user with clightning group membership. They can interact with c-lightning via group-readable/writable socket. I didn't remove the lightning-cli option because I think it's a good idea to always run it under the clightning user (otherwise it's executed with root permissions in nodeinfo), please take a look at this.

bitcoinrpc group is removed and bitcoin group membership is reduced to a minimum. Secrets are gotten via preStart with PermissionsStartOnly.

electrs also has an improved and truly optional TLSProxy.

I wasn't able to reduce the permissions of nodeinfo, onion-chef and create-webindex from root. I couldn't find an elegant way to fulfill their functions as a less-privileged user without introducing excessive complexity.

LND was out-of-scope for this PR, because I don't run it and have no experience with it.

Closes https://github.com/fort-nix/nix-bitcoin/issues/145
Closes https://github.com/fort-nix/nix-bitcoin/issues/170


modules/presets/secure-node.nix
```diff
@@ -132,8 +132,6 @@ in {
     services.onion-chef.enable = true;
     services.onion-chef.access.operator = [ "bitcoind" "clightning" "nginx" "liquidd" "spark-wallet" "electrs" "sshd" ];
 
-    # Unfortunately c-lightning doesn't allow setting the permissions of the rpc socket
-    # https://github.com/ElementsProject/lightning/issues/1366
     security.sudo.configFile =
      (optionalString cfg.clightning.enable ''
```
jonasnick, line 137: Yeah I'm fine with keeping the lightning-cli option with sudo. It's simpler than having to deal groups.

modules/clightning.nix
```diff
@@ -111,6 +110,10 @@ in {
         while [[ ! -e ${cfg.dataDir}/bitcoin/lightning-rpc ]]; do
             sleep 0.1
         done
+        # give group "search" access to allow clightning group processes to find lightning-rpc
+        if [[ -e ${cfg.dataDir}/bitcoin ]]; then
+          chmod g+X ${cfg.dataDir}/bitcoin
+        fi
       '';
     };
   };
```
jonasnick, line 119: I think we can remove group read access from the config file (chmodded in preStart). Otherwise spark-wallet, lightning-charge, etc. can read the bitcoin rpc credentials. This was probably necessary back when the operator did not sudo to the clightning user to run lightning-cli.

modules/nodeinfo.nix
```diff
@@ -0,0 +1,82 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.services.nodeinfo;
+  inherit (config) nix-bitcoin-services;
```
jonasnick, line 7: not used

modules/nodeinfo.nix
```diff
@@ -0,0 +1,82 @@
...
+  dataDir = "/var/lib/nodeinfo/";
```
jonasnick, line 8: not used

pkgs/default.nix
```diff
@@ -1,6 +1,5 @@
 { pkgs ? import <nixpkgs> {} }:
 {
-  nodeinfo = pkgs.callPackage ./nodeinfo { };
```
jonasnick, line 3: why make nodeinfo a module? I can see some benefits like being able to easily access nix variables but this does not seem to happen in the module.

jonasnick: Oh I missed that the operator user is used in the script! Would be better if the operator-name was an option in the secure-node.nix

modules/electrs.nix
```diff
@@ -106,7 +106,7 @@ in {
     users.users.${cfg.user} = {
       description = "electrs User";
       group = cfg.group;
-      extraGroups = [ "bitcoinrpc" "bitcoin"];
+      extraGroups = mkIf cfg.high-memory [ "bitcoin" ];
```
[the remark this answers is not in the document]
nixbitcoin: Does it matter if this is done on `cfg.high-memory` or `config.services.bitcoind.dataDirReadableByGroup`?

modules/clightning.nix
```diff
@@ -111,6 +110,10 @@ in {
         while [[ ! -e ${cfg.dataDir}/bitcoin/lightning-rpc ]]; do
             sleep 0.1
         done
+        # give group "search" access to allow clightning group processes to find lightning-rpc
```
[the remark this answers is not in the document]
nixbitcoin: spark-wallet and lightning-charge need to be able to "look" into the directory to find `lightning-rpc`. Otherwise you get
```
Lightning client connection error { Error: connect EACCES /var/lib/clightning/bitcoin
```
in spark-wallet for example.

modules/clightning.nix
```diff
@@ -111,6 +110,10 @@ in {
...
+        if [[ -e ${cfg.dataDir}/bitcoin ]]; then
+          chmod g+X ${cfg.dataDir}/bitcoin
```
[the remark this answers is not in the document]
nixbitcoin: Agree, because we apply it directly to a directory and the user execute permissions are always set.

modules/bitcoind.nix
```diff
@@ -268,7 +268,7 @@ in {
           mkdir -m 0770 -p '${cfg.dataDir}/blocks'
         fi
         chown -R '${cfg.user}:${cfg.group}' '${cfg.dataDir}'
-        chmod -R g+rX '${cfg.dataDir}/blocks'
+        ${if cfg.dataDirReadableByGroup then "chmod -R g+rX '${cfg.dataDir}/blocks'" else "chmod -R g-rX '${cfg.dataDir}/blocks'"}
```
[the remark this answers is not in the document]
nixbitcoin: We have to set permissions here because we need to make _existing_ blocks directories and their contents group-readable/unreadable depending on the `cfg.dataDirReadableByGroup` option. This is relevant for users switching `cfg.high-memory` on and off. If we only do it on creation we don't handle this case.

If we create both `cfg.dataDir` and `cfg.dataDir/blocks` with `systemd.tmpfiles.rules -d` we don't handle the blocks directory's contents. If we try to use `systemd.tmpfiles.rules -Z` to handle the blocks directory's contents, we can't handle directories and files differently like with `chmod g+X`

So I believe the way we do it now is best.

https://www.freedesktop.org/software/systemd/man/tmpfiles.d.html

modules/lightning-charge.nix
```diff
@@ -24,25 +24,37 @@ in {
...
+    users.groups.lightning-charge = {};
+
     environment.systemPackages = [ pkgs.nix-bitcoin.lightning-charge ];
     systemd.services.lightning-charge = {
       description = "Run lightning-charge";
       wantedBy = [ "multi-user.target" ];
       requires = [ "clightning.service" ];
       after = [ "clightning.service" ];
+      preStart = ''
+        # Give existing lightning-charge.db right permissions
```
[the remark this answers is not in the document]
nixbitcoin: Resolved in b1cfad4

modules/clightning.nix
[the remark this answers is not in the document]
nixbitcoin: Resolved in 664ce4b

modules/clightning.nix
```diff
@@ -111,6 +110,10 @@ in {
         while [[ ! -e ${cfg.dataDir}/bitcoin/lightning-rpc ]]; do
             sleep 0.1
         done
+        # give group "search" access to allow clightning group processes to find lightning-rpc
+        if [[ -e ${cfg.dataDir}/bitcoin ]]; then
```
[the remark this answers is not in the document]
nixbitcoin: Resolved in 0b8a343

modules/electrs.nix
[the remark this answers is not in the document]
nixbitcoin: Resolved in bedcaf8

examples/configuration.nix
```diff
@@ -119,6 +119,9 @@
   networking.hostName = "nix-bitcoin";
   time.timeZone = "UTC";
 
+  # FIXME: Change less-privileged-user name
```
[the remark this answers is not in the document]
nixbitcoin: Resolved in ca8476a

modules/nodeinfo.nix
nixbitcoin: Resolved in ca8476a

modules/nodeinfo.nix
nixbitcoin: Resolved in ca8476a

pkgs/default.nix
nixbitcoin: Resolved in ca8476a

modules/nodeinfo.nix
```diff
@@ -0,0 +1,82 @@
...
+      ''
+        exec ${pkgs.bash}/bin/bash ${nodeinfo-script} "$@"
+      '';
+      description = "Script to execute nodeinfo.";
+    };
+  };
+
+  config = mkIf cfg.enable {
+    environment.systemPackages = [ (hiPrio cfg.cli) ];
+  };
```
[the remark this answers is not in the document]
nixbitcoin: Resolved in 9dc2e63

modules/clightning.nix
nixbitcoin: Resolved in a0e579d

modules/lightning-charge.nix
```diff
@@ -24,23 +29,29 @@ in {
...
+
     environment.systemPackages = [ pkgs.nix-bitcoin.lightning-charge ];
     systemd.services.lightning-charge = {
       description = "Run lightning-charge";
       wantedBy = [ "multi-user.target" ];
       requires = [ "clightning.service" ];
       after = [ "clightning.service" ];
       preStart = ''
-        # Give existing lightning-charge.db right permissions
+        # Move existing lightning-charge.db
```
jonasnick, line 44: Perhaps add a TODO that this can be removed eventually.

modules/bitcoind.nix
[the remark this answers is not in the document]
jonasnick: Ok makes sense

modules/lightning-charge.nix
```diff
@@ -14,35 +14,51 @@ in {
...
   config = mkIf cfg.enable {
+    users.users.lightning-charge = {
+        description = "lightning-charge User";
+        group = "lightning-charge";
+        extraGroups = [ "clightning" ];
+    };
+    users.groups.lightning-charge = {};
+
+
+    systemd.tmpfiles.rules = [
```
jonasnick, line 33: see https://github.com/fort-nix/nix-bitcoin/pull/174#issuecomment-629636278

nixbitcoin: See response

jonasnick: all good

modules/lightning-charge.nix
nixbitcoin: Resolved in 1165fc6

jonasnick:
ACK want to squash?

nixbitcoin:
Should I squash all into one or just the jonasnick fixups?

jonasnick:
just fixups

nixbitcoin:
squashed

jonasnick:
Oh I meant squashing the fixups into the corresponding original commits.

nixbitcoin:
Done. Didn't squash `clightning module: remove config group read access` because it is a separate hardening, but I moved it to a more appropriate place.

erikarvstedt:
I'll review this later today, please wait before merging.

erikarvstedt:
Thanks, @nixbitcoin, for these very sensible changes.

Could you, in the future, add a `fixup!` prefix to titles of commits that are meant to be squashed?

Also, please use separate commits for larger, orthogonal changes.
For example `lightning-charge: Update pkg and run under lightning-charge user` consists of these single-topic commits:
   - lightning-charge: 0.4.14 -> 0.4.19
   - lightning-charge: add dedicated user
   - clightning: allow group access to RPC socket 

Here are [my fixups](https://github.com/erikarvstedt/nix-bitcoin/commits/hardening-DAC).

modules/presets/secure-node.nix
nixbitcoin: @erikarvstedt Re: 0e66ec47489732002ba9ab96dafd3ddd401a5ae8 I thought it would be good to keep, so `lightning-cli` would never be executed as root by accident. But I guess the simplicity win outweighs that. So ACK.

nixbitcoin:
Thank you for the great `fixup !`'s  :smile: I squashed them into the appropriate commits.

This was my first try at structuring commits properly. My next PR will be more diligent in separating into single-topic commits and obeying the nix naming conventions.

I separated said commit. I didn't separate 95afba4 because the changes aren't that orthogonal and splitting wasn't that easy.

erikarvstedt:
Thanks! :smiley_cat:
[These fixups](https://github.com/erikarvstedt/nix-bitcoin/commits/hardening-DAC) remain.

Could you please remove the ephemeral fixup messages from the squashed messages?
The squashed messages should only contain information relevant to the committed content. Historical discussion about ultimately rejected and uncommited changes should not be part of the commit message, unless it provides an insightful rationale for the actual changes.

#### Example
Problematic message:
```
nodeinfo: Convert to module and allow alternative operator username

operatorname: make into option in secure-node.nix

@jonasnick fixup

nodeinfo module: fix whitespace & remove unused

@jonasnick fixup

fixup! rename services.operatorname -> nix-bitcoin.operatorName

operatorName is not a service

fixup! improve nodeinfo

rename services.nodeinfo -> program.nodeinfo, following NixOS naming
conventions. (nodeinfo is not a service.)

currently, nodeinfo has presets/secure-node.nix as a strict
dependency as it requires onion-chef and the 'operatorName' option.
and nix-bitcoin-webindex.nix has nodeinfo as a dependecy.

so don't add nodeinfo and webindex to modules.nix because they will fail on standalone use.
```
Fixed message:
```
nodeinfo: Convert to module and allow custom operator username

Currently, nodeinfo has presets/secure-node.nix as a strict
dependency as it requires onion-chef and the 'operatorName' option.
Also, nix-bitcoin-webindex.nix has nodeinfo as a dependecy.

So don't add nodeinfo to modules.nix (and remove webindex)
because they both fail on standalone use.
```

erikarvstedt:
Please rename commit `clightning module: remove ...` -> `clightning: remove`

nixbitcoin:
@erikarvstedt 

Could you format your approvals in the future like so: `ACK <top commit hash>`. That way it shows up in @jonasnick's merge commits.

erikarvstedt:
@nixbitcoin ACK :smiley:

PR #174, opened 2020-05-06
Hardening systemd

nixbitcoin:
This PR is based on #171 and includes hardenings via systemd. 

All modules, with exceptions where absolutely necessary, now have `CapabilityBoundingSets`, `ProtectSystem = strict`, `PrivateUsers`, `RestrictSUIDSGID`, disallowed `@clock`, `@cpu-emulation` and `@resources` syscalls, `RemoveIPC`, `RestrictRealtime`, `ProtectHostname`.

`ProtectSystem = strict` makes it impossible to create dataDir's that act as `ReadWritePaths` in `preStart`. Seeing as `PermissionsStartOnly`, which is required for directory creation via `PreStart`, is being deprecated anyway (NixOS/nixpkgs/issues/53852), I made the switch to creating initial data directories with `systemd.tmpfiles.rules` (NixOS/nixpkgs#56265). We should completely switch to `systemd.tmpfiles.rules` directory creation in the future before `PermissionsStartOnly` gets deprecated.

Sample `systemd-analyze security bitcoind.service` "Overall exposure level for bitcoind.service" after this PR 1.2 (!!)

Note: All my commits starting at b97eec9dee5a19d82905dede3689f239b122d9f0 will be @opentimestamps timestamped.


jonasnick:
> We should completely switch to systemd.tmpfiles.rules directory creation in the future before PermissionsStartOnly gets deprecated.

I don't know if this is just a documentation/naming issue, but `man tmpfiles.d` clearly states

> tmpfiles.d - Configuration for creation, deletion and cleaning of volatile and temporary files

But our data directories are not temporary, nor volatile. If I understand correctly, the directories would be immediately deleted when the service is disabled. So if you make a mistake while experimenting with your configuration.nix you may find your wallets, Lightning channeldbs, etc suddenly gone. To me that seems to be quite a drawback. The bitcoind module in nixpkgs also went the tmpfiles route. Remind me, why are we not using systemd's StateDirectory again?

erikarvstedt:
`tmpfiles` was a misnomer on part of systemd, it's also useful for setting up persistent files.

`systemd.tmpfiles` is now the preferred way to setup service-related files/dirs in NixOS. Search for `systemd.tmpfiles.rules` in the NixOS source to see some examples.

nixbitcoin:
Confirmed that disabling a service does _not_ delete the tmpfiles-created directory by turning `lightning-charge` and `nanopos` off in `configuration.nix` and deploying. `/var/lib/lightning-charge` persisted.

jonasnick:
Happy to hear that tmpfiles are not deleted. That's quite a big misnomer then.

Kixunil:
Why `StateDirectory` is not used though? Seems much clearer and I would love to know if there's a good reason I should do the same in my project.

nixbitcoin:
Fixed issues that @jonasnick submitted to me in private correspondence.

nixbitcoin:
> Why `StateDirectory` is not used though? Seems much clearer and I would love to know if there's a good reason I should do the same in my project.

I implemented `systemd.tmpfiles` because nixpkgs uses `systemd.tmpfiles`.

@erikarvstedt Why does nixpkgs use `systemd.tmpfiles`?

From the `systemd.exec` manpage:

"tmpfiles.d(5) provides functionality that overlaps with these options. Using these options is recommended, because the lifetime of the directories is tied directly to the lifetime of the unit, and it is not necessary to ensure that the tmpfiles.d configuration is executed before the unit is started."

erikarvstedt:
@nixbitcoin because its declarative design makes it more secure and robust than privileged preStart scripts. NixOS takes care of the lifetime issues by always setting up tmpfiles on boot and when switching configurations.

erikarvstedt:
@nixbitcoin, @Kixunil Addendum: For services where the data dir path is not configurable, NixOS indeed uses `stateDirectory`, which prescribes a `/var/lib` path prefix. 
In nix-bitcoin, we could use `stateDirectory` for services that have only trivial (say, ≤50 MiB) storage demands.

nixbitcoin:
@erikarvstedt Can you review this PR so we can merge?

erikarvstedt:
Again, great improvements! :smiley_cat:

Here are some [minor fixups](https://github.com/erikarvstedt/nix-bitcoin/commits/hardening-systemd-ea).

* Regarding my commit `fixup! All modules with preStart: Use systemd.tmpfiles.rules`:
  We should reintroduce the `bitcoin-rpc` group so that we can rid of `PermissionsStartOnly` in clightning, electrs, liquid and lnd.

* Could you rename these commit descriptions?
  `All modules: Make serviceConfig precedent over nix-bitcoin-services.nix` ->
  `All modules: Give service config precedence over defaultHardening`

  `nix-bitcoin-services: Add PrivateUsers` ->
  `service hardening: Add PrivateUsers`


* Could you merge these single-line commits into one commit `service hardening: add more restrictions`
  ```
  Add ProtectHostname
  Add RestrictRealtime
  Add RemoveIPC
  Add RestrictSUIDSGID
  ```

* In description of commit `All modules with preStart: Use systemd.tmpfiles.rules`

  Please rephrase paragraph
  `The supported ...` ->
  `This is NixOS' recommended way to setup service dirs https://github.com/NixOS/nixpkgs/pull/56265`

  The paragraph `Further on we might have...` is incorrect and can be deleted:
  We can still use `preStart` with root permissions by prefixing executables with `!`
  ```
  ExecStartPre = "!${rootExecutable} ; ${nonRootExecutable} ; ...";
  ```

* Regarding `SystemCallFilter`: This massive set of syscalls, with zero rationale in comments, looks suspicious and hard to audit.
  I'd propose to replace it with just `@system-service`. This is also the recommended setting in systemd's [strict security profile](https://github.com/systemd/systemd/blob/master/src/portable/profile/strict/service.conf).

  Here the [set differences](https://gist.github.com/b5665da8cda24dc020650130e0eaef49) when comparing `@system-service` with the current value from this PR. [Here's the script](https://gist.github.com/977d3a08275f1f3ce171157a47a5c35e) that produced this output.
  The extra system calls pulled in by `@system-service` are entirely benign when used in combination with the remaining hardening settings.

Kixunil:
@erikarvstedt thanks, I understand now!

nixbitcoin:
@erikarvstedt Your fixups and comments are truly appreciated! :smile_cat:

> Here are some minor fixups.
>     * Regarding my commit `fixup! All modules with preStart: Use systemd.tmpfiles.rules`:
>       We should reintroduce the `bitcoin-rpc` group so that we can get rid of `PermissionsStartOnly` in clightning, electrs, liquid and lnd.

First three fixups squashed into the corresponding commits. Fourth fixup 65e46d9b61f786f919efd30643c74f5492e72b24 developed into its own commit d0fba3a 

>     * Could you rename these commit titles?
>       `All modules: Make serviceConfig precedent over nix-bitcoin-services.nix` ->
>       `All modules: Give service config precedence over defaultHardening`
>       `nix-bitcoin-services: Add PrivateUsers` ->
>       `service hardening: Add PrivateUsers`
> 
>     * Could you merge these single-line commits into one commit `service hardening: add more restrictions`
>       ```
>       Add ProtectHostname
>       Add RestrictRealtime
>       Add RemoveIPC
>       Add RestrictSUIDSGID
>       ```
>
>     * In description of commit `All modules with preStart: Use systemd.tmpfiles.rules`
>       Please rephrase paragraph
>       `The supported ...` ->
>       `This is NixOS' recommended way to setup service dirs https://github.com/NixOS/nixpkgs/pull/56265`
>       The paragraph `Further on we might have...` is incorrect and can be deleted:
>       We can still use `preStart` with root permissions by prefixing executables with `!`:
>       ```
>       ExecStartPre = "!${rootExecutable} ; ${nonRootExecutable} ; ...";
>       ```

Done

>     * Regarding `SystemCallFilter`: This massive set of syscalls, with zero rationale in comments, looks suspicious and hard to audit.
>       I'd propose to replace it with just `@system-service`. This is also the recommended setting in systemd's [strict security profile](https://github.com/systemd/systemd/blob/master/src/portable/profile/strict/service.conf).
>       Here are the [set differences](https://gist.github.com/erikarvstedt/40bacc9af927e3158fb1420ed9669d60) when comparing `@system-service` to the current value from this PR. [Here's the script](https://gist.github.com/977d3a08275f1f3ce171157a47a5c35e) that produced this output.
>       The extra system calls pulled in by `@system-service` are entirely benign when used in combination with the remaining hardening settings.

Implemented in 8f088b4 
I added the syscalls included in the `@system-service` whitelist but excluded in https://docs-stage.docker.com/engine/security/seccomp/#significant-syscalls-blocked-by-the-default-profile to a separate blacklist.

nixbitcoin:
I was having problems with the `elements.conf` because at some point it had been generated with read-only user permissions. Had to delete the file and run again with `chmod 640` in `preStart`. Should we create all config files with `systemd.tmpfiles`?

erikarvstedt:
Good judgement with the fixups and adding a new commit! But can you set yourself as the author of `Remove PermissionStartOnly ...`?

The lnd test fails because it needs its dataDir created with systemd-tmpfiles.

What's the problem exactly with `elements.conf`?  The liquidd test succeds. The preStart script creates the file with mode `640`, so how did you end up with mode `400`?

nixbitcoin:
> The lnd test fails because it needs its dataDir created with systemd-tmpfiles.

I think we can't remove `PermissionsStartOnly` because lnd also needs it in the postStart for copying the lnd seed mnemonic.

> What's the problem exactly with `elements.conf`? The liquidd test succeds. The preStart script creates the file with mode `640`, so how did you end up with mode `400`?

At some point I must have changed it manually, and preStart no longer had the permissions to override it. Hope this was only an anomaly for me. @jonasnick How does yours look?

erikarvstedt:
Please rename `bitcoind: Only set zmqoptions if cfg.lnd.enable` -> `Only enable bitcoind zmqpub if lnd.enable` (because this mainly affects lnd, not bitcoind.)

> because lnd also needs it in

Ah, right. [Fixed here](https://github.com/erikarvstedt/nix-bitcoin/commits/hardening-systemd).

erikarvstedt:
I've updated my lnd fix.

nixbitcoin:
> Please rename `bitcoind: Only set zmqoptions if cfg.lnd.enable` -> `Only enable bitcoind zmqpub if lnd.enable` (because this change is pushed by lnd, not bitcoind.)

Done. Prepended topic `lnd:`



> Ah, right. [Fixed here](https://github.com/erikarvstedt/nix-bitcoin/commits/hardening-systemd).

Cherry pick your commits, squashed the first one, and put them in the right place. Left you as an author for second commit.

erikarvstedt:
You missed my last [change](https://github.com/erikarvstedt/nix-bitcoin/commits/hardening-systemd) by a few minutes.
Please replace commit `lnd: remove PermissionStartOnly` with `lnd: add strict hardening`, and make sure this commit ends up after `All modules: ProtectSystem = strict` so that the commit title makes sense.

erikarvstedt:
Thanks. But by changing my commit (`lnd: add strict hardening`), the commit message got outdated. Please use this [message](https://gist.github.com/036337e8feb67be1817953b2032847e4).

erikarvstedt:
Alright, [one last change](https://github.com/erikarvstedt/nix-bitcoin/commits/hardening-systemd) and we're good.

nixbitcoin:
@erikarvstedt I think that rounds it off nicely. Ready for merge?

PR #177, opened 2020-05-16
Fix shell session in qemu example

jonasnick:
Otherwise when running the c() function, the variables for the ssh
command are empty.


erikarvstedt:
I've made a more general and robust fix: https://github.com/fort-nix/nix-bitcoin/pull/178

PR #178, opened 2020-05-17
examples: execute bash sessions in script environment

erikarvstedt:
Previously, the sessions contained only explicitly exported variables and functions.
This was fragile and in part buggy due to lacking exports.

Interactive features like user-defined aliases and functions are still working as before.


erikarvstedt:
It's too restrictive. You can't inspect the VM, can't copy files to/from the node, etc.
The usability is improved in the latest commit.

PR #179, opened 2020-05-17
Add BTCPayServer

jb55:
This is a simplified version of #88, with an update script this time!


pkgs/btcpayserver/dotnet-build.nix
```diff
@@ -0,0 +1,182 @@
...
+  , project
+  , src
+  , meta
+  , nugetPackages ? null
+  , installExtra ? ""
+  , path ? "./"
+  , config ? "Release"
+  , target ? "linux-x64", ... }:
+    let
+      nuget-pkg-json = if nugetPackages == null
```
nixbitcoin, line 113: trailing whitespace

pkgs/btcpayserver/dotnet-build.nix
```diff
@@ -0,0 +1,182 @@
...
+  , src
+  , meta
+  , nugetPackages ? null
+  , installExtra ? ""
+  , path ? "./"
+  , config ? "Release"
+  , target ? "linux-x64", ... }:
+    let
+      nuget-pkg-json = if nugetPackages == null 
+                          then lib.importJSON (lesrc + "/${path}${project}/nuget-packages.json")
```
nixbitcoin, line 114: trailing whitespace

nixbitcoin:
1. Ran `./update.sh`. This gives me wildly different `btcpayserver-packages.json` and `nbxplorer-packages.json`. Diffs available [here](https://gist.github.com/nixbitcoin/e0c5501c3b248fa48b03b5ae6e92555d) and [here](https://gist.github.com/nixbitcoin/cffea2305bd3518c9dc65c2828adedfb).

2. Successfully built with `nix-build -E 'with import <nixpkgs> {}; callPackage ./btcpayserver.nix {}'` and `nix-build -E 'with import <nixpkgs> {}; callPackage ./nbxplorer.nix {}'` in our nix-shell

3. Started built binaries and checked BTCPayserver web interface was reachable

Making this into a module should be as easy as running both nbxplorer and btcpayserver with systemd and letting the user configure with the web-interface. We should also provide an easy option to switch between a Tor-only deployment and a clearnet-allowed deployment. In the case of clearnet deployments all outgoing connections like price-queries should still be forwarded over the Tor proxy.

pkgs/btcpayserver/update
```diff
@@ -0,0 +1,105 @@
...
+  dotnet2nix "$@"
+
+  cp nuget-packages.json "$here"/"$repo"-packages.json
+
+  ver=$rev
+  if [ "${rev::1}" == "v" ]; then
+    ver=${rev:1}
+  fi
+
+  sha256=$(nix-prefetch-github "$owner" "$repo" --rev "$rev" | jq -r .sha256)
```
nixbitcoin, line 81: What do you think about using fetchurl and calculating the sha256 with `git archive` from Nicolas Dorier's signed git commits (0x6618763EF09186FE)? Similar to `pkgs/electrs/default.nix` and `pkgs/electrs/get-sha256.sh`.

Kixunil:
@nixbitcoin feel free to take inspiration from my project. I have onion-by-default, clearnet-optional implemented already. It can handle other apps (like RTL) too. However, it's still not a privacy feature (just to get a domain and bypass NAT). Accesses are not torified yet.

pkgs/btcpayserver/update
jb55: oh yeah this would be easy to do, I think I had this for testing git revisions. We should just be doing nix-prefetch-url with signed tarballs instead

jb55:
> @nixbitcoin  Ran ./update.sh. This gives me wildly different btcpayserver-packages.json and nbxplorer-packages.json. Diffs available here and here.

I noticed this as well, I don't remember it being like this before, it's possible I introduced some non-deterministic parallelism in dotnet2nix somehow. I'll try to see what's up.

PR #187, opened 2020-05-29
Network Namespaces

nixbitcoin:
Network Namespaces are a significant security improvement, mitigating the bind race leak security vulnerability and unnecessary communications between nix-bitcoin services. Network namespaces round-off our service isolation to cover all levels of inter-service interaction.

Special thank you to @erikarvstedt for generalizing the `makeNetnsServices` :rocket:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,8 @@ env:
     # CACHIX_SIGNING_KEY
   - secure: "xXCFZ7g+k5YmCGm8R8l3bZElVmt+RD1KscG3kGr5w4HyyDPTzFetPo+sT8bUpysDU0u3HWhfVhHtpog2mhNhwVl3tQwKXea3dHKC1i6ypBg3gjDngmJRR5wo++ocYDpK8qPaU7m/jHQTNFnTA4CbmMcc05GcYx/1Ai/ZGkNwWFjdIcVeOUoiol33gykMOXIGDg2qlXudt33wP53FHbX8L4fxzodWfAuxKK4AoGprxy5eSnU7LCaXxxJmu4HwuV+Ux2U1NfE/E33cvhlUvTQCswVSZFG06mg8rwhMG1ozsDvlL2itZlu/BeUQH5y3XMMlnJIUXUazkRBibf1w/ebVjpOF+anqkqmq8tcbFEa7T+RJeVTIsvP+L8rE8fcmuZtdg9hNmgRnLmaeT0vVwD1L2UqW9HdRyujdoS0jPYuoc1W7f1JQWfAPhBPQ1SrtKyNNqcbVJ34aN7b+4vCzRpQL1JTbmjzQIWhkiKN1qMo1v/wbIydW8yka4hc4JOfdQLaAJEPI1eAC1MLotSAegMnwKWE1dzm66MuPSipksYjZrvsB28cV4aCVUffIuRhrSr1i2afRHwTpNbK9U4/576hah15ftUdR79Sfkcoi1ekSQTFGRvkRIPYtkKLYwFa3jVA41qz7+IIZCf4TsApy3XDdFx91cRub7yPq9BeZ83A+qYQ="
   jobs:
-  - TestModules=1 STABLE=1
+  - TestModules=1 STABLE=1 SCENARIO=default
+  - TestModules=1 STABLE=1 SCENARIO=withnetns
   - PKG=hwi STABLE=1
   - PKG=hwi STABLE=0
   - PKG=lightning-charge STABLE=1
@@ -38,7 +39,7 @@ script:
                 exit 1
             fi
             sudo chmod go+rw /dev/kvm
-            test/run-tests.sh exprForCI
+            test/run-tests.sh --scenario $SCENARIO exprForCI
         else
             echo "(import ./. {}).$PKG"
         fi
--- a/modules/bitcoind.nix
+++ b/modules/bitcoind.nix
@@ -18,6 +18,7 @@ let
     ${optionalString (cfg.assumevalid != null) "assumevalid=${cfg.assumevalid}"}
 
     # Connection options
+    ${optionalString cfg.listen "bind=${cfg.bind}"}
     ${optionalString (cfg.port != null) "port=${toString cfg.port}"}
     ${optionalString (cfg.proxy != null) "proxy=${cfg.proxy}"}
     listen=${if cfg.listen then "1" else "0"}
@@ -30,6 +31,8 @@ let
       (rpcUser: "rpcauth=${rpcUser.name}:${rpcUser.passwordHMAC}")
       (attrValues cfg.rpc.users)
     }
+    ${lib.concatMapStrings (rpcbind: "rpcbind=${rpcbind}\n") cfg.rpcbind}
+    ${lib.concatMapStrings (rpcallowip: "rpcallowip=${rpcallowip}\n") cfg.rpcallowip}
     ${optionalString (cfg.rpcuser != null) "rpcuser=${cfg.rpcuser}"}
     ${optionalString (cfg.rpcpassword != null) "rpcpassword=${cfg.rpcpassword}"}
 
@@ -67,6 +70,13 @@ in {
         type = types.path;
         default = "/var/lib/bitcoind";
         description = "The data directory for bitcoind.";
+      };
+      bind = mkOption {
+        type = types.str;
+        default = "127.0.0.1";
+        description = ''
+          Bind to given address and always listen on it.
+        '';
       };
       user = mkOption {
         type = types.str;
@@ -116,6 +126,20 @@ in {
             RPC user information for JSON-RPC connnections.
           '';
         };
+      };
+      rpcbind = mkOption {
+        type = types.listOf types.str;
+        default = [ "127.0.0.1" ];
+        description = ''
+          Bind to given address to listen for JSON-RPC connections.
+        '';
+      };
+      rpcallowip = mkOption {
+        type = types.listOf types.str;
+        default = [ "127.0.0.1" ];
+        description = ''
+          Allow JSON-RPC connections from specified source.
+        '';
       };
       rpcuser = mkOption {
           type = types.nullOr types.str;
--- a/modules/bitcoind.nix
+++ b/modules/bitcoind.nix
@@ -257,11 +257,20 @@ in {
       };
       cli = mkOption {
         type = types.package;
+        default = cfg.cli-nonetns-exec;
+        description = "Binary to connect with the bitcoind instance.";
+      };
+      # Needed because bitcoind-import-banlist already executes inside
+      # nb-bitcoind, hence it doesn't need netns-exec prefixed.
+      cli-nonetns-exec = mkOption {
         readOnly = true;
+        type = types.package;
         default = pkgs.writeScriptBin "bitcoin-cli" ''
           exec ${cfg.package}/bin/bitcoin-cli -datadir='${cfg.dataDir}' "$@"
         '';
-        description = "Binary to connect with the bitcoind instance.";
+        description = ''
+          Binary to connect with the bitcoind instance without netns-exec.
+        '';
       };
       enforceTor =  nix-bitcoin-services.enforceTor;
     };
@@ -321,7 +330,7 @@ in {
       bindsTo = [ "bitcoind.service" ];
       after = [ "bitcoind.service" ];
       script = ''
-        cd ${cfg.cli}/bin
+        cd ${cfg.cli-nonetns-exec}/bin
         # Poll until bitcoind accepts commands. This can take a long time.
         while ! ./bitcoin-cli getnetworkinfo &> /dev/null; do
           sleep 1
--- a/modules/netns-isolation.nix
+++ b/modules/netns-isolation.nix
@@ -85,6 +85,9 @@ in {
       };
 
       nix-bitcoin.netns-isolation.services = {
+        bitcoind = {
+          id = 12;
+        };
       };
 
       systemd.services = {
@@ -107,6 +110,8 @@ in {
             RemainAfterExit = "yes";
           };
         };
+
+        bitcoind-import-banlist.serviceConfig.NetworkNamespacePath = "/var/run/netns/nb-bitcoind";
       } //
       (let
         makeNetnsServices = n: v: let
@@ -158,6 +163,20 @@ in {
       in foldl (services: n:
         services // (makeNetnsServices n netns.${n})
       ) {} (builtins.attrNames netns));
+
+      # bitcoin: Custom netns configs
+      services.bitcoind = {
+        bind = netns.bitcoind.address;
+        rpcbind = [
+          "${netns.bitcoind.address}"
+          "127.0.0.1"
+        ];
+        rpcallowip = [
+          "127.0.0.1"
+        ] ++ lib.lists.concatMap (s: [
+          "${netns.${s}.address}"
+        ]) netns.bitcoind.availableNetns;
+      };
 
     })
     # Custom netns config option values if netns-isolation not enabled
--- a/modules/netns-isolation.nix
+++ b/modules/netns-isolation.nix
@@ -176,6 +176,9 @@ in {
         ] ++ lib.lists.concatMap (s: [
           "${netns.${s}.address}"
         ]) netns.bitcoind.availableNetns;
+        cli = pkgs.writeScriptBin "bitcoin-cli" ''
+          netns-exec nb-bitcoind ${config.services.bitcoind.package}/bin/bitcoin-cli -datadir='${config.services.bitcoind.dataDir}' "$@"
+        '';
       };
 
     })
--- a/modules/netns-isolation.nix
+++ b/modules/netns-isolation.nix
@@ -96,6 +96,10 @@ in {
           id = 14;
           connections = [ "bitcoind" ];
         };
+        liquidd = {
+          id = 15;
+          connections = [ "bitcoind" ];
+        };
       };
 
       systemd.services = {
@@ -213,6 +217,21 @@ in {
           netns-exec nb-lnd sudo -u lnd ${config.services.lnd.package}/bin/lncli --tlscertpath ${config.nix-bitcoin.secretsDir}/lnd-cert \
             --macaroonpath '${config.services.lnd.dataDir}/chain/bitcoin/mainnet/admin.macaroon' "$@"
         '';
+      };
+
+      # liquidd: Custom netns configs
+      services.liquidd = mkIf config.services.liquidd.enable {
+        bind = netns.liquidd.address;
+        rpcbind = [
+          "${netns.liquidd.address}"
+          "127.0.0.1"
+        ];
+        rpcallowip = [
+          "127.0.0.1"
+        ] ++ lib.lists.concatMap (s: [
+          "${netns.${s}.address}"
+        ]) netns.liquidd.availableNetns;
+        mainchainrpchost = netns.bitcoind.address;
       };
 
     })
--- a/modules/netns-isolation.nix
+++ b/modules/netns-isolation.nix
@@ -232,6 +232,12 @@ in {
           "${netns.${s}.address}"
         ]) netns.liquidd.availableNetns;
         mainchainrpchost = netns.bitcoind.address;
+        cli = pkgs.writeScriptBin "elements-cli" ''
+          netns-exec nb-liquidd ${pkgs.nix-bitcoin.elementsd}/bin/elements-cli -datadir='${config.services.liquidd.dataDir}' "$@"
+        '';
+        swap-cli = pkgs.writeScriptBin "liquidswap-cli" ''
+          netns-exec nb-liquidd ${pkgs.nix-bitcoin.liquid-swap}/bin/liquidswap-cli -c '${config.services.liquidd.dataDir}/elements.conf' "$@"
+        '';
       };
 
     })
--- a/modules/netns-isolation.nix
+++ b/modules/netns-isolation.nix
@@ -115,6 +115,10 @@ in {
           # communicates with clightning over lightning-rpc socket
           connections = [];
         };
+        nanopos = {
+          id = 19;
+          connections = [ "nginx" "lightning-charge" ];
+        };
       };
 
       systemd.services = {
@@ -270,6 +274,12 @@ in {
 
       # lightning-charge: Custom netns configs
       services.lightning-charge.host = mkIf config.services.lightning-charge.enable netns.lightning-charge.address;
+
+      # nanopos: Custom netns configs
+      services.nanopos = mkIf config.services.nanopos.enable {
+        charged-url = "http://${netns.lightning-charge.address}:9112";
+        host = netns.nanopos.address;
+      };
 
     })
     # Custom netns config option values if netns-isolation not enabled
--- a/modules/netns-isolation.nix
+++ b/modules/netns-isolation.nix
@@ -92,6 +92,10 @@ in {
           id = 13;
           connections = [ "bitcoind" ];
         };
+        lnd = {
+          id = 14;
+          connections = [ "bitcoind" ];
+        };
       };
 
       systemd.services = {
@@ -189,6 +193,20 @@ in {
       services.clightning = mkIf config.services.clightning.enable {
         bitcoin-rpcconnect = netns.bitcoind.address;
         bind-addr = "${netns.clightning.address}:${toString config.services.clightning.onionport}";
+      };
+
+      # lnd: Custom netns configs
+      services.lnd = mkIf config.services.lnd.enable {
+        listen = netns.lnd.address;
+        rpclisten = [
+          "${netns.lnd.address}"
+          "127.0.0.1"
+        ];
+        restlisten = [
+          "${netns.lnd.address}"
+          "127.0.0.1"
+        ];
+        bitcoind-host = netns.bitcoind.address;
       };
 
     })
--- a/modules/netns-isolation.nix
+++ b/modules/netns-isolation.nix
@@ -119,6 +119,11 @@ in {
           id = 19;
           connections = [ "nginx" "lightning-charge" ];
         };
+        recurring-donations = {
+          id = 20;
+          # communicates with clightning over lightning-rpc socket
+          connections = [];
+        };
       };
 
       systemd.services = {
--- a/modules/netns-isolation.nix
+++ b/modules/netns-isolation.nix
@@ -124,6 +124,10 @@ in {
           # communicates with clightning over lightning-rpc socket
           connections = [];
         };
+        nginx = {
+          id = 21;
+          connections = [];
+        };
       };
 
       systemd.services = {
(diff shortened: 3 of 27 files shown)


Kixunil:
Regarding 4) I solved a similar problem for `lncli` by writing [a wrapper script](https://github.com/Kixunil/cryptoanarchy-deb-repo-builder/blob/master/lnd/xlncli) and named it `lncli` using `update-alternatives`. I plan to do the same for `bitcoin-cli` Maybe a similar approach can work here?

modules/netns-chef.nix
```diff
@@ -0,0 +1,257 @@
...
+      foldl (xs: s2:
+        xs // { "${s2}" = xs.${s2} ++ [ s1 ]; }
+      ) xs cfg.services.${s1}.connections
+    ) base (builtins.attrNames base);
+
+  enabledServices = filterAttrs (n: v: isEnabled n) cfg.services;
+  isEnabled = x: config.services.${x}.enable;
+
+  ip = "${pkgs.iproute}/bin/ip";
+  iptables = "${pkgs.iptables}/bin/iptables";
```
mmilata, line 35: Consider using `config.networking.firewall.package` instead for nftables. (Though if nixos/nixpkgs#81172 should make this uneccessary.)

modules/netns-chef.nix
```diff
@@ -0,0 +1,257 @@
...
+  };
+
+  config = mkIf cfg.enable {
+    # Prerequisites
+    networking.dhcpcd.extraConfig = "noipv4ll";
+    services.tor.client.socksListenAddress = "169.254.${toString cfg.addressblock}.10:9050";
+    networking.firewall.interfaces.br0.allowedTCPPorts = [ 9050 ];
+    boot.kernel.sysctl."net.ipv4.ip_forward" = true;
+
+    nix-bitcoin.netns-isolation.services = {
```
mmilata, line 70: For the sake of modularity it might be preferable to define the implementation in this file and then set the concrete mapping and other parameters elsewhere, e.g. in secure-node.nix?

nixbitcoin:
> Regarding 4) I solved a similar problem for `lncli` by writing [a wrapper script](https://github.com/Kixunil/cryptoanarchy-deb-repo-builder/blob/master/lnd/xlncli) and named it `lncli` using `update-alternatives`. I plan to do the same for `bitcoin-cli` Maybe a similar approach can work here?

We've used wrappers before, check out [this](https://github.com/fort-nix/nix-bitcoin/blob/master/modules/bitcoind.nix#L238). The problem is that `ip netns exec` requires elevated privileges, which I don't think is a good idea to give the `operator` user. If the `operator` user can run `ip netns exec` with root privileges, he can run any following command with root privileges.

I've been thinking about how to solve 4. and I think the best idea is to create a separate `nb-operator` namespace with access to all other namespaces. All the user scripts will automatically execute with the right flags, like `-rpcconnect` for `bitcoin-cli`. When the operator user ssh's into the box he's automatically placed into a shell inside the `nb-operator` network namespace. Thoughts?

modules/netns-chef.nix
```diff
@@ -0,0 +1,257 @@
...
+      bitcoind = {
+        id = 12;
```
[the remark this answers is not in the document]
nixbitcoin: Why would you want to do that? It makes everything simpler IMO to just have hardcoded addresses in a changeable addressblock.

modules/netns-chef.nix
nixbitcoin: Is nftables better than iptables for what we're trying to do?

modules/netns-chef.nix
nixbitcoin: @jonasnick and I talked about restructuring this file like so: parameters in each individual module, mapping in `secure-node.nix` and general implementation in `netns-chef.nix`. What are your thoughts on this approach?

modules/netns-chef.nix
mmilata: Sorry, I meant to write "... instead for *compatibility with* nftables". Using `config.networking.firewall.package` instead of a concrete package would make it work regardless of whether the system is set to use iptables or nftables.

IMO nftables is the future but both work fine for this use case.

modules/netns-chef.nix
mmilata: Sounds reasonable!

mmilata:
What about [setcap wrapper](https://nixos.org/nixos/options.html#security.wrappers)? Namespaces require CAP_SYS_ADMIN which is [pretty much root](https://lwn.net/Articles/486306/) but I guess it's marginally better than the sudo rule and the capability won't be inherited by the program passed to `ip netns exec` as an argument?

erikarvstedt:
Please rename `netns-chef.nix` -> `netns-isolation.nix`. chef is amusing in the context of onions, but it's inappropriate here.

modules/netns-chef.nix
```diff
@@ -0,0 +1,146 @@
...
+
+  ip = "${pkgs.iproute}/bin/ip";
+  iptables = "${pkgs.iptables}/bin/iptables";
+
+in {
+  options.nix-bitcoin.netns-isolation = {
+    enable = mkEnableOption "netns isolation";
+
+    addressblock = mkOption {
+      type = types.str;
```
jonasnick, line 42: Why not `type.u8`?

modules/netns-chef.nix
```diff
@@ -0,0 +1,146 @@
...
+      description = ''
+        Specify the N address block in 169.254.N.0/24.
+      '';
+    };
+
+    services = mkOption {
+      default = {};
+      type = types.attrsOf (types.submodule {
+        options = {
+          id = mkOption { type = types.int; };
```
jonasnick, line 53: Description would be helpful. "must be unique", "must not be 10"

modules/netns-chef.nix
```diff
@@ -0,0 +1,146 @@
...
+            default = [];
+          };
+        };
+      });
+    };
+  };
+
+  config = mkIf cfg.enable {
+    # Prerequisites
+    networking.dhcpcd.extraConfig = "noipv4ll";
```
jonasnick, line 65: Would be nice to have a comment here explaining why this is needed. Couldn't we use dhcpcd's `denyinterfaces` option to ignore the interfaces altogether?

modules/netns-chef.nix
```diff
@@ -0,0 +1,146 @@
...
+    services.tor.client.socksListenAddress = "169.254.${toString cfg.addressblock}.10:9050";
```
jonasnick, line 66: Better `let bridgeIp = "...";` because it's used quite a few times in this file.

modules/bitcoind.nix
```diff
@@ -117,6 +119,27 @@ in {
           '';
         };
       };
+      rpcbind = mkOption {
+        type = types.lines;
```
jonasnick, line 123: Would be more consistent and simpler if this was just a list of ip addresses (same with `rpcallowip` and in the liquid module)

jonasnick:
It would be helpful to document how this works and in particular how this can be customized to allow access through clearnet or wireguard interfaces. As an example here's the diagram I sketched while trying to understand what is happening.

crappy-diagram

modules/netns-chef.nix
nixbitcoin: Because we literally only change the N in `169.254.N.0/24`

nixbitcoin:
I force pushed a new and improved HEAD that includes lnd, an operator namespace, and several fixes. It also removes the clightning autolisten option, as documented in the commit.

These todos remain:

I've tried implementing these and have gotten errors that I can't solve (please help @erikarvstedt @jonasnick):
1. Restructure the custom netns configs so they are only set when the individual services are actually enabled.
2. Make `bitcoind-import-banlist` run in the `nb-bitcoind namespace`.

I need your conceptual input on these:
1. Think about placing custom netns configs in each individual module, mapping in `secure-node.nix` and general implementation in `netns-isolation.nix`.
2. Drop the operator user into a shell inside `nb-operator` and give him aliases that include flags like `rpcconnect`. Something like `ip netns exec nb-operator sudo -u operator bash` should do.
3. Mitigate or document LND's breaking changes.
4. Find a better way to amend the LND cert than hardcoding the ip address.
5. Update the testing framework to test with and without network namespaces.

I can take care of these but they depend on the above:
1. Document clearnet and wireguard access.

modules/netns-chef.nix
nixbitcoin: See `fixup! architecture: pkgs.iptables -> config.networking.firewall.package`

modules/netns-chef.nix
nixbitcoin: See `fixup! id option: add description`

modules/netns-chef.nix
nixbitcoin: See `fixup! dhcpcd: noipv4ll -> denyinterfaces`

modules/netns-chef.nix
nixbitcoin: See `fixup! let: add bridgeIp` and `fixup! let: add operatorIp`

modules/bitcoind.nix
nixbitcoin: See `fixup! liquid: make rpcbind and rpcallowip lists of ip addresses` and `fixup! bitcoind: make rpcbind and rpcallowip lists of ip addresses`

nixbitcoin:
I got rid of the operator shell and replaced it with custom cli scripts that use Jonas Nick's `netns-exec` to execute `bitcoin-cli`, `lncli`, and `elements-cli` in the appropriate network namespaces. Operator can use `netns-exec` because of a `security.wrappers` program with `cap_sys_admin`.

I also rewrote the `test_script.py` to use netns-isolation. If we want to test without netns-isolation we need to write a second `test_script.py`. Tests run nicely now.

Only remaining issues are the breaking changes to `lnd` (hardcoded ip address in cert and new hidden service).

Very close to mergeable IMO.

nixbitcoin:
Fixed some nits.

test/test-script.py
```diff
@@ -38,13 +38,13 @@ def assert_running(unit):
 assert_matches("su operator -c 'bitcoin-cli getnetworkinfo' | jq", '"version"')
 
 assert_running("electrs")
-machine.wait_for_open_port(4224)  # prometeus metrics provider
+machine.wait_until_succeeds("ip netns exec nb-electrs nc -z localhost 4224")
```
nixbitcoin, line 41: Comment missing.

test/test-script.py
```diff
@@ -58,31 +58,49 @@ def assert_running(unit):
...
+machine.wait_until_succeeds("ip netns exec nb-nginx nc -z localhost 80")
+assert_matches("ip netns exec nb-nginx curl localhost", "nix-bitcoin")
+assert_matches("ip netns exec nb-nginx curl -L localhost/store", "tshirt")
 
 machine.wait_until_succeeds(log_has_string("bitcoind-import-banlist", "Importing node banlist"))
 assert_no_failure("bitcoind-import-banlist")
 
+### Security tests
+
+machine.fail("ip netns exec nb-bitcoind ping -c 1 169.254.1.17")
```
nixbitcoin, line 90: Fix tests by reformatting, setting ping deadline, and also testing successful pings.

modules/netns-isolation.nix
```diff
@@ -0,0 +1,296 @@
...
+      ) xs cfg.services.${s1}.connections
+    ) base (builtins.attrNames base);
+
+  enabledServices = filterAttrs (n: v: isEnabled n) cfg.services;
+  isEnabled = x: config.services.${x}.enable;
+
+  ip = "${pkgs.iproute}/bin/ip";
+  iptables = "${config.networking.firewall.package}/bin/iptables";
+
+  bridgeIp = "169.254.${toString cfg.addressblock}.10";
```
nixbitcoin, line 37: Remove `toString`

modules/netns-isolation.nix
```diff
@@ -0,0 +1,296 @@
...
+      systemd.services = {
+        netns-bridge = {
+          description = "Create bridge";
+          requiredBy = [ "tor.service" ];
+          before = [ "tor.service" ];
+          script = ''
+            ${ip} link add name br0 type bridge
+            ${ip} link set br0 up
+            ${ip} addr add ${bridgeIp}/24 brd + dev br0
+            ${iptables} -w -t nat -A POSTROUTING -s 169.254.${toString cfg.addressblock}.0/24 -j MASQUERADE
```
nixbitcoin, line 142: Remove toString

modules/netns-isolation.nix
```diff
@@ -0,0 +1,296 @@
...
+          '';
+          preStop = ''
+            ${iptables} -w -t nat -D POSTROUTING -s 169.254.${toString cfg.addressblock}.0/24 -j MASQUERADE
```
nixbitcoin, line 145: Remove toString

modules/netns-isolation.nix
```diff
@@ -0,0 +1,296 @@
...
+        cli = pkgs.writeScriptBin "bitcoin-cli" ''
+          netns-exec nb-bitcoind ${config.services.bitcoind.package}/bin/bitcoin-cli -datadir='${config.services.bitcoind.dataDir}' "$@"
+        '';
+      };
+
+      # clightning: Custom netns configs
+      services.clightning = mkIf config.services.clightning.enable {
+        host = "${netns.clightning.address}";
+        bitcoin-rpcconnect = "${netns.bitcoind.address}";
+        bind-addr = "${netns.clightning.address}:${toString config.services.clightning.onionport}";
```
nixbitcoin, line 228: Remove toString

modules/netns-isolation.nix
```diff
@@ -0,0 +1,296 @@
...
+        swap-cli = pkgs.writeScriptBin "liquidswap-cli" ''
+          netns-exec nb-liquidd ${pkgs.nix-bitcoin.liquid-swap}/bin/liquidswap-cli -c '${config.services.liquidd.dataDir}/elements.conf' "$@"
+        '';
+      };
+
+      # electrs: Custom netns configs
+      services.electrs = mkIf config.services.electrs.enable {
+        host =  if config.services.electrs.TLSProxy.enable then "${toString netns.nginx.address}" else "${toString netns.electrs.address}";
+        address = "${toString netns.electrs.address}";
+        daemonrpc = "${toString netns.bitcoind.address}:${toString config.services.bitcoind.rpc.port}";
```
nixbitcoin, line 268: Remove toString's

modules/netns-isolation.nix
```diff
@@ -0,0 +1,296 @@
...
+      };
+
+      # spark-wallet: Custom netns configs
+      services.spark-wallet = mkIf config.services.spark-wallet.enable {
+        host = "${netns.spark-wallet.address}";
+        extraArgs = "--no-tls";
+      };
+
+      # lightning-charge: Custom netns configs
+      services.lightning-charge.host = mkIf config.services.lightning-charge.enable "${toString netns.lightning-charge.address}";
```
nixbitcoin, line 278: Remove toString

modules/netns-isolation.nix
```diff
@@ -0,0 +1,296 @@
...
+      # electrs: Custom netns configs
+      services.electrs = mkIf config.services.electrs.enable {
+        host =  if config.services.electrs.TLSProxy.enable then "${toString netns.nginx.address}" else "${toString netns.electrs.address}";
+        address = "${toString netns.electrs.address}";
+        daemonrpc = "${toString netns.bitcoind.address}:${toString config.services.bitcoind.rpc.port}";
+      };
+
+      # spark-wallet: Custom netns configs
+      services.spark-wallet = mkIf config.services.spark-wallet.enable {
+        host = "${netns.spark-wallet.address}";
```
nixbitcoin, line 273: Remove dollar_curly

modules/bitcoind.nix
```diff
@@ -68,6 +70,13 @@ in {
         default = "/var/lib/bitcoind";
         description = "The data directory for bitcoind.";
       };
+      host = mkOption {
```
jonasnick, line 73: I feel like we don't really need these host options. In some modules (like bitcoind) they aren't even used. As far as I can see they are mainly used for the hidden service `toHost` option. For clightning for example we can directly use bind-addr. For lnd we can keep host (or rename to listen, because that's how the option is called in lnd). For bitcoind we can add a `bind` option (we should do that anyway) which if set will result in an additional line `bind=${...}` in the bitcoind config.

modules/netns-isolation.nix
```diff
@@ -0,0 +1,296 @@
...
+              ${ipNetns} link set lo up
+              ${ip} link add ${vethName} type veth peer name br-${vethName}
+              ${ip} link set ${vethName} netns ${netnsName}
+              ${ipNetns} addr add ${v.address}/24 dev ${vethName}
+              ${ip} link set br-${vethName} up
+              ${ipNetns} link set ${vethName} up
+              ${ip} link set br-${vethName} master br0
+              ${ipNetns} route add default via ${bridgeIp}
+              ${netnsIptables} -w -P INPUT DROP
+              ${netnsIptables} -w -A INPUT -s 127.0.0.1,${bridgeIp} -j ACCEPT
```
jonasnick, line 182: It should be possible to allow the `ipNetns` to connect to `ipNetns`. In order to allow that I think we need to add `${ipNetns}` here and in the `OUTPUT` rule below. That allows simplifying the tests by testing connectivity from the same container. Also we can remove additional connection from clightning to spark in the tests.

test/test-script.py
nixbitcoin: Fixed in a2dc3da63530a8793e1e689927b226e15e062cd1

modules/netns-isolation.nix
nixbitcoin: Actually do need `toString`

modules/netns-isolation.nix
nixbitcoin: Actually do need `toString`

modules/netns-isolation.nix
nixbitcoin: Actually do need `toString`

modules/netns-isolation.nix
nixbitcoin: Fixed in b51b7d6530e6f123e63da78442e70c03e7babeb3

modules/netns-isolation.nix
nixbitcoin: Fixed in aea2de95045a7a5071815e43c24b7a261c344b93

modules/netns-isolation.nix
nixbitcoin: Fixed in ac22262bf762fdc2b5b49f7da7b8a24e55024287

modules/bitcoind.nix
```diff
@@ -233,12 +256,22 @@ in {
       };
       cli = mkOption {
         type = types.package;
-        readOnly = true;
         default = pkgs.writeScriptBin "bitcoin-cli" ''
```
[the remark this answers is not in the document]
nixbitcoin: Fixed in dda5047e92b41bb0bc076971e9d6005ec3af1ced

modules/bitcoind.nix
nixbitcoin: Fixed where possible. Nanopos, lightning-charge and spark-wallet actually have a host option, so naming is correct. It's needed in electrs for now to direct traffic to the nginx netns, can be removed when #203 is merged.

modules/netns-isolation.nix
nixbitcoin: Fixed in 43a34be11cf79d252cbc0b846b12623b61c4fdbb

test/test-script.py
nixbitcoin: Fixed in a2dc3da63530a8793e1e689927b226e15e062cd1

nixbitcoin:
Rebased and fixed @jonasnick comments. 

<details>
  <summary>a2dc3da63530a8793e1e689927b226e15e062cd1 notes</summary>

- Disabled module by default
- Inserted ${ipNetns} firewall rule
- Removed unnecessary toStrings
- Removed unnecessary DOLLAR_CURLY's
- Defaulted bitcoin cli to cli-nonetns-exec
- Got rid of host options and used service-specific options
- Added `rpclisten` and `restlisten` options of type `types.listOf types.str` to lnd module
- Fixed missing comment in test_script.py
- Simplified spark-wallet test by testing from same container
- Set ping deadline
- Implemented must succeed pings
- Implemented netns-exec tests
  - That it drops capabilities
  - That it can't execute in bad namespaces
</details>

Running tests with and without `netns-isolation` requires creating something like
```
nodes =
    { machine1 =
        { config, pkgs, ... }: { … };
      machine2 =
        { config, pkgs, ... }: { … };
      …
    };
```
where `nix-bitcoin.netns-isolation.enable = mkForce true;` for one of the machines. `test-script.py` would have to be rewritten to perform tests on both machines concurrently. I think @erikarvstedt is the best person to implement this elegantly.

test/test-script.py
```diff
@@ -58,31 +66,67 @@ def assert_running(unit):
...
+assert_matches("ip netns exec nb-nginx curl -L localhost/store", "tshirt")
 
 machine.wait_until_succeeds(log_has_string("bitcoind-import-banlist", "Importing node banlist"))
 assert_no_failure("bitcoind-import-banlist")
 
+### Security tests
+
+# Positive ping tests
+machine.succeed(
+    "ip netns exec nb-bitcoind ping -c 1 -w 1 169.254.1.12 && \
```
jonasnick, line 100: Can you define constants for these addresses? That would give them a name and make it easy to see what they're referring to.

test/test-script.py
```diff
@@ -58,31 +66,67 @@ def assert_running(unit):
...
+    ip netns exec nb-nanopos ping -c 1 -w 1 169.254.1.13 || \
+    ip netns exec nb-nanopos ping -c 1 -w 1 169.254.1.14 || \
+    ip netns exec nb-nanopos ping -c 1 -w 1 169.254.1.15 || \
+    ip netns exec nb-nanopos ping -c 1 -w 1 169.254.1.16 || \
+    ip netns exec nb-nanopos ping -c 1 -w 1 169.254.1.17 || \
+    ip netns exec nb-nanopos ping -c 1 -w 1 169.254.1.20"
+)
+
+# netns-exec tests
+machine.fail("netns-exec nb-electrs ip a")
```
jonasnick, line 124: Can you add a comment to explain what this does and why this must fail?

test/test-script.py
```diff
@@ -58,31 +66,67 @@ def assert_running(unit):
...
+assert_matches_exactly(
+    "su operator -c 'netns-exec nb-bitcoind capsh --print | grep Current | tr -d \"\\n\"'",
+    "Current: =",
```
jonasnick, line 127: Can't you match against "Current: =\n" instead of using `tr`?

test/test-script.py
jonasnick: And how about checking that netns-exec can not be executed by users that are not operator?

.travis.yml
```diff
@@ -14,7 +14,8 @@ env:
     # CACHIX_SIGNING_KEY
   - secure: "xXCFZ7g+k5YmCGm8R8l3bZElVmt+RD1KscG3kGr5w4HyyDPTzFetPo+sT8bUpysDU0u3HWhfVhHtpog2mhNhwVl3tQwKXea3dHKC1i6ypBg3gjDngmJRR5wo++ocYDpK8qPaU7m/jHQTNFnTA4CbmMcc05GcYx/1Ai/ZGkNwWFjdIcVeOUoiol33gykMOXIGDg2qlXudt33wP53FHbX8L4fxzodWfAuxKK4AoGprxy5eSnU7LCaXxxJmu4HwuV+Ux2U1NfE/E33cvhlUvTQCswVSZFG06mg8rwhMG1ozsDvlL2itZlu/BeUQH5y3XMMlnJIUXUazkRBibf1w/ebVjpOF+anqkqmq8tcbFEa7T+RJeVTIsvP+L8rE8fcmuZtdg9hNmgRnLmaeT0vVwD1L2UqW9HdRyujdoS0jPYuoc1W7f1JQWfAPhBPQ1SrtKyNNqcbVJ34aN7b+4vCzRpQL1JTbmjzQIWhkiKN1qMo1v/wbIydW8yka4hc4JOfdQLaAJEPI1eAC1MLotSAegMnwKWE1dzm66MuPSipksYjZrvsB28cV4aCVUffIuRhrSr1i2afRHwTpNbK9U4/576hah15ftUdR79Sfkcoi1ekSQTFGRvkRIPYtkKLYwFa3jVA41qz7+IIZCf4TsApy3XDdFx91cRub7yPq9BeZ83A+qYQ="
   jobs:
-  - TestModules=1 STABLE=1
+  - DefaultTestModules=1 STABLE=1
+  - NetnsTestModules=1 STABLE=1
```
jonasnick, line 18: Perhaps simpler to directly define a `SCENARIO=` variable that is either "default" or "withnetns".

test/scenarios/default.py
```diff
@@ -1,32 +1,3 @@
...
-
-
-def assert_running(unit):
-    machine.wait_for_unit(unit)
-    assert_no_failure(unit)
-
-
-# Don't execute the following test suite when this script is running in interactive mode
-if "is_interactive" in vars():
-    raise Exception()
```
jonasnick, line 28: Seems like this is missing from `lib.py` now.

test/test-script.py
nixbitcoin: Fixed in 9d7181b5aaadf326728a3a5c9b255f0cae36598c

test/test-script.py
nixbitcoin: Both fixed in 9d7181b5aaadf326728a3a5c9b255f0cae36598c

test/test-script.py
nixbitcoin: Fixed in a3aaff192b1988064f22ad77a25514c2e535ee1f

.travis.yml
nixbitcoin: Fixed in 8a7c9b2839b1628679c14091a08bf36d549ae0d1
Much better

test/scenarios/default.py
nixbitcoin: Fixed in d220ab8e55c5f0536d2bf57aca7bdf6a3bb995a1

nixbitcoin:
Tests now run with multiple different configurations called `scenarios`. Travis automatically tests all specified scenarios. `./run.tests.sh --scenario <scenario> debug` confirmed working.

test/scenarios/withnetns.py
```diff
@@ -0,0 +1,139 @@
...
+assert_matches("ip netns exec nb-nginx curl -L localhost/store", "tshirt")
+
+machine.wait_until_succeeds(log_has_string("bitcoind-import-banlist", "Importing node banlist"))
+assert_no_failure("bitcoind-import-banlist")
+
+### Security tests
+
+# Positive ping tests (non-exhaustive)
+machine.succeed(
+    "bitcoindip='169.254.1.12' && \
```
jonasnick, line 64: Better to define these vars in python. Then you only have to do it once. See  https://github.com/jonasnick/nix-bitcoin/commits/withnetns-test-fix

nixbitcoin: Also replaced the ping commands and netns ip's in the entire script with python vars. ceaf1890fdb7a5b8c6d63436c37d193bc80db9f9

nixbitcoin:
Had to make bind dependent on listen with `${optionalString cfg.listen "bind=${cfg.bind}"}` in `liquidd` and `bitcoind`. Tests were failing with `Error: Cannot set -bind or -whitebind together with -listen=0`.

nixbitcoin:
Moved `nix-bitcoin.netns-isolation.enable` to `configuration.nix`

nixbitcoin:
Fixed final nits

PR #188, opened 2020-06-02
lnd: add option for configuring REST port

mmilata:
I've got a slightly contrived setup where I'm running nix-bitcoin in a container that shares network ports with another container which also uses 8080 and it's harder to change than in lnd ...

### DIFF
--- a/modules/lnd.nix
+++ b/modules/lnd.nix
@@ -14,6 +14,7 @@ let
     tlskeypath=${secretsDir}/lnd-key
 
     rpclisten=localhost:${toString cfg.rpcPort}
+    restlisten=localhost:${toString cfg.restPort}
 
     bitcoin.active=1
     bitcoin.node=bitcoind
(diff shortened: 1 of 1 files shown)


erikarvstedt:
Could you add [this fixup](https://github.com/erikarvstedt/nix-bitcoin/commit/1dee15af54649f463f8d264427d9a9e5afea6fef)?

modules/lnd.nix
```diff
@@ -45,10 +46,15 @@ in {
       description = "The data directory for LND.";
     };
     rpcPort = mkOption {
-      type = types.ints.u16;
```
nixbitcoin, line 48: Is it necessary to change this? From the NixOS module: "types.port: A port number. This type is an alias to types.ints.u16."

If so, should we change all occurrences of `types.ints.u16` -> `types.port`

nixbitcoin:
Concept ACK

modules/lnd.nix
erikarvstedt: Yes, we should change them separately: https://github.com/fort-nix/nix-bitcoin/pull/190

mmilata:
Updated, PTAL.

Issue #37, opened 2019-04-13
Enable Pruning


nixbitcoin:
**Time Horizon:** Medium term

**Description:** Enable pruning by pulling blocks that have been pruned but are required by clightning from the [Blockstream Explorer](https://blockstream.info)

**Motivation:** Pruning drastically reduces the footprint of the Bitcoin blockchain and therefore improves UX. Clightning might need pruned blocks and could therefore not work on pruned Blockchains. These blocks could be fetched from [Blockstream Explorer](https://blockstream.info) without drastically changing the security model.

**Possible Implementation:** Waiting for comment by @jonasnick

darosior:
Hi,

C-lightning does support pruning, as we fetch blocks only once : it can only be a problem if `lightningd` is restarted after a long time (such as `lightningd` known height is < pruned height). I.e. for a (v. roughly) 90 days breathing room a 20G pruned `bitcoind` is acceptable :
```python
>>> 1.5 * 6 * 24 * 90
19440.0
```

> Description: Enable pruning by pulling blocks that have been pruned but are required by clightning from the Blockstream Explorer

With the [Bitcoin backend modularization](https://lightning.readthedocs.io/PLUGINS.html#bitcoin-backend) you could have a mix between [`bcli`](https://github.com/ElementsProject/lightning/blob/master/plugins/bcli.c) (the default backend which polls `bcli`) and something like [`sauron`](https://github.com/lightningd/plugins/tree/master/sauron) or [`esplora_clnd_plugin`](https://github.com/lvaccaro/esplora_clnd_plugin) (they poll the Esplora API).

jonasnick:
Hi,

> C-lightning does support pruning

Yes, on my personal node I've been relying on this behavior and it's very easy with nix-bitcoin to enable pruning if you want to. Since running with a pruned node is not documented in the c-lightning repo, I'd be hesitant to recommend this to nix-bitcoin users. Also, when I played with pruning in the past, there were quite a few edge cases were bitcoind would sync and prune too fast for c-lightning to catch up.

darosior:
> Since running with a pruned node is not documented in the c-lightning repo,

There is now [a section in the README](https://github.com/ElementsProject/lightning/#pruning) :)
But I understand.

jonasnick:
[...]

PR #189, opened 2020-06-02
Update configuration.nix

nixbitcoin:
This PR updates `configuration.nix` comments to the most current inter-option dependencies and introduces an `assertions.nix` to verify that all these dependencies are met.

Closes https://github.com/fort-nix/nix-bitcoin/issues/37

### DIFF
--- a/examples/configuration.nix
+++ b/examples/configuration.nix
@@ -45,7 +45,7 @@
   ### LND
   # Disable clightning and uncomment the following line in order to enable lnd,
   # a lightning implementation written in Go.
-  # services.lnd.enable = assert (!config.services.clightning.enable); true;
+  # services.lnd.enable = true;
   ## WARNING
   # If you use lnd, you should manually backup your wallet mnemonic
   # seed. This will allow you to recover on-chain funds. You can run the
--- a/examples/configuration.nix
+++ b/examples/configuration.nix
@@ -23,7 +23,11 @@
   ### BITCOIND
   # Bitcoind is enabled by default if nix-bitcoin is enabled
   #
-  # You can override default settings from secure-node.nix as follows
+  # Enable this option to set pruning to a specified MiB value.
+  # clightning is compatible with pruning. See
+  # https://github.com/ElementsProject/lightning/#pruning for more information.
+  # LND and electrs are not compatible with pruning.
+  # Note: You can override default settings from secure-node.nix as follows
   # services.bitcoind.prune = lib.mkForce 100000;
   #
   # You can add options that are not defined in modules/bitcoind.nix as follows
--- a/examples/configuration.nix
+++ b/examples/configuration.nix
@@ -63,11 +63,10 @@
 
   ### ELECTRS
   # Enable this module to use electrs, an efficient re-implementation of
-  # Electrum Server in Rust. Only enable this if hardware wallets are
-  # disabled.
+  # Electrum Server in Rust.
   # services.electrs.enable = true;
   # If you have more than 8GB memory, enable this option so electrs will
-  # sync faster.
+  # sync faster. Only available if hardware wallets are disabled.
   # services.electrs.high-memory = true;
 
   ### LIQUIDD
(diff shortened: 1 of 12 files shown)


erikarvstedt:
Awesome :star2:
- [trivial fixups](https://github.com/erikarvstedt/nix-bitcoin/commit/e16dfaa2080d47c4e569f7c7d6e3be241b23a312)
- Remove assertion `cfg.recurring-donations.enable -> cfg.clightning.enable`, instead add `services.clightning.enable = true` to `recurring-donations.nix`. 
  In general, service modules should enable services they depend on. Same for `spark-wallet`, `webindex`, `nanopos`, `lightning-charge`. 
- The remaining assertions should be defined directly in the relevant module files.
  - E.g., `cfg.electrs.enable -> cfg.bitcoind.prune` belongs to `electrs`.
  - `cfg.lnd.enable -> !cfg.clightning.enable` has no definite location, but we should move it to `lnd.nix`
- The prune assertions for lnd and electrs fail when `prune = null`. I think it's best if we remove the possible `null` value for `bitcoind.prune` and set `prune = 0` in bitcoind as a default. This is will make the assertions work. We can then also remove `prune = 0` in `secure-node.nix` and the `mkForce` in `configuration.nix` (`bitcoind.prune = lib.mkForce `).
- Commit titles: Rename `configuration:` -> `example config:`

nixbitcoin:
I squashed your fixup, clarified nix-bitcoin will auto-detect invalid settings in all other comments as well, removed the possible null value for `bitcoind.prune`, and renamed the commits.

However, I don't think it's a good idea for service modules to enable services they depend on or to define assertions directly in the relevant module files.

- Example1: What would happen if a user wanted to use `lightning-charge` with a different clightning module named `clightning-with-plugins`? Nix-bitcoin would try to enable clightning in the background and throw a confusing error.
- Example 2: What would happen if a user tried to enable electrs outside nix-bitcoin with a custom bitcoind module that didn't even have a pruning option? 

IMO this reduces modularity and should instead be handled in higher-level nix-bitcoin expressions like `configuration.nix`.

erikarvstedt:
Great fixups, thanks!

I'm still convinced we should add the assertions to the corresponding modules (like in NixOS) as they are already tightly coupled:
- Example1: `lightning-charge` is directly dependent on `clightning` via the reference to `config.services.clightning.dataDir`. It would already fail to be compatible with a hypothetic `services.clightning-with-plugins`.
- Example 2: `electrs` is referencing the option `config.services.bitcoind.dataDirReadableByGroup` so it's already dependent on our nix-bitcoin specific bitcoind module.

You could indeed add the `cfg.lnd.enable -> !cfg.clightning.enable` assertion to `modules.nix`, because it belongs to more than one service module.

nixbitcoin:
Thanks for the explanation, I agree now on adding assertions to the modules. I'll work on implementing.

I still am not convinced it's a good idea to cross-enable modules without the users explicit direction. But I'm of course willing to reconsider. Any comment on this @jonasnick ?

jonasnick:
Imo allowing modules to be interchangeable is only very rarely useful. Instead of using a different clightning module it should be possible to make the existing clightning module more flexible. Moreover, having more logic in the modules themselves simplifies writing and maintaining them which I clearly noticed when @erikarvstedt added the option to set something like `nix-bitcoin.secrets.bitcoin-rpcpassword` directly in the module vs. the old way of defining all secrets in network.nix.

As for modules implicitly enabling other modules, I'd prefer to not do that to allow users to easily see and opt-in to what software they actually use on their node (for now at least... the number of modules is still manageable).

nixbitcoin:
Moved assertions to individual modules and split them up into separate commits.

erikarvstedt:
[Fixups](https://github.com/erikarvstedt/nix-bitcoin/commits/update-configuration.nix).
I've removed redundant antecedents in the assertion conditions and improved the assertion messages.
It's okay, even encouraged, to gather self-contained and related changes in a single commit, to ease reviewing.
`assertions: add lnd, clightning exclusivity` deserves its own commit because it contains two changes (hunks) in different files, where one change depends on the other.

nixbitcoin:
Just realized we need to ignore the lnd, clightning exclusivity assertion in order to run the tests, do you know how to do that?

erikarvstedt:
I've added [this](https://github.com/erikarvstedt/nix-bitcoin/commit/2213761f179258f5261b1a793d035578c3df72fc) to the fixups.

nixbitcoin:
Had to slightly correct your fixup and added a comment, explaining the `||`

nixbitcoin:
Changed the HWI assertion to check disablewallet directly, as this is the speicifc option that makes it incompatible.

PR #199, opened 2020-06-11
banlist: update to newest version

### LINKED ISSUE #198: Outdated bitcoind banlist
Currently it's 7 months old and https://people.xiph.org/~greg/banlist.cli.txt returns 404. Is there a more up-to-date one?

nixbitcoin:
Received by E-Mail from gmaxwell with comment:

> On 6/11/20 09:41 AM, Greg Maxwell wrote:
>> Yeah, sorry-- my webserver got wiped.

Closes #198


nixbitcoin:
I see that `systemd.services.bitcoind-import-banlist` only ever adds nodes to the banlist. Should we worry about unblocking nodes that get off greg's banlist?

jonasnick:
> A pragmatic solution for unblocking is to offer an option to delete banlist.dat before starting bitcoind.

Then you'd also lose the parts of the banlist you added manually.

gmaxwell:
There is a lifetime on the bans so if you remove one they'll expire eventually.  Early on I used to remove ones but I found abusive hosts coming back on the same IPs so lately I haven't been removing any.

nixbitcoin:
So, I think we should keep `systemd.services.bitcoind-import-banlist` append-only and merge the updated banlist.

jonasnick:
fwiw 31557600 seconds is one year

Issue #167, opened 2020-04-28
Backup module


nixbitcoin:
**Time Horizon:** Medium Term

**Description:** Create a module to backup all critical files (Bitcoin Wallet files, lnd channelstate, clightning hsmsecret, JoinMarket wallet files) and optionally time intensive data like Bitcoin blockchain and electrs database.

**Motivation:** We want to offer the user an easy way to backup his/her nix-bitcoin node in case of data loss or migration.

**Implementation:** Shell script, oneshot systemd script, duplicity

PR #200, opened 2020-06-12
backups: add module

nixbitcoin:
This PR introduces nix-bitcoin's own backups module which backups up all important files in `/var/lib` to `/var/lib/localBackups` (fetchable with `nixops scp`) or a remote target url once a day using duplicity. Backups are encrypted with an apg generated password. A user can also backup the blockchains and electrs data directory with `cfg.with-bulk-data` (off-by-default).

The idea is to create a generic `backups.nix` module, which at first has duplicity support but supports extending to a variety of backup programs like restic, git-annex and borg.

Closes https://github.com/fort-nix/nix-bitcoin/issues/167

### DIFF
--- a/modules/backups.nix
+++ b/modules/backups.nix
@@ -1,0 +1,87 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.services.backups;
+  filelist = pkgs.writeText "filelist.txt" ''
+    ${optionalString (!cfg.with-bulk-data) "- ${config.services.bitcoind.dataDir}/blocks"}
+    ${optionalString (!cfg.with-bulk-data) "- ${config.services.bitcoind.dataDir}/chainstate"}
+    ${config.services.bitcoind.dataDir}
+    ${config.services.clightning.dataDir}
+    ${config.services.lnd.dataDir}
+    /secrets/lnd-seed-mnemonic
+    ${optionalString (!cfg.with-bulk-data) "- ${config.services.liquidd.dataDir}/*/blocks"}
+    ${optionalString (!cfg.with-bulk-data) "- ${config.services.liquidd.dataDir}/*/chainstate"}
+    ${config.services.liquidd.dataDir}
+    ${optionalString cfg.with-bulk-data "${config.services.electrs.dataDir}"}
+    ${config.services.lightning-charge.dataDir}
+    /var/lib/tor
+    # Extra files
+    ${cfg.extraFiles}
+
+    # Exclude all unspecified files and directories
+    - /
+  '';
+
+in {
+  options.services.backups = {
+    enable = mkEnableOption "Backups service";
+    program = mkOption {
+      type = types.enum [ "duplicity" ];
+      default = "duplicity";
+      description = ''
+        Program with which to do backups.
+      '';
+    };
+    with-bulk-data = mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Whether to also backup Bitcoin blockchain and other bulk data.
+      '';
+    };
+    destination = mkOption {
+      type = types.str;
+      default = "file:///var/lib/localBackups";
+      description = ''
+        Where to back up to.
+      '';
+    };
+    frequency = mkOption {
+      type = types.nullOr types.str;
+      default = null;
+      description = ''
+        Run backup with the given frequency. If null, do not run automatically.
+      '';
+    };
+    extraFiles = mkOption {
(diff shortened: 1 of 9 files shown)


pkgs/generate-secrets/generate-secrets.sh
```diff
@@ -11,10 +11,12 @@ makePasswordSecret lnd-wallet-password
 makePasswordSecret liquid-rpcpassword
 makePasswordSecret lightning-charge-token
 makePasswordSecret spark-wallet-password
+makePasswordSecret backup-encryption-password
 
 [[ -e lightning-charge-env ]] || echo "API_TOKEN=$(cat lightning-charge-token)" > lightning-charge-env
 [[ -e nanopos-env          ]] || echo "CHARGE_TOKEN=$(cat lightning-charge-token)" > nanopos-env
 [[ -e spark-wallet-login   ]] || echo "login=spark-wallet:$(cat spark-wallet-password)" > spark-wallet-login
+[[ -e backup-encryption-env ]] || echo "PASSPHRASE=$(cat backup-encryption-password)" > backup-encryption-env
```
Kixunil, line 19: This is not safe as any user running `ps aux | grep 'echo PASSPHRASE` in the right moment (can be achieved using infinite loop) will see the passphrase. Proper way to do it:

```bash
echo -n PASSPHRASE= | cat - backup-encryption-password > backup-encryption-env
```

Of course, the other lines seem to be vulnerable too. FYI, I'm not even sure if passing the passphrase via environment is safe - I've read that it isn't but I couldn't figure out how to abuse it.

modules/backups.nix
```diff
@@ -0,0 +1,76 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.services.backups;
+  duplicity-filelist = pkgs.writeText "filelist.txt" ''
```
jonasnick, line 7: I think it would be good if the filelist can be extended with extra lines.

Generally, the most bare-bones but still helpful version backup module is just the list of files that are important to back up. That would  then allow users to integrate nix-bitcoin with their own backup tools without having to go through the step of having the `localBackups` directory (and an additional duplicity password). Therefore it would be nice if (1) the filelist can read from other modules and (2) the filelist can be read from the operator user's shell.

pkgs/generate-secrets/generate-secrets.sh
erikarvstedt: `echo` is a bash builtin, so its arguments are invisible outside of the script process.

nixbitcoin:
Added an extraFiles option and added `full-if-older-than` so duplicity perform a full backup if an incremental backup is requested, but the latest full backup in the collection is older than the given time, 1 month.

modules/backups.nix
nixbitcoin: > I think it would be good if the filelist can be extended with extra lines.
Resolved

nixbitcoin: > Generally, the most bare-bones but still helpful version backup module is just the list of files that are important to back up. That would then allow users to integrate nix-bitcoin with their own backup tools without having to go through the step of having the localBackups directory (and an additional duplicity password). Therefore it would be nice if (1) the filelist can read from other modules and (2) the filelist can be read from the operator user's shell.

The operator user can already read the filelist from the nix-bitcoin node's nix store.

We could make the filelist an option that defaults to our filelist, similar to nodeinfo script, so it can be read from all modules. I don't think it should be too difficult to translate the duplicity filelist syntax into something other programs understand.

nixbitcoin:
> Concept ACK. It's nice and simple. I wonder what the long term impact on the disk space usage is. Is there an option to prune the duplicity backups?

The user can always run `duplicity remove-all-but-n-full <count> --force target_url` to prune his backups directory. This deletes all backups sets that are older than the count:th last full backup (in other words, keep the last count full backups and associated incremental sets).

pkgs/generate-secrets/generate-secrets.sh
Kixunil: Would it make sense to write `builtin echo` in order to make double sure nothing screws up? I'm not sure how much is bash reliable in calling builtins.

erikarvstedt: `echo` has been a builtin since the first Bash release in the late 80s. It's totally safe to use. :wink:

nixbitcoin:
Fixed nits

nixbitcoin:
Rebased and fixed nits

PR #201, opened 2020-06-21
Update: nixpkgs and elementsd

nixbitcoin:
This PR updates nixpkgs (unstable rev includes bitcoind 0.20.0) and elementsd. It also adds an elementsd `get-sha256.sh` to easily determine verified sha256 sums for updates.


jonasnick:
according to the lnd v0.10.2 release notes, lnd isn't compatible with bitcoin 0.20 with versions before v0.10.2 https://github.com/lightningnetwork/lnd/releases/tag/v0.10.2-beta.rc1. So it would be best to update lnd first in nixpkgs.

nixbitcoin:
lnd v0.10.3 update included and bumped elementsd to 0.18.1.8

Issue #191, opened 2020-06-03
bitcoind RPC whitelist


nixbitcoin:
**Time Horizon:** Short Term

**Description:** Now that Bitcoin 0.20.0 is released we can utilize rpc whitelists to further lock down the bitcoind rpc server.

**Motivation:** clightning, lnd, liquid and electrs only need access to public bitcoind rpc calls.

**Implementation:** https://github.com/bitcoin/bitcoin/pull/12763

Kixunil:
FYI, the [list of methods](https://github.com/Kixunil/btc-rpc-proxy/blob/master/btc_rpc_proxy.toml#L7) verified to work with:

* lnd
* electrs

`getpeerinfo` is needed for NBXplorer due to https://github.com/MetacoSA/NBitcoin/issues/864

Issue #183, opened 2020-05-26
bitcoind should use RPC cookie auth not user/password


reardencode:
`--rpccookiefile=<path>` should be used to put the cookie file in a path that accessible with components that require bitcoind RPC access

Kixunil:
You might also be interested in [my article](https://github.com/Kixunil/security_writings/blob/master/cookie_files.md) explaining why it's more secure. (Maybe not entirely relevant after hardening, but having more security layers shouldn't hurt.)

nixbitcoin:
Sounds good to me. Any reason why we shouldn't switch over @jonasnick ? If you ACK I'll take care of it.

jonasnick:
We should use RPC cookie auth where possible, but we'll still need user/pass for services that don't support cookie authentication (lnd, elementsd).

Kixunil:
@jonasnick I achieved "middle ground" in my repo by using RPC cookie in the RPC proxy and having a user with fixed password and harmless permissions (public calls only). Kills two birds with one stone. :)

Of course, we should probably complain at those projects or make PRs.

nixbitcoin:
Switching to cookie authentication doesn't make sense if we're not using it for every service. Otherwise the security-win is negligible, because attackers can still connect to bitcoind with the static password.

Let's wait what bitcoind's rpc whitelists bring before we try to finesse cookie authentication. If with https://github.com/bitcoin/bitcoin/pull/12763 it's possible to give every service it's own rpc user/password and then restrict the users to public calls only, I think we should go that route.

Kixunil:
@nixbitcoin it's a little bit more involved. The native RPC filtering helps with "public" (harmless) calls. However, there's no way to secure non-public ones, as the Bitcoin Core filtering feature doesn't support cookies.

If there's any service that doesn't support cookies, but requires more dangerous features (if I remember correctly, Eclair is such service), it's impossible to solve without adding cookie support into such service. However, if all your services satisfy *at least one* of these requirements:

* supports cookies
* uses harmless calls

Then the safe way is to give a cookie to the services which support it and use RPC filtering for the services which don't. If I understand it correctly, Core can't use both cookie files and passwords, so the only possible options seem to be:

* PR to Core
* `btc-rpc-proxy` - this is what I use and maintain

PR #202, opened 2020-06-22
RPC Whitelist

nixbitcoin:
This PR changes bitcoind rpc authentication to the `rpcauth` feature with HMAC-SHA-256 hashed passwords. It implements an `rpcwhitelist` of harmless rpc calls for services with rpc access (`clightning`, `lnd`, `liquidd`, and `electrs`).

Right now, I transfer the HMAC's from the `/secrets` directory by replacing a placeholder inside `bitcoin.conf` via `sed` in the `systemd.services.bitcoind.preStart`. I'm not sure if this is the optimal solution.

Closes https://github.com/fort-nix/nix-bitcoin/issues/191
Closes https://github.com/fort-nix/nix-bitcoin/issues/183

### DIFF
--- a/modules/bitcoind.nix
+++ b/modules/bitcoind.nix
@@ -13,7 +13,7 @@ let
 
     ${optionalString cfg.testnet "testnet=1"}
     ${optionalString (cfg.dbCache != null) "dbcache=${toString cfg.dbCache}"}
-    "prune=${toString cfg.prune}
+    prune=${toString cfg.prune}
     ${optionalString (cfg.sysperms != null) "sysperms=${if cfg.sysperms then "1" else "0"}"}
     ${optionalString (cfg.disablewallet != null) "disablewallet=${if cfg.disablewallet then "1" else "0"}"}
     ${optionalString (cfg.assumevalid != null) "assumevalid=${cfg.assumevalid}"}
--- a/modules/bitcoind.nix
+++ b/modules/bitcoind.nix
@@ -5,6 +5,7 @@ with lib;
 let
   cfg = config.services.bitcoind;
   inherit (config) nix-bitcoin-services;
+  secretsDir = config.nix-bitcoin.secretsDir;
 
   configFile = pkgs.writeText "bitcoin.conf" ''
     # We're already logging via journald
@@ -33,8 +34,8 @@ let
     }
     ${lib.concatMapStrings (rpcbind: "rpcbind=${rpcbind}\n") cfg.rpcbind}
     ${lib.concatMapStrings (rpcallowip: "rpcallowip=${rpcallowip}\n") cfg.rpcallowip}
-    ${optionalString (cfg.rpcuser != null) "rpcuser=${cfg.rpcuser}"}
-    ${optionalString (cfg.rpcpassword != null) "rpcpassword=${cfg.rpcpassword}"}
+    # Credentials for bitcoin-cli
+    rpcuser=${cfg.rpc.users.privileged.name}
 
     # Wallet options
     ${optionalString (cfg.addresstype != null) "addresstype=${cfg.addresstype}"}
@@ -110,7 +111,7 @@ in {
                 '';
               };
               passwordHMAC = mkOption {
-                type = with types; uniq (strMatching "[0-9a-f]+\\$[0-9a-f]{64}");
+                type = types.str;
                 example = "f7efda5c189b999524f151318c0c86$d5b51b3beffbc02b724e5d095828e0bc8b2456e9ac8757ae3211a5d9b16a22ae";
                 description = ''
                   Password HMAC-SHA-256 for JSON-RPC connections. Must be a string of the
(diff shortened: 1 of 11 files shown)


modules/presets/secure-node.nix
```diff
@@ -65,6 +65,45 @@ in {
...
+          "verifytxoutproof"
+          "createrawtransaction"
+          "decoderawtransaction"
+          "decodescript"
+          "getrawtransaction"
+          "sendrawtransaction"
+          "estimatefee"
+          "estimatepriority"
+          "estimatesmartfee"
+          "estimatesmartpriority"
```
Kixunil, line 104: Few more harmless calls to consider:

* `validateaddress`
* `testmempoolaccept`
* `signrawtransactionwithkey` - the key is provided in the call, so stealing is not possible
* `scantxoutset` - I assume this operation takes some time, so maybe it could be used for DoS, but probably not too huge issue?

nixbitcoin:
Rebased

modules/presets/secure-node.nix
nixbitcoin: Incorporated these harmless calls in 88c7348105bdbf0a1e7947ddb32c0a3263b674f7. Thank you for the suggestions.

jonasnick: estimatepriority and estimatesmartpriority were removed in [0.15](https://bitcoincore.org/en/releases/0.15.0/).

nixbitcoin:
Rebased, expanded tests, removed deprecated rpc calls

Kixunil:
FYI, I found yesterday that NBXplorer/BTCPayServer don't work without `getpeerifo` **on regtest**.

If you have regtest packages, then you may want to special-case this.

pkgs/generate-secrets/default.nix
```diff
@@ -1,6 +1,9 @@
 { pkgs }: with pkgs;
 
+let
+rpcauth = pkgs.callPackage ./rpcauth { };
```
jonasnick, line 4: I think you can replace this with:
```
  rpcauth = pkgs.writeScriptBin "rpcauth" (builtins.readFile ./rpcauth/rpcauth.py);
```
(mind the indentation) and remove `rpcauth/default.nix`

modules/bitcoind.nix
```diff
@@ -266,7 +267,9 @@ in {
         readOnly = true;
         type = types.package;
         default = pkgs.writeScriptBin "bitcoin-cli" ''
-          exec ${cfg.package}/bin/bitcoin-cli -datadir='${cfg.dataDir}' "$@"
+          exec ${cfg.package}/bin/bitcoin-cli \
+          -rpcuser=${cfg.rpc.users.privileged.name} -rpcpassword=$(cat ${secretsDir}/bitcoin-rpcpassword-privileged) \
```
jonasnick, line 271: Perhaps better to leave the rpcuser and rpcpassword in the bitcoin.conf so we don't have to specify them here

Kixunil: `-rpcpassword` being entered from command line is actually insecure. `bitcoin-cli` has an option `-stdinrpcpass` to solve this problem - just read it from that file.

jonasnick:
Here's an exhaustive list of the RPCs I'd consider "public" from bitcoin 0.21.0:
```
# Blockchain
"getbestblockhash"
"getblock"
"getblockchaininfo"
"getblockcount"
"getblockfilter"
"getblockhash"
"getblockheader"
"getblockstats"
"getchaintips"
"getchaintxstats"
"getdifficulty"
"getmempoolancestors"
"getmempooldescendants"
"getmempoolentry"
"getmempoolinfo"
"getrawmempool"
"gettxout"
"gettxoutproof"
"gettxoutsetinfo"
"scantxoutset"
"verifytxoutproof"
# Mining
"getblocktemplate"
"getmininginfo"
"getnetworkhashps"
# Network
"getnetworkinfo"
# Rawtransactions
"analyzepsbt"
"combinepsbt"
"combinerawtransaction"
"converttopsbt"
"createpsbt"
"createrawtransaction"
"decodepsbt"
"decoderawtransaction"
"decodescript"
"finalizepsbt"
"fundrawtransaction"
"getrawtransaction"
"joinpsbts"
"sendrawtransaction"
"signrawtransactionwithkey"
"testmempoolaccept"
"utxoupdatepsbt"
# Util
"createmultisig"
"deriveaddresses"
"estimatesmartfee"
"getdescriptorinfo"
"signmessagewithprivkey"
"validateaddress"
"verifymessage"
# Zmq
"getzmqnotifications"
```

Every time we update bitcoind we can check for new public RPCs and add them to the list.

nixbitcoin:
Rebased, fix nits, and Jonas Nick public rpc call list implemented

modules/bitcoind.nix
nixbitcoin: Fixed in fb94d19
I think it's even better to use the config file than `-stdinrpcpass`

pkgs/generate-secrets/default.nix
nixbitcoin: Fixed in fb94d19

nixbitcoin:
@Kixunil We're taking note of all the NBXplorer issues here and will get to them when we merge BTCPayserver

modules/bitcoind.nix
nixbitcoin: BTW, this is just defense in depth since we already forbid users from viewing other's commands with hidepid

PR #209, opened 2020-07-10
Lightning loop

nixbitcoin:
This PR integrates lightning-loop in nix-bitcoin. Lightning Loop is a [Non-Custodial Off/On Chain Bridge](https://github.com/lightninglabs/loop). It can be used for improvised CoinSwaps that utilize the privacy properties of LN.

Details

-  Based on https://github.com/fort-nix/nix-bitcoin/pull/201.
-  For testing purposes it includes https://github.com/NixOS/nixpkgs/pull/89668 in 7d512a5. This can be removed as soon as https://github.com/NixOS/nixpkgs/pull/89668 is merged, and the pinned nixpkgs in 3cbb691 are bumped.
- AFAICT the datadir can't be specified for loopd, but because of `ReadWritePaths` it's created in `/var/lib/lnd/.loopd` with the right permissions, which is kind of what we want.

Tested with a succesful `loop out` tx `f8178663a03d5b492d756f43b9989c1e7f42640b0c837d4957f29080d025d3bd`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ env:
   - PKG=electrs STABLE=1
   - PKG=electrs STABLE=0
   - PKG=liquid-swap STABLE=1
+  - PKG=lightning-loop STABLE=0
   - PKG=nixops19_09 STABLE=1
 script:
   - printf '%s (%s)\n' "$NIX_PATH" "$VER"
--- a/examples/configuration.nix
+++ b/examples/configuration.nix
@@ -129,6 +129,12 @@
   # a network-level as much as possible.
   # nix-bitcoin.netns-isolation.enable = true;
 
+  ### lightning-loop
+  # Enable this module to use lightninglab's non-custodial off/on chain bridge.
+  # loopd (lightning-loop daemon) will be started automatically. Users can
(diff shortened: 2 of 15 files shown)


nixbitcoin:
Rebased with upstream lnd

sputn1ck:
I have not looked at the code, but tested a loop in and loop out in my krops setup and it worked great!

nixbitcoin:
Thank you for the blind trust!

nixbitcoin:
Check that lightning-loop fails with the right error, making sure lightning-loop can connect to lnd

Kixunil:
Today I attempted to integrate loop into my deb repository and found out it [doesn't support authentication yet](https://github.com/lightninglabs/loop/issues/194) How did you secure it? Do you rely on namespaces only?

nixbitcoin:
Yes, lightning-loop relies on running inside its own network namespace for security. Only users inside the `nb-lightning-loop` namespace can reach the `loopd` daemon, and only the semi-privileged operator user can place processes into that namespace. Ideally we would get defense-in-depth with authentication. Thank you for linking the issue, we will update lightning-loop as soon as it's closed.

Issue #207, opened 2020-07-08
c-lightning times out and doesn't restart when started at the same time as bitcoind


nixbitcoin:
Clightning is timing out waiting for bitcoind to accept rpc calls. It doesn't restart on this failure, as should be expected with `Restart=on-failure`.

```
Jul 08 13:39:46 nix-bitcoin systemd[1]: Starting Run clightningd...
Jul 08 13:40:29 nix-bitcoin lightningd[28136]: Timed out while waiting for (one of) the Bitcoin backend plugin(s) to complete the handshake.
Jul 08 13:40:29 nix-bitcoin systemd[1]: clightning.service: Main process exited, code=exited, status=1/FAILURE
Jul 08 13:41:20 nix-bitcoin systemd[1]: clightning.service: Start-post operation timed out. Stopping.
Jul 08 13:41:20 nix-bitcoin systemd[1]: clightning.service: Control process exited, code=killed, status=15/TERM
Jul 08 13:41:20 nix-bitcoin systemd[1]: clightning.service: Failed with result 'exit-code'.
Jul 08 13:41:20 nix-bitcoin systemd[1]: Stopped Run clightningd.
Jul 08 13:41:20 nix-bitcoin systemd[1]: clightning.service: Consumed 1min 8.809s CPU time, received 16.7K IP traffic, sent 15.2K IP traffic.
```

PR #210, opened 2020-07-19
bitcoind: postStart wait until bitcoind can receive rpc calls

nixbitcoin:
https://github.com/ElementsProject/lightning/pull/3565 added a timeout if the Bitcoin plugin never completes the handshake, instead of waiting indefinitely for bitcoind to come up like before.

Because the `ExecStartPost` runs concurrently to the `ExecStart`, the systemd service doesn't fail when `ExecStart` fails on the aforementioned clightning timeout. Instead, systemd times out the `ExecStartPost` itself and shuts down the whole service gracefully, preventing `Restart=on-failure` from restarting the service.

This situation can also be alleviated by 
```
TimeoutStartFailureMode = "kill";
```
However, this leads to the equally unwanted result that dependent services like spark-wallet don't restart when clightning restarts successfully (https://github.com/systemd/systemd/issues/1312).

Hence, I think the best course of action is implementing a `while` loop in the clightning `preStart` to wait until bitcoind can receive rpc calls before clightning executes `ExecStart`.

Closes https://github.com/fort-nix/nix-bitcoin/issues/207

### DIFF
--- a/modules/bitcoind.nix
+++ b/modules/bitcoind.nix
@@ -260,8 +260,8 @@ in {
         default = cfg.cli-nonetns-exec;
         description = "Binary to connect with the bitcoind instance.";
       };
-      # Needed because bitcoind-import-banlist already executes inside
-      # nb-bitcoind, hence it doesn't need netns-exec prefixed.
+      # Needed because bitcoin-cli commands executed through systemd already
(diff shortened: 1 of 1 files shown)


nixbitcoin:
Decided to generalize the wait until bitcoind can receive rpc calls by placing it in the bitcoind postStart instead of the individual services' preStart.

PR #212, opened 2020-07-26
Fix deploy-nixops.sh conditional binary operator

gotcha:
Not being a bash expert, I am not sure this fix/syntax works everywhere.

### DIFF
--- a/examples/deploy-nixops.sh
+++ b/examples/deploy-nixops.sh
@@ -8,7 +8,7 @@ set -euo pipefail
 # Feel free to modify or to run nix-shell and execute individual statements of this
 # script in the interactive shell.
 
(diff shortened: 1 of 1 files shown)


jonasnick:
Thanks for the PR but I think the space is required. A better fix would be something like this https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash/13864829#13864829.

nixbitcoin:
This might be the solution 537b280056fc0250130f844b77d23d60116f2597

erikarvstedt:
You're probably getting an error because you're using an outdated Bash version that's shipping with macOS.
To fix this, just run the examples in a nix-shell, as shown in the README.

erikarvstedt:
This PR should be closed.

PR #213, opened 2020-07-27
Fix systemd leak

nixbitcoin:
Mitigates a security issue that allows unprivileged users to read other unprivileged user's processes' credentials from CGroup using `systemctl status`. Also adds tests to make sure this issue doesn't reoccur.

### DIFF
--- a/modules/dbus.nix
+++ b/modules/dbus.nix
@@ -1,0 +1,55 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  inherit (config) nix-bitcoin-services;
+  dataDir = "/var/lib/dbus-hardening";
+  # Mitigates a security issue that allows unprivileged users to read
+  # other unprivileged user's processes' credentials from CGroup using
+  # `systemctl status`.
+  dbus-hardening = pkgs.writeText "dbus.conf" ''
+    <?xml version="1.0" encoding="UTF-8"?> <!-- -*- XML -*- -->
+
+    <!DOCTYPE busconfig PUBLIC
+     "-//freedesktop//DTD D-BUS Bus Configuration 1.0//EN"
+     "http://www.freedesktop.org/standards/dbus/1.0/busconfig.dtd">
+
+    <busconfig>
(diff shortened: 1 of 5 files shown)


erikarvstedt:
What exactly is the security issue here?
According to my understanding, all data returned by `systemctl status` is public information available to all users, e.g. via `/proc`.
Also, if this were an issue this should also be fixed upstream.

jonasnick:
`systemctl status clightning` would reveal the bitcoin-cli command including the `-rpcpassword` arguments in the cgroups process tree. I believe @nixbitcoin has a script to reproduce this.

erikarvstedt:
The `ARGV` of all processes are public, e.g. via `/proc/<PID>/cmdline`.
If clightning leaks secrets through subprocess arguments, that's a clightning bug which can't be fixed by this PR.

jonasnick:
Doesn't work on my machine ("file not found"). It's prevented by `security.hideProcessInformation = true;`, which is part of the hardened kernel and set explicitly in this PR.

jonasnick:
> If clightning leaks secrets through subprocess arguments

And yeah, that'd be bad and should be fixed upstream as well.

erikarvstedt:
Ah, I missed that `security.hideProcessInformation` was enabled.
I also have to correct myself: Some process info like `cmdline` are only exposed by `/proc` and not by further means like syscalls.

This is indeed a systemd bug, great job catching this, @nixbitcoin!
Could you file a bug for systemd? `GetUnitProcesses` should only show info for the current user's unit when `/proc` access is restricted.

This PR still has some critical issues, which are fixed [here](https://github.com/fort-nix/nix-bitcoin/pull/225).

nixbitcoin:
> Could you file a bug for systemd? `GetUnitProcesses` should only show info for the current user's unit when `/proc` access is restricted.

Bug filed @ https://github.com/systemd/systemd/issues/16825

jonasnick:
Bug filed @ https://github.com/ElementsProject/lightning/issues/3984

PR #215, opened 2020-08-02
Improve abstraction in test scenarios

jonasnick:
This will probably change as we add more tests but for now this solution seems fine.


test/scenarios/default.py
```diff
@@ -1,100 +1,41 @@
...
-# Restating rpcuser & rpcpassword overrides privileged credentials
-machine.fail(
-    "bitcoin-cli -rpcuser=publicrpc -rpcpassword=$(cat /secrets/bitcoin-rpcpassword-public) help"
-)
-machine.wait_until_succeeds(
-    log_has_string("bitcoind", "RPC User publicrpc not allowed to call method help")
-)
+def spark_wallet():
+    machine.wait_for_open_port(9737)
+    spark_auth = re.search("login=(.*)", succeed("cat /secrets/spark-wallet-login"))[1]
```
nixbitcoin, line 7: Is there a reason why you changed the order of `machine.wait_for_open_port` & `spark_auth`?

test/scenarios/default.py
```diff
@@ -1,100 +1,41 @@
...
-# Stop electrs from spamming the test log with 'wait for bitcoind sync' messages
-succeed("systemctl stop electrs")
 
-assert_running("liquidd")
-machine.wait_until_succeeds("elements-cli getnetworkinfo")
-assert_matches("su operator -c 'elements-cli getnetworkinfo' | jq", '"version"')
-succeed("su operator -c 'liquidswap-cli --help'")
+def lightning_charge():
+    machine.wait_for_open_port(9112)
+    charge_auth = re.search("API_TOKEN=(.*)", succeed("cat /secrets/lightning-charge-env"))[1]
```
nixbitcoin, line 13: Is there a reason why you changed the order of `machine.wait_for_open_port` & `charge`?

test/scenarios/lib.py
```diff
@@ -32,3 +32,107 @@ def assert_running(unit):
...
+    machine.fail(
+        "bitcoin-cli -rpcuser=publicrpc -rpcpassword=$(cat /secrets/bitcoin-rpcpassword-public) help"
+    )
+    machine.wait_until_succeeds(
+        log_has_string("bitcoind", "RPC User publicrpc not allowed to call method help")
+    )
+
+    assert_running("electrs")
+    extra_tests.pop("electrs")()
+
```
nixbitcoin, line 60: Unnecessary newline

test/scenarios/withnetns.py
```diff
@@ -10,153 +10,91 @@
...
+    assert_matches("ip netns exec nb-lightning-charge curl %s:9116" % nanopos_ip, "tshirt")
+
+
+def web_index():
+    machine.wait_until_succeeds("ip netns exec nb-nginx nc -z localhost 80")
+    assert_matches("ip netns exec nb-nginx curl localhost", "nix-bitcoin")
+    assert_matches("ip netns exec nb-nginx curl -L localhost/store", "tshirt")
+
+
+def post_clightning():
```
nixbitcoin, line 49: I think this should be called `netns_isolation` since all its tests concern Network Namespace Isolation

nixbitcoin:
What about making lnd and clightning run at the same time by adding a port number to `listen` and `bind-addr` respectively. This would allow us to specify the port number to listen on locally, while leaving the hidden service onionport the same. We could then change the lnd/clightning exclusivity assertion to a port exclusivity assertion, allowing us to run both lnd and clightning tests at the same time, reducing the complexity of our two stage test process.

test/scenarios/default.py
jonasnick: yes, you usually define variables when you use them and not before

test/scenarios/withnetns.py
jonasnick: if this is called netns isolation it's not clear if it's executed in the clightning or in the lnd world. Makes more sense to me to name the keys in extra_tests according to the location it will be executed.

nixbitcoin:
I've added the necessary commits to make clightning & lnd tests run concurrently to my branch @ https://github.com/nixbitcoin/nix-bitcoin/commits/abstract-tests

test/scenarios/default.py
nixbitcoin: ACK

test/scenarios/default.py
nixbitcoin: ACK

test/scenarios/withnetns.py
nixbitcoin: What about `final`?

nixbitcoin:
> I've added the necessary commits to make clightning & lnd tests run concurrently to my branch @ https://github.com/nixbitcoin/nix-bitcoin/commits/abstract-tests

Let's do this in another PR?

jonasnick:
> Let's do this in another PR?

good idea

PR #220, opened 2020-08-11
btcpayserver: init at 1.0.5.4

kcalvinalvin:
Another method of packaging dotnet projects. I'm following the method that is used across nixpkgs as seen in these packages:
[osu-lazer](https://github.com/NixOS/nixpkgs/blob/8af72d55305fe1e0ba8657a2b69fb20ca347fc9a/pkgs/games/osu-lazer/default.nix)
[ryujinx](https://github.com/NixOS/nixpkgs/blob/2e2f87db13e0d072f90c99787321edb69a26ad42/pkgs/misc/emulators/ryujinx/default.nix)
[wasabi](https://github.com/NixOS/nixpkgs/blob/847ddfc8674ad2267b525762bc8749bd082ac2f1/pkgs/applications/blockchains/wasabibackend/default.nix)

The packages used here is also pushed upstream and isn't yet merged but is passing tests there.
https://github.com/NixOS/nixpkgs/pull/95104


pkgs/nbxplorer/result
```diff
@@ -0,0 +1 @@
+/nix/store/bxwbv6xfgcl173zjfixvmk1bjirgf0la-NBXplorer
```
nixbitcoin, line 1: Why is this file included?

pkgs/btcpayserver/default.nix
```diff
@@ -0,0 +1,79 @@
...
+      ln -s $out/bin/run.sh $out/bin/btcpayserver
+  '';
+
+  dontStrip = true;
+
+  meta = with lib; {
+    description = "BTCPay Server is a self-hosted, open-source cryptocurrency payment processor. It's secure, private, censorship-resistant and free";
+    longDescription = ''
+      BTCPay Server is a free and open-source cryptocurrency payment processor which allows you to receive payments in Bitcoin and altcoins directly, with no fees, transaction cost or a middleman.
+
```
nixbitcoin, line 63: Is it normal to include newlines between longDescription lines?

pkgs/nbxplorer/result
kcalvinalvin: Yeah that doesn't need to be there. I'm doing cleanup at the moment.

pkgs/btcpayserver/default.nix
kcalvinalvin: I don't think there are any standards for longDescription. Will remove them

jb55:
this method is definitely simpler and leverages dotnet to fetch deps instead of dotnet2nix which would be more maintainable. Concept ACK!

Kixunil:
FYI, the issue around NBXplorer capability detection was fixed, so strict RPC permissions are no longer a problem for PayJoin. :tada:

pkgs/default.nix
```diff
@@ -12,6 +12,7 @@
   nixops19_09 = pkgs.callPackage ./nixops { };
   netns-exec = pkgs.callPackage ./netns-exec { };
   lightning-loop = pkgs.callPackage ./lightning-loop { };
+  btcpayserver = pkgs.callPackage ./btcpayserver { };
```
jb55, line 15: `nix build -f . btcpayserver` didn't work for me, I had to do:

```diff
diff --git a/pkgs/default.nix b/pkgs/default.nix
index f87f774..0dbb183 100644
--- a/pkgs/default.nix
+++ b/pkgs/default.nix
@@ -1,5 +1,5 @@
 { pkgs ? import <nixpkgs> {} }:
-{
+rec {
   lightning-charge = pkgs.callPackage ./lightning-charge { };
   nanopos = pkgs.callPackage ./nanopos { };
   spark-wallet = pkgs.callPackage ./spark-wallet { };
@@ -12,7 +12,8 @@
   nixops19_09 = pkgs.callPackage ./nixops { };
   netns-exec = pkgs.callPackage ./netns-exec { };
   lightning-loop = pkgs.callPackage ./lightning-loop { };
-  btcpayserver = pkgs.callPackage ./btcpayserver { };
+  nbxplorer = pkgs.callPackage ./nbxplorer { };
+  btcpayserver = pkgs.callPackage ./btcpayserver { inherit nbxplorer; };
 
   pinned = import ./pinned.nix;
 }

```

jb55:
There are some impurities when I try to run it:
```
jb55@monad> cd ./result/bin
jb55@monad> ./btcpayserver
which: no btcpayserver in (...)
readlink: missing operand
Try 'readlink --help' for more information.
dirname: missing operand
Try 'dirname --help' for more information.
./btcpayserver: line 4: dotnet: command not found
```

pkgs/btcpayserver/default.nix
```diff
@@ -0,0 +1,75 @@
...
+    dotnet restore --source nixos BTCPayServer.Rating/BTCPayServer.Rating.csproj
+    dotnet build --no-restore -c Release BTCPayServer.Rating/BTCPayServer.Rating.csproj
+
+    dotnet restore --source nixos BTCPayServer/BTCPayServer.csproj
+    dotnet build --no-restore -c Release BTCPayServer/BTCPayServer.csproj
+  '';
+
+  installPhase = ''
+      mkdir -p $out/bin/
+      echo -e '#!/usr/bin/env bash \n#launch BTCpayserver \nscriptdir=`dirname $(readlink $(which btcpayserver))` \ndotnet run --no-launch-profile --no-build -c Release -p "$scriptdir/../BTCPayServer/BTCPayServer.csproj" -- $@
```
jb55, line 51: `dotnet` needs nix-store path substitution. I achieve this by using sed: 

https://github.com/fort-nix/nix-bitcoin/blob/33c17b2e44c1bb889f4cead682c77de454aabef4/pkgs/btcpayserver/dotnet-build.nix#L118

https://github.com/fort-nix/nix-bitcoin/blob/33c17b2e44c1bb889f4cead682c77de454aabef4/pkgs/btcpayserver/dotnet-build.nix#L167

pkgs/default.nix
nixbitcoin: I solved this problem by adding 

```diff
diff --git a/pkgs/default.nix b/pkgs/default.nix
index f87f774a47a..6209b00dba9 100644
--- a/pkgs/default.nix
+++ b/pkgs/default.nix
@@ -12,6 +12,7 @@
   nixops19_09 = pkgs.callPackage ./nixops { };
   netns-exec = pkgs.callPackage ./netns-exec { };
   lightning-loop = pkgs.callPackage ./lightning-loop { };
+  nbxplorer = pkgs.callPackage ./nbxplorer { };
   btcpayserver = pkgs.callPackage ./btcpayserver { };
```
and removing
```diff
diff --git a/pkgs/btcpayserver/default.nix b/pkgs/btcpayserver/default.nix
index 0acb62dd447..dcf5a3ec42d 100644
--- a/pkgs/btcpayserver/default.nix
+++ b/pkgs/btcpayserver/default.nix
@@ -1,4 +1,4 @@
-{ stdenv, fetchFromGitHub, fetchurl, nbxplorer, dotnetPackages, dotnetCorePackages, lib }:
+{ stdenv, fetchFromGitHub, fetchurl, dotnetPackages, dotnetCorePackages, lib }:
```
and then called both packages from the module. I think this is the better solution as someone might want to use NBXplorer without BTCPayServer.

pkgs/btcpayserver/default.nix
nixbitcoin: I solved this problem by doing

```diff
-      echo -e '#!/usr/bin/env bash \n#launch BTCpayserver \nscriptdir=`dirname $(readlink $(which btcpayserver))` \ndotnet run --no-launch-profile --no-build -c Release -p "$scriptdir/../BTCPayServer/BTCPayServer.csproj" -- $@
-' > $out/bin/run.sh
+      echo -e "#!/usr/bin/env bash \n#launch BTCpayserver \ndotnet run --no-launch-profile --no-build -c Release -p \"$out/BTCPayServer/BTCPayServer.csproj\" -- \$@" > $out/bin/run.sh
```
and
```diff
-    echo -e '#!/usr/bin/env bash \n#launch NBXplorer \nscriptdir=`dirname $(readlink $(which NBXplorer))` \ndotnet exec --additionalprobingpath $scriptdir/../home/.nuget/packages/ $scriptdir/../NBXplorer/bin/Release/netcoreapp3.1/NBXplorer.dll $@' > $out/bin/run.sh
+    echo -e "#!/usr/bin/env bash \n#launch NBXplorer \ndotnet exec --additionalprobingpath $out/home/.nuget/packages/ $out/NBXplorer/bin/Release/netcoreapp3.1/NBXplorer.dll \$@" > $out/bin/run.sh
```

nixbitcoin:
> FYI, the issue around NBXplorer capability detection was fixed, so strict RPC permissions are no longer a problem for PayJoin. tada

In my module (under development) I created a new `btcpayserver` rpc user that has several more rpc calls.

pkgs/btcpayserver/default.nix
kcalvinalvin: Fixed the impurities by directly using ```$out```

pkgs/default.nix
kcalvinalvin: Applied @nixbitcoin's solution

jonasnick:
@kcalvinalvin thank you! Since you've also opening a PR to nixpkgs wouldn't it be best to get it merged there and close this PR to avoid splitting the discussion?

pkgs/btcpayserver/default.nix
kcalvinalvin: Improved upon that with the method @jb55 suggested

kcalvinalvin:
@jonasnick I thought maybe a separate PR would be useful since nix-bitcoin is fairly independent from nixpkgs. Will close if there is no need for this PR

Issue #193, opened 2020-06-03
Allow clearnet routing


nixbitcoin:
**Time Horizon:** Mid Term

**Description:** Allow off-by-default clearnet routing on top of network namespaces.

**Motivation:** It should be as easy as toggling a switch for user's who need it to route receive and send clearnet traffic in nix-bitcoin. This is especially important for companies and large infrastructure providers. Mentioned in other issues https://github.com/fort-nix/nix-bitcoin/issues/94 https://github.com/fort-nix/nix-bitcoin/pull/110 https://github.com/fort-nix/nix-bitcoin/pull/179 https://github.com/fort-nix/nix-bitcoin/issues/182

**Implementation:** Single option NAT forward for all public-facing services.

nixbitcoin:
I propose adding a generalized option for each network-facing service (electrs, spark-wallet, c-lightning, lnd) that allows a user to specify exactly which kinds of networking he wants to accept.

The options would look something like this
```
services.electrs.networking.tor = true;
services.electrs.networking.clearnet = "plain";
services.electrs.networking.wireguard true;
```
allowing him to accept connections over Tor HS, clearnet (plain, tls-selfsigned OR tls-acme), and Wireguard. All connection information should show up in `nodeinfo`. Wireguard should display a QR code of the connection information in `nodeinfo`.

Tor HS configuration can remain as before.

Clearnet connections should all be routed over nginx, as this makes using TLS easy.

For Wireguard, the AllowedIPs field should be set to the IP behind the connection FQDN, so as to only forward traffic for that specific destination over Wireguard.

This would also close https://github.com/fort-nix/nix-bitcoin/issues/182

nixbitcoin:
This is possible and demonstrated under https://github.com/fort-nix/nixbitcoin.org

PR #221, opened 2020-08-12
nbxplorer/btcpayserver: add module

nixbitcoin:
Builds on https://github.com/fort-nix/nix-bitcoin/pull/220

Notes
* enforceTor is currently disabled for `systemd.services.btcpayserver` because BTCPayServer needs to fetch price rates from outside and doing this over Tor isn't supported yet. Maybe we need to implement a Tor transparent proxy and custom routing rules.
* I'm generating a custom macaroon with limited permissions à la https://github.com/Kixunil/cryptoanarchy-deb-repo-builder/blob/216a581f3296bfe37438f13def88a160a2f83509/lnd/gen_macaroons.sh#L34 and placing it inside `/var/lib/lnd/invoice+readonly.macaroon` because `/var/lib/lnd/chain/bitcoin/mainnet` is user-readable-only and I don't want to start changing folder permissions. Maybe there is a better place?
* I added a drive-by fix to `netns-isolation.nix` allowing return traffic to outgoing connections initiated by the service itself in the iptables rules. I think this is harmless.
* I'm using BTCPayServer's internal webserver over a Tor v3 HS. I think we should use a custom nginx with tls when https://github.com/fort-nix/nix-bitcoin/issues/193 is implemented.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,10 @@ env:
   - PKG=elementsd STABLE=0
   - PKG=electrs STABLE=1
   - PKG=electrs STABLE=0
+  - PKG=nbxplorer STABLE=1
+  - PKG=nbxplorer STABLE=0
+  - PKG=btcpayserver STABLE=1
+  - PKG=btcpayserver STABLE=0
   - PKG=liquid-swap STABLE=1
   - PKG=lightning-loop STABLE=0
   - PKG=nixops19_09 STABLE=1
--- a/examples/configuration.nix
+++ b/examples/configuration.nix
@@ -76,6 +76,20 @@
   # sync faster. Only available if hardware wallets are disabled.
   # services.electrs.high-memory = true;
 
+  ### BTCPayServer
+  # Enable this module to use BTCPayServer, a self-hosted, open-source
+  # cryptocurrency payment processor.
+  # Privacy Warning: BTCPayServer currently looks up price rates without
+  # proxying them through Tor. This means an outside observer can correlate
+  # your BTCPayServer usage, like invoice creation times, with your IP address.
+  # services.btcpayserver.enable = true;
+  # Enable this option to connect BTCPayServer to clightning.
+  # services.btcpayserver.lightningBackend = "clightning";
+  # Enable this option to connect BTCPayServert to lnd.
+  # services.btcpayserver.lightningBackend = "lnd";
+  # Afterwards you need to go into Store > General Settings > Lightning Nodes
+  # and click to use "the internal lightning node of this BTCPay Server".
+
   ### LIQUIDD
   # Enable this module to use Liquid, a sidechain for an inter-exchange
   # settlement network linking together cryptocurrency exchanges and
--- a/modules/btcpayserver.nix
+++ b/modules/btcpayserver.nix
@@ -1,0 +1,211 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+
+let
+  cfg = config.services;
+  inherit (config) nix-bitcoin-services;
+in {
+  options.services = {
+    nbxplorer = {
+      package = mkOption {
+        type = types.package;
+        default = pkgs.nix-bitcoin.nbxplorer;
+        defaultText = "pkgs.nix-bitcoin.nbxplorer";
+        description = "The package providing nbxplorer binaries.";
+      };
+      dataDir = mkOption {
+        type = types.path;
+        default = "/var/lib/nbxplorer";
+        description = "The data directory for nbxplorer.";
+      };
+      user = mkOption {
+        type = types.str;
+        default = "nbxplorer";
+        description = "The user as which to run nbxplorer.";
+      };
+      group = mkOption {
+        type = types.str;
+        default = cfg.nbxplorer.user;
+        description = "The group as which to run nbxplorer.";
+      };
+      bind = mkOption {
+        type = types.str;
+        default = "127.0.0.1";
+        description = "The address on which to bind.";
+      };
+      enable = mkOption {
+        # This option is only used by netns-isolation
+        internal = true;
+        default = cfg.btcpayserver.enable;
+      };
+      enforceTor = nix-bitcoin-services.enforceTor;
+    };
+
+    btcpayserver = {
+      enable = mkEnableOption "btcpayserver";
+      package = mkOption {
+        type = types.package;
+        default = pkgs.nix-bitcoin.btcpayserver;
+        defaultText = "pkgs.nix-bitcoin.btcpayserver";
+        description = "The package providing btcpayserver binaries.";
+      };
+      dataDir = mkOption {
+        type = types.path;
+        default = "/var/lib/btcpayserver";
+        description = "The data directory for btcpayserver.";
+      };
+      user = mkOption {
+        type = types.str;
+        default = "btcpayserver";
+        description = "The user as which to run btcpayserver.";
+      };
+      group = mkOption {
+        type = types.str;
+        default = cfg.btcpayserver.user;
+        description = "The group as which to run btcpayserver.";
+      };
+      bind = mkOption {
+        type = types.str;
+        default = "127.0.0.1";
+        description = "The address on which to bind.";
+      };
+      lightningBackend = mkOption {
+        type = types.nullOr (types.enum [ "clightning" "lnd" ]);
+        default = null;
+        description = "The lightning node implementation to use.";
+      };
+      enforceTor = nix-bitcoin-services.enforceTor;
+    };
+  };
+
+  config = mkIf cfg.btcpayserver.enable {
+    assertions = let
+      backend = cfg.btcpayserver.lightningBackend;
+    in [
+      { assertion = (backend != null) -> cfg.${backend}.enable;
+        message = "btcpayserver requires ${backend}.";
+      }
+    ];
+
+    systemd.tmpfiles.rules = [
+      "d '${cfg.nbxplorer.dataDir}' 0770 ${cfg.nbxplorer.user} ${cfg.nbxplorer.group} - -"
+      "d '${cfg.btcpayserver.dataDir}' 0770 ${cfg.btcpayserver.user} ${cfg.btcpayserver.group} - -"
+    ];
+
+    systemd.services.nbxplorer = let
+      configFile = builtins.toFile "config" ''
+        network=mainnet
+        btcrpcuser=${cfg.bitcoind.rpc.users.btcpayserver.name}
+        btcrpcurl=http://${builtins.elemAt config.services.bitcoind.rpcbind 0}:8332
+        btcnodeendpoint=${config.services.bitcoind.bind}:8333
+        bind=${cfg.nbxplorer.bind}
+      '';
+    in {
+      description = "Run nbxplorer";
+      wantedBy = [ "multi-user.target" ];
+      requires = [ "bitcoind.service" ];
+      after = [ "bitcoind.service" ];
+      preStart = ''
+        install -m 600 ${configFile} ${cfg.nbxplorer.dataDir}/settings.config
+        echo "btcrpcpassword=$(cat ${config.nix-bitcoin.secretsDir}/bitcoin-rpcpassword-btcpayserver)" \
+          >> '${cfg.nbxplorer.dataDir}/settings.config'
+      '';
+      serviceConfig = nix-bitcoin-services.defaultHardening // {
+        ExecStart = ''
+          ${cfg.nbxplorer.package}/bin/nbxplorer --conf=${cfg.nbxplorer.dataDir}/settings.config \
+            --datadir=${cfg.nbxplorer.dataDir}
+        '';
+        User = cfg.nbxplorer.user;
+        Restart = "on-failure";
+        RestartSec = "10s";
+        ReadWritePaths = cfg.nbxplorer.dataDir;
+        MemoryDenyWriteExecute = "false";
+      } // (if cfg.nbxplorer.enforceTor
+        then nix-bitcoin-services.allowTor
+        else nix-bitcoin-services.allowAnyIP
+      );
+    };
+
+    systemd.services.btcpayserver = let
+      configFile = builtins.toFile "config" (''
+        network=mainnet
+        socksendpoint=${cfg.tor.client.socksListenAddress}
+        btcexplorerurl=http://${cfg.nbxplorer.bind}:24444/
+        btcexplorercookiefile=${cfg.nbxplorer.dataDir}/Main/.cookie
+        bind=${cfg.btcpayserver.bind}
+      '' + optionalString (cfg.btcpayserver.lightningBackend == "clightning") ''
+        btclightning=type=clightning;server=unix:///${cfg.clightning.dataDir}/bitcoin/lightning-rpc
+      '');
+      lndConfig =
+        "btclightning=type=lnd-rest;" +
+        "server=https://${toString cfg.lnd.listen}:${toString cfg.lnd.restPort}/;" +
+        "macaroonfilepath=/run/lnd/btcpayserver.macaroon;" +
+        "certthumbprint=";
+    in let self = {
+      wantedBy = [ "multi-user.target" ];
+      requires = [ "nbxplorer.service" ]
+                 ++ optional (cfg.btcpayserver.lightningBackend != null) "${cfg.btcpayserver.lightningBackend}.service";
+      after = self.requires;
+      preStart = ''
+        install -m 600 ${configFile} ${cfg.btcpayserver.dataDir}/settings.config
+        ${optionalString (cfg.btcpayserver.lightningBackend == "lnd") ''
+          {
+            echo -n "${lndConfig}";
+            ${pkgs.openssl}/bin/openssl x509 -noout -fingerprint -sha256 -in ${config.nix-bitcoin.secretsDir}/lnd-cert \
+              | sed -e 's/.*=//;s/://g';
+          } >> ${cfg.btcpayserver.dataDir}/settings.config
+        ''}
+      '';
+      serviceConfig = nix-bitcoin-services.defaultHardening // {
+        ExecStart = ''
+          ${cfg.btcpayserver.package}/bin/btcpayserver --conf=${cfg.btcpayserver.dataDir}/settings.config \
+            --datadir=${cfg.btcpayserver.dataDir}
+        '';
+        User = cfg.btcpayserver.user;
+        Restart = "on-failure";
+        RestartSec = "10s";
+        ReadWritePaths = cfg.btcpayserver.dataDir;
+        MemoryDenyWriteExecute = "false";
+      } // (if cfg.btcpayserver.enforceTor
+        then nix-bitcoin-services.allowTor
+        else nix-bitcoin-services.allowAnyIP
+      );
+    }; in self;
+
+    services.lnd.macaroons.btcpayserver = mkIf (cfg.btcpayserver.lightningBackend == "lnd") {
+      inherit (cfg.btcpayserver) user;
+      permissions = ''{"entity":"info","action":"read"},{"entity":"onchain","action":"read"},{"entity":"offchain","action":"read"},{"entity":"address","action":"read"},{"entity":"message","action":"read"},{"entity":"peers","action":"read"},{"entity":"signer","action":"read"},{"entity":"invoices","action":"read"},{"entity":"invoices","action":"write"},{"entity":"address","action":"write"}'';
+    };
+
+    users.users.${cfg.nbxplorer.user} = {
+      description = "nbxplorer user";
+      group = cfg.nbxplorer.group;
+      extraGroups = [ "bitcoinrpc" ];
+      home = cfg.nbxplorer.dataDir;
+    };
+    users.groups.${cfg.nbxplorer.group} = {};
+    users.users.${cfg.btcpayserver.user} = {
+      description = "btcpayserver user";
+      group = cfg.btcpayserver.group;
+      extraGroups = [ "nbxplorer" ]
+                    ++ optional (cfg.btcpayserver.lightningBackend == "clightning") cfg.clightning.user;
+      home = cfg.btcpayserver.dataDir;
+    };
+    users.groups.${cfg.btcpayserver.group} = {};
+
+    services.bitcoind.rpc.users.btcpayserver = {
+      passwordHMACFromFile = true;
+      rpcwhitelist = cfg.bitcoind.rpc.users.public.rpcwhitelist ++ [
+        "setban"
+        "generatetoaddress"
+        "getpeerinfo"
+      ];
+    };
+    nix-bitcoin.secrets.bitcoin-rpcpassword-btcpayserver = {
+      user = "bitcoin";
+      group = "nbxplorer";
+    };
+    nix-bitcoin.secrets.bitcoin-HMAC-btcpayserver.user = "bitcoin";
+  };
+}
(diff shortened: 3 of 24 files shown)


modules/btcpayserver.nix
```diff
@@ -0,0 +1,225 @@
...
+      after = [ "nbxplorer.service" ]
+      ++ (optionals (cfg.btcpayserver.lightning-node == "clightning" ) [ "clightning.service" ])
+      ++ (optionals (cfg.btcpayserver.lightning-node == "lnd" ) [ "lnd.service" ]);
+      preStart = ''
+        cp ${btcpayserver-configFile} ${cfg.btcpayserver.dataDir}/settings.config
+        chown -R '${cfg.btcpayserver.user}:${cfg.btcpayserver.group}' '${cfg.btcpayserver.dataDir}'
+        chmod 600 ${cfg.btcpayserver.dataDir}/settings.config
+        ${optionalString (cfg.btcpayserver.lightning-node == "lnd") ''
+          certthumbprint=$(openssl x509 -noout -fingerprint -sha256 -in ${toString config.nix-bitcoin.secretsDir}/lnd-cert | sed -e 's/.*=//;s/://g')
+          echo -e "btclightning=type=lnd-rest;server=https://${toString cfg.lnd.listen}:${toString cfg.lnd.restPort}/;macaroonfilepath=${toString cfg.lnd.dataDir}/invoice+readonly.macaroon;certthumbprint=$certthumbprint" >> ${cfg.btcpayserver.dataDir}/settings.config
```
Kixunil, line 209: I think you forgot dot? `btc.lightning`

modules/lnd.nix
```diff
@@ -213,6 +213,21 @@ in {
...
+              ${pkgs.curl}/bin/curl -s \
+                -H "Grpc-Metadata-macaroon: $(${pkgs.xxd}/bin/xxd -ps -u -c 99999 '${mainnetDir}/admin.macaroon')" \
+                --cacert ${secretsDir}/lnd-cert \
+                -X POST \
+                -d '{"permissions":[{"entity":"info","action":"read"},{"entity":"onchain","action":"read"},{"entity":"offchain","action":"read"},{"entity":"address","action":"read"},{"entity":"message","action":"read"},{"entity":"peers","action":"read"},{"entity":"signer","action":"read"},{"entity":"invoices","action":"read"},{"entity":"invoices","action":"write"},{"entity":"address","action":"write"}]}' \
+                https://127.0.0.1:${restPort}/v1/macaroon |\
+                ${pkgs.jq}/bin/jq -c '.macaroon' | ${pkgs.xxd}/bin/xxd -p -r > "$invoicereadonlymacaroon"
+            fi
+            chown lnd: "$invoicereadonlymacaroon"
+            chmod 440 "$invoicereadonlymacaroon"
```
Kixunil, line 230: There's a race condition: a different process might read the file before the permissions are restricted. TO fix it, you need to create an empty file with those permissions before you call curl.

This is exactly what [I do](https://github.com/Kixunil/cryptoanarchy-deb-repo-builder/blob/216a581f3296bfe37438f13def88a160a2f83509/lnd/gen_macaroons.sh#L30) + I also write to a temporary file to ensure no corrupted file will be present if the process crashes/power goes down midway.

.travis.yml
```diff
@@ -27,6 +27,10 @@ env:
   - PKG=elementsd STABLE=0
   - PKG=electrs STABLE=1
   - PKG=electrs STABLE=0
+  - pkg=nbxplorer STABLE=1
+  - pkg=nbxplorer STABLE=0
+  - pkg=btcpayserver STABLE=1
+  - pkg=btcpayserver STABLE=0
```
jonasnick, line 33: we won't need this if btcpayserver gets merged upstream

modules/btcpayserver.nix
```diff
@@ -0,0 +1,225 @@
...
+    ]
+    # OpenSSL needed to generate cert fingerprint
+    ++ (optionals (cfg.btcpayserver.lightning-node == "lnd" ) [ openssl xxd ]);
+    services.nbxplorer.enable = true;
+
+    services.bitcoind.rpc.users.btcpayserver = {
+      name = "btcpayserver";
+      # Placeholder to be sed'd out by bitcoind preStart
+      passwordHMAC = "bitcoin-HMAC-btcpayserver";
+      rpcwhitelist = cfg.bitcoind.rpc.users.public.rpcwhitelist ++ [
```
jonasnick, line 128: :+1:

test/test.nix
```diff
@@ -56,6 +56,9 @@ import ./make-test.nix rec {
 
     services.backups.enable = true;
 
+    services.btcpayserver.enable = true;
+    services.btcpayserver.lightning-node = "lnd";
```
jonasnick, line 60: Hm, no clightning test?

modules/lnd.nix
```diff
@@ -213,6 +213,21 @@ in {
             while ! { exec 3>/dev/tcp/127.0.0.1/${toString cfg.rpcPort}; } &>/dev/null; do
               sleep 0.1
             done
+
+            invoicereadonlymacaroon=${cfg.dataDir}/invoice+readonly.macaroon
```
jonasnick, line 217: Would be better to add an `invoicereadonlymacaroon` option to lnd.nix which is set by btcpayserver to with path + permissions and this module creates a corresponding macaroon for you. Afaics the `chown lnd` isn't necessary as `lnd` never reads this. As a sidenote this is a bit annoying because just changing your nix config isn't enough to revoke a macaroon.

modules/btcpayserver.nix
nixbitcoin: No, there is no dot
```
--btclightning              Easy configuration of lightning for the server administrator: Must be a UNIX socket of c-lightning (lightning-rpc) or URL to a charge server (default: empty)
```

modules/netns-isolation.nix
```diff
@@ -182,6 +182,8 @@ in {
             ${ipNetns} route add default via ${bridgeIp}
             ${netnsIptables} -w -P INPUT DROP
             ${netnsIptables} -w -A INPUT -s 127.0.0.1,${bridgeIp},${v.address} -j ACCEPT
+            # allow return traffic to outgoing connections initiated by the service itself
+            ${netnsIptables} -w -A INPUT -m conntrack --ctstate ESTABLISHED -j ACCEPT
```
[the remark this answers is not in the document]
nixbitcoin: Because it's the first module where we disable `enforceTor`. Before, connections were just being routed over Tor on `${bridgeIp}`

test/test.nix
nixbitcoin: If we implement a clightning test we need to create 4 new tests: btcpayserver lnd with netns, btcpayserver lnd without netns, btcpayserver clightning with netns, btcpayserver clightning without netns. Is this complexity really justified seeing that btcpayserver won't test the connection to `lightning-node` until the node is fully synced anyway?

modules/btcpayserver.nix
Kixunil: Ah, you're passing it on the command line, not in the config file.

nixbitcoin: No, we're not. But I assumed it follows the same naming conventions. Apparently BTCPayServer is forgiving when it comes to this because the configuration option works.

modules/lnd.nix
nixbitcoin: Implemented in 94fcf27

modules/lnd.nix
nixbitcoin: Implemented in 94fcf27

modules/bitcoind.nix
```diff
@@ -308,6 +308,7 @@ in {
         cfgpre=$(cat ${configFile}; printf "rpcpassword="; cat "${secretsDir}/bitcoin-rpcpassword-privileged")
         cfg=$(echo "$cfgpre" | \
         sed "s/bitcoin-HMAC-privileged/$(cat ${secretsDir}/bitcoin-HMAC-privileged)/g" | \
+        ${optionalString config.services.btcpayserver.enable ''sed "s/bitcoin-HMAC-btcpayserver/$(cat ${secretsDir}/bitcoin-HMAC-btcpayserver)/g" | \''}
```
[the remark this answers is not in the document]
nixbitcoin: > Architecture-wise it would be much better to avoid any reference to btcpayserver in this file. How about we map all the passwordHMAC options to sed strings instead of manually stating them here?

Implemented ff2981e

> I don't think they're necessary really because our /secrets should be things that are useful to share across machines but this is purely an internal communication between bitcoind and btcpayserver.

I disagree, creating HMAC's locally and distributing them to your node with the correct permissions strikes me as simpler than creating a standalone service to create HMAC's.

test/test.nix
nixbitcoin: Double-checked that BTCPayServer doesn't connect to Lightning Node until you select the proper node in `General Settings` UI, the reason why I use LND is to test the macaroon creation.

examples/configuration.nix
```diff
@@ -76,6 +76,17 @@
   # sync faster. Only available if hardware wallets are disabled.
   # services.electrs.high-memory = true;
 
+  ### BTCPayServer
+  # Enable this module to use BTCPayServer, a self-hosted, open-source
+  # cryptocurrency payment processor.
```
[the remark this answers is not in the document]
nixbitcoin: Implemented in 879d342

.travis.yml
nixbitcoin: Noted

erikarvstedt:
I've added https://github.com/fort-nix/nix-bitcoin/pull/229 which allows us to add a btcpayserver RPC user without modifying `bitcoind.nix`.

erikarvstedt:
@nixbitcoin, if you agree with https://github.com/fort-nix/nix-bitcoin/pull/229, please rebase and I'll jump straight to reviewing this great PR.

nixbitcoin:
Rebased

modules/lnd.nix
```diff
@@ -155,6 +181,8 @@ in {
         ${optionalString cfg.announce-tor "echo externalip=$(cat /var/lib/onion-chef/lnd/lnd) >> '${cfg.dataDir}/lnd.conf'"}
       '';
       serviceConfig = nix-bitcoin-services.defaultHardening // {
+        RuntimeDirectory = "lnd";
+        RuntimeDirectoryMode = "710";
```
erikarvstedt, line 185: At first glance, permission `g=x` won't allow access to btcpayserver. Have you tested that this works?

erikarvstedt (changes requested):
Nice, thanks for tackling this.

Some [fixups](https://github.com/erikarvstedt/nix-bitcoin/commits/btcpayserver-module-ea).

Some style tips, following the nixpkgs coding style:
- Use camelCase for naming. Exceptions are file paths, systemd service names or options that mirror names from upstream packages.
- Keep lines below a reasonable max. length like 120 chars.
- Use `optional cond x` instead of `optionals cond [x]` 

Another issue with the `lnd.macaroons` impl is that all macaroons have the same `lnd` group permissions which defeats the macaroon's purpose to allow for fine-grained access permissions.
Maybe it's better to just add a catch-all public macaroon for now?

modules/lnd.nix
nixbitcoin: `g=x` allows directory read access to users in the lnd group, read permissions on custom macaroons are then individually handled.

modules/btcpayserver.nix
```diff
@@ -0,0 +1,233 @@
...
+    systemd.tmpfiles.rules = [
+      "d '${cfg.nbxplorer.dataDir}' 0770 ${cfg.nbxplorer.user} ${cfg.nbxplorer.group} - -"
+      "d '${cfg.btcpayserver.dataDir}' 0770 ${cfg.btcpayserver.user} ${cfg.btcpayserver.group} - -"
+    ];
+
+    systemd.services.nbxplorer = {
+      description = "Run nbxplorer";
+      path  = with pkgs; [
+        nix-bitcoin.nbxplorer
+        which
```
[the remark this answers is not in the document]
nixbitcoin: Double checked, not needed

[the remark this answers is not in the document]
nixbitcoin: Fixed in 5b6bae3

modules/btcpayserver.nix
```diff
@@ -0,0 +1,233 @@
...
+        cp ${nbxplorer-configFile} ${cfg.nbxplorer.dataDir}/settings.config
+        chown -R '${cfg.nbxplorer.user}:${cfg.nbxplorer.group}' '${cfg.nbxplorer.dataDir}'
+        chmod 600 ${cfg.nbxplorer.dataDir}/settings.config
+        echo "btcrpcpassword=$(cat ${config.nix-bitcoin.secretsDir}/bitcoin-rpcpassword-btcpayserver)" >> '${cfg.nbxplorer.dataDir}/settings.config'
+      '';
+      postStart = ''
+        while [[ ! -e ${cfg.nbxplorer.dataDir}/Main/.cookie ]]; do
+          sleep 0.1
+        done
+        chmod 640 ${cfg.nbxplorer.dataDir}/Main/.cookie
```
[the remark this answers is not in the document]
nixbitcoin: Implemented in fe77af1

nixbitcoin:
> Another issue with the lnd.macaroons impl is that all macaroons have the same lnd group permissions which defeats the macaroon's purpose to allow for fine-grained access permissions.
Maybe it's better to just add a catch-all public macaroon for now?

I added a new `ExecStartPre` that creates macaroons under selected users (7426a00). This allows us to remove the `lnd` `extraGroup` on the `btcpayserver` user, avoiding this security issue.

If you ACK the changes, I can squash all the additional commits and then all that's left is to wait for btcpayserver to get into the upstream nixpkgs.

erikarvstedt:
Final [fixups](https://github.com/erikarvstedt/nix-bitcoin/commits/btcpayserver-module-ea).

- Regarding commit `don't chown -R dataDir`:
We still use `chown -R` in a few services. This is only relevant when data dirs have the wrong ownership after restoring from a backup. I'll add a PR to remove this treewide and add documentation on how to use `/var/lib/nixos/{g,u}id-map` to preserve ownership when restoring updates.

- Regarding `use umask for setting file permissions`:
Include this accoding to your taste. I slightly prefer the `umask` approach, it's a common unix idiom.

- Regarding `lnd: fix mnemonic file access vulnerability`:
Move this commit before `lnd: generate custom macaroons` when rebasing.

It's currently not possible to enable nbxplorer just on its own (`nbxplorer.enable` is a no-op). For fixing this, it would be best to move nbxplorer to its own module.

modules/lnd.nix
erikarvstedt: Yes, you're right! :thumbsup:

nixbitcoin:
ACK `use umask for setting file permissions`, I think it's nice that we don't have to unnecessarily expose write permissions just for one write.

I squashed all your fixups, except `lnd: fix mnemonic file access vulnerability`, they truly round-off this PR.

> It's currently not possible to enable nbxplorer just on its own (nbxplorer.enable is a no-op). For fixing this, it would be best to move nbxplorer to its own module.

I couldn't think of an NBXplorer usecase without BTCPayServer, so I think it's fine we leave them in the same module. I used `config.services.nbxplorer` and `config.services.btcpayserver` for clarity reasons inside the module.

erikarvstedt:
Could you then remove the `nbxplorer.enable` option?

Also, I think we should rename `driveby netns fix: allow return ...` -> `netns: allow return ...`

modules/btcpayserver.nix
```diff
@@ -0,0 +1,242 @@
...
+        type = types.str;
+        default = "http://127.0.0.1:8332";
+        description = "The RPC server url.";
+      };
+      btcnodeendpoint = mkOption {
+        type = types.str;
+        default = "127.0.0.1:8333";
+        description = ''
+          The p2p connection to a Bitcoin node, make sure you are whitelisted
+        '';
```
erikarvstedt, line 43:
```suggestion
        description = "The p2p connection to a Bitcoin node.";
```
The part about whitelisting is a bit irritating, because it doesn't apply in general.
It just refers to a nix-bitcoin implementation detail, namely that our "public" RPC profile isn't sufficient for nbxplorer. But we're already taking care of this with the custom nbxplorer RPC user.

nixbitcoin: I took the description verbatim from `NBXplorer --help`, but I agree with your point. Fixed in 5d150e5

nixbitcoin:
> Could you then remove the `nbxplorer.enable` option?

The `nbxplorer.enable` is needed to make the `netns-isolation` logic create an own network namespace for `nbxplorer`.

> Also, I think we should rename driveby netns fix: allow return ... -> netns: allow return ...

Renamed

erikarvstedt:
We should set `nbxplorer.enable` as internal then, [like so](https://github.com/erikarvstedt/nix-bitcoin/commits/btcpayserver-module-ea).

nixbitcoin:
> We should set `nbxplorer.enable` as internal then, [like so](https://github.com/erikarvstedt/nix-bitcoin/commits/btcpayserver-module-ea).

ACK, this is the best solution. Implemented in 055a935

erikarvstedt:
The upstream nbxplorer/btcpayserver pkgs had some major issues.
I've requested fixups [here](https://github.com/NixOS/nixpkgs/pull/95104#pullrequestreview-482091348) and integrated these fixups into this PR [here](https://github.com/erikarvstedt/nix-bitcoin/commits/btcpayserver-module-ea).
As upstream will probably take a while, we can merge this PR straight away when ready.

modules/netns-isolation.nix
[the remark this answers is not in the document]
jonasnick: I see

modules/bitcoind.nix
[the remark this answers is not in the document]
jonasnick: Simpler in some way but unnecessary in others. Okay for now.

test/test.nix
jonasnick: Hm, yeah that wouldn't help.

modules/lnd.nix
jonasnick: sweet!

> As a sidenote this is a bit annoying because just changing your nix config isn't enough to revoke a macaroon.

It actually is enough with your implementation, because the macaroons are created in a temporary directory on every start of the service.

test/scenarios/withnetns.py
```diff
@@ -18,6 +20,20 @@ def electrs():
     )  # prometeus metrics provider
 
 
+def nbxplorer():
+    machine.wait_until_succeeds("ip netns exec nb-nbxplorer nc -z %s 24444" % nbxplorer_ip)
+
+
+def btcpayserver():
```
jonasnick, line 27: Also add pingtest?

modules/btcpayserver.nix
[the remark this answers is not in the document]
jonasnick: Is that really necessary? The datadir already has 0 permissions for other.

modules/btcpayserver.nix
```diff
@@ -0,0 +1,242 @@
...
+      backend = cfg.btcpayserver.lightningBackend;
+    in [
+      { assertion = (backend != null) -> cfg.${backend}.enable;
+        message = "btcpayserver requires ${backend}.";
+      }
+    ];
+
+    environment.systemPackages = with pkgs.nix-bitcoin; [
+      nbxplorer
+      btcpayserver
```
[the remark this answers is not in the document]
erikarvstedt: Agreed.

erikarvstedt:
> Is that really necessary? The datadir already has 0 permissions for other.

Yeah, with the `dataDir` permissions this is indeed not needed.
(Due to a github bug I couldn't use an inline reply)

Regarding `dataDir`: Why do we need `770` permissions instead of `750`?

nixbitcoin:
> Is that really necessary? The datadir already has 0 permissions for other.

Fixed in b0336be

test/scenarios/withnetns.py
nixbitcoin: Implemented in b0336be

modules/btcpayserver.nix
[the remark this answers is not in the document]
nixbitcoin: Fixed in b0336be

modules/btcpayserver.nix
```diff
@@ -0,0 +1,242 @@
...
+        btcrpcuser=${cfg.bitcoind.rpc.users.btcpayserver.name}
+        btcrpcurl=${cfg.nbxplorer.btcrpcurl}
+        btcnodeendpoint=${cfg.nbxplorer.btcnodeendpoint}
+        bind=${cfg.nbxplorer.bind}
+      '';
+    in {
+      description = "Run nbxplorer";
+      path = with pkgs; [
+        nix-bitcoin.nbxplorer
+        dotnet-sdk_3
```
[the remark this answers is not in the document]
nixbitcoin: Fixed in b0336be

nixbitcoin:
@erikarvstedt I had to correct the lnd server configuration, because breaking it up over multiple lines caused issues with service detection.

nixbitcoin:
I saw that you pushed a couple new commits on https://github.com/erikarvstedt/nix-bitcoin/commits/btcpayserver-module-ea. I decided to squash 3e3fc7ad4f643a80a745d59791704da0bcc8b219 into `temp: mirror erikarvstedt btcpayserver`, because it had several great improvements.

nixbitcoin:
Pulled in newest https://github.com/NixOS/nixpkgs/pull/95104. Split `fixups! jonasnick` into specific fixup commits.

Can we squash and merge?

jonasnick:
Yeah fixups look good feel free to squash

nixbitcoin:
Squashed

erikarvstedt:
@nixbitcoin, why did you revert [this change](https://github.com/erikarvstedt/nix-bitcoin/commits/btcpayserver-module-ea)?

nixbitcoin:
> @erikarvstedt I had to correct the lnd server configuration, because breaking it up over multiple lines caused issues with service detection.

erikarvstedt:
Okay thanks, fixed it [here](https://github.com/erikarvstedt/nix-bitcoin/commits/btcpayserver-module-ea).

nixbitcoin:
How do you run `./update.sh` in the nix-bitcoin pkgs?

erikarvstedt:
Updating is now fixed in the above branch.

nixbitcoin:
Squashed your fixups.

I had to put brackets around the `optionalString` in order for the configs to merge properly:

```diff
diff --git a/modules/btcpayserver.nix b/modules/btcpayserver.nix
index 780d996bf35..6465ab05c0b 100644
--- a/modules/btcpayserver.nix
+++ b/modules/btcpayserver.nix
@@ -128,7 +128,7 @@ in {
     };
 
     systemd.services.btcpayserver = let
-      configFile = builtins.toFile "config" ''
+      configFile = builtins.toFile "config" (''
         network=mainnet
         socksendpoint=${cfg.tor.client.socksListenAddress}
         btcexplorerurl=http://${cfg.nbxplorer.bind}:24444/
@@ -136,7 +136,7 @@ in {
         bind=${cfg.btcpayserver.bind}
       '' + optionalString (cfg.btcpayserver.lightningBackend == "clightning") ''
         btclightning=type=clightning;server=unix:///${cfg.clightning.dataDir}/bitcoin/lightning-rpc
-      '';
+      '');
       lndConfig =
         "btclightning=type=lnd-rest;" +
         "server=https://${toString cfg.lnd.listen}:${toString cfg.lnd.restPort}/;" +
```

erikarvstedt (approved):
ACK c9c844de802947b1e771bb56464253dcbf1a9002

jonasnick (commented):
ACK c9c844de802947b1e771bb56464253dcbf1a9002

PR #240, opened 2020-09-23
push-release: fix bug where assets are not pushed

jonasnick:
In a previous commit we found and replaced $1 with $TAG_NAME, but that wasn't
correct for the post_asset() function.


erikarvstedt (approved):
utACK

Sorry for introducing this bug, should have set up a private repo for testing.

µ-nit: Rename commit to `push-release: ...`

jonasnick:
fixed nit

erikarvstedt (approved):
utACK 0c1fb9fd4497db0619d779c9cbfe37cdd3ac355a

PR #243, opened 2020-09-24
Modularize Tests

erikarvstedt:
This improves debugging and experimenting by making it easy to compose fine-grained
scenarios that have specific tests and features enabled.

Scenarios can now be defined like this:
```nix
# A node with two services. The corresponding tests are automatically enabled.
simple = {
  services.electrs.enable = true;
  services.clightning.enable = true;
};

# Full netns scenario, but with only two tests enabled
netnsTwoTests = {
   imports = [ scenarios.netns ];
   tests = mkForce {
     btcpayserver = true;
     netns-isolation = true;
   };
 };
```

Custom scenarios can conveniently be added by setting the env variable `scenarioOverridesFile`.

For undefined scenarios the tests are run with an adhoc scenario where `services.<scenario>` is enabled:
```
run-tests.sh -s clightning
run-tests.sh -s electrs
```

Also, the test output now includes the subtest name and duration and subtests use nested logging, which improves the html log.

### Containers
Scenarios can now be run in containers (for now, without the python test suite) for super fast
debugging and experimenting.
Like in the examples, creating NixOS containers needs root permissions.

```bash
## Run single command in container, delete container afterwards
run-tests.sh -s bitcoind container --run c systemctl status bitcoind

## Start shell session in container, delete container afterwards
run-tests.sh -s bitcoind container
# Run command in container
c systemctl status bitcoind
# Run via SSH
cssh systemctl status bitcoind
# enter container
c
# enter via SSH
cssh
```

### Review helper

Use the following to audit https://github.com/fort-nix/nix-bitcoin/pull/243/commits/45485760cdad98a6e3e26176c695c55393db23ad `(extra-container: pre-release -> 0.4)`
```
tmp=$(mktemp -d -p /tmp)
mkdir -p $tmp/{old,new}
fetchRelease() { curl -SL https://github.com/erikarvstedt/extra-container/archive/$1.tar.gz | tar xz --strip-components=1 -C $2;}
fetchRelease 6cced2c26212cc1c8cc7cac3547660642eb87e71 $tmp/old
fetchRelease 0.4 $tmp/new
git diff --no-index $tmp/{old,new}
rm -rf $tmp
```

PR #245, opened 2020-09-25
Fix tests without secure-node

erikarvstedt:
This PR is based on [#243 Modularize Tests](https://github.com/fort-nix/nix-bitcoin/pull/243).
It fixes the tests for scenarios without secure-node.

The following is now working:
```
test/run-tests.sh -s full
test/run-tests.sh -s bitcoind

cat > /tmp/scenarios <<EOF
{...}: {
  custom = {
    services.lnd.enable = true;
  };
}
EOF
scenarioOverridesFile=/tmp/scenarios test/run-tests.sh -s custom
```

### DIFF
--- a/test/run-tests.sh
+++ b/test/run-tests.sh
@@ -16,8 +16,8 @@
 #     Example:
 #     ./run-tests.sh -s electrs
 #
-#   Run test and link results to avoid garbage collection
-#   ./run-tests.sh [--scenario <scenario>] --out-link-prefix /tmp/nix-bitcoin-test build
+#   Run test(s) and link results to avoid garbage collection
+#   ./run-tests.sh [--scenario <scenario>] --out-link-prefix /tmp/nix-bitcoin-test
 #
 #   Pass extra args to nix-build
 #   ./run-tests.sh build --builders 'ssh://mybuildhost - - 15'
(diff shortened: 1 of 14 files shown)


nixbitcoin:
Concept ACK.

Once https://github.com/fort-nix/nix-bitcoin/pull/243 has been merged you should rebase, and then I'll ACK.

test/run-tests.sh
```diff
@@ -130,32 +149,54 @@ exprForCI() {
...
+    scenario=default buildTest "$@"
+    scenario=netns buildTest "$@"
+    scenario=full evalTest "$@"
+}
+
+all() {
+    scenario=default buildTest "$@"
+    scenario=netns buildTest "$@"
+    scenario=full buildTest "$@"
+}
```
jonasnick, line 184: 2 min 30 vs 3 min 15 on my laptop so we can switch to all by default in the future if we miss issues with the full test.

jonasnick (commented):
Awesome. Looks good to me so far.

erikarvstedt:
Rebased to master.

jonasnick (approved):
ACK bfed10b2faea4ed43d21936b4564e799c669ae0d

PR #247, opened 2020-10-12
Add module 'versioning'

erikarvstedt:
This PR adds option `nix-bitcoin.configVersion` to inform users about backwards-incompatible changes and to provide migration hints.

### Example
A notification about the recent lightning-loop update could have been implemented like this:
```nix
{
  version = "0.0.18";
  condition = config.services.lightning-loop.enable;
  message = ''
    The lightning-loop data dir location was changed.
    To move it to the new location, run the following shell command on your nix-bitcoin node:
    sudo mv ${config.services.lnd.dataDir}/.loop ${config.services.lightning-loop.dataDir}
  '';
}
```
To see it in action, checkout [this example branch](https://github.com/erikarvstedt/nix-bitcoin/commits/versioning-example) and run `test/run-tests.sh eval`

Example output:
```
This version of nix-bitcoin contains the following changes
that are incompatible with your config (version 0.0.17):

- The lightning-loop data dir location was changed.
To move it to the new location, run the following shell command on your nix-bitcoin node:
sudo mv /var/lib/lnd/.loop /var/lib/lightning-loop
(This change was introduced in version 0.0.18)

- dummy
(This change was introduced in version 0.1)

After addressing the above changes, set nix-bitcoin.configVersion = "0.1";
in your nix-bitcoin configuration.
```

### Details

Note that `nix-bitcoin.configVersion` in `examples/configuration.nix` only needs to
be updated after adding new breaking changes.

### DIFF
--- a/modules/versioning.nix
+++ b/modules/versioning.nix
@@ -1,0 +1,59 @@
+{ config, pkgs, lib, ... }:
+
+with lib;
+let
+  version = config.nix-bitcoin.configVersion;
+
+  # Sorted by increasing version numbers
+  changes = [
+    # None yet
+    # {
+    #   version = "0.1";
+    #   condition = config.services.foo.enabled;
+    #   message = ''
+    #     demo message
+    #   '';
+    # }
+  ];
+
+  incompatibleChanges = optionals
+    (version != null && versionOlder lastChange)
+    (builtins.filter (change: versionOlder change && (change.condition or true)) changes);
+
+  errorMsg = ''
+
+    This version of nix-bitcoin contains the following changes
+    that are incompatible with your config (version ${version}):
+
+    ${concatMapStringsSep "\n" (change: ''
+      - ${change.message}(This change was introduced in version ${change.version})
+    '') incompatibleChanges}
+    After addressing the above changes, set nix-bitcoin.configVersion = "${lastChange.version}";
+    in your nix-bitcoin configuration.
+  '';
+
+  versionOlder = change: (builtins.compareVersions change.version version) > 0;
+  lastChange = builtins.elemAt changes (builtins.length changes - 1);
+in
+{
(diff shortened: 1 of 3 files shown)


nixbitcoin:
Strong Concept ACK. I think this is pretty much the ideal fix for our release migration issues.

Could we also throw alerts on new features that may interest a user who's forked off the `examples/configuration.nix`?

erikarvstedt:
home-manager implements news alerts [like this](https://github.com/nix-community/home-manager/blob/master/modules/misc/news.nix). The date of the latest news that was read is stored inside `$HOME`.

In nix-bitcoin, the news items to be shown could be determined by `configVersion`, analogous to the breaking change messages.
We could certainly add this later on.

modules/versioning.nix
```diff
@@ -0,0 +1,59 @@
...
+    #   condition = config.services.foo.enabled;
+    #   message = ''
+    #     demo message
+    #   '';
+    # }
+  ];
+
+  incompatibleChanges = optionals
+    (version != null && versionOlder lastChange)
+    (builtins.filter (change: versionOlder change && (change.condition or true)) changes);
```
jonasnick, line 21: what does this "or true" do?

jonasnick (commented):
That's super helpful. ACK mod nit.

modules/versioning.nix
erikarvstedt: It returns `true` if the change has no `condition` attribute.

jonasnick: I see. Hadn't noticed that construction before.

jonasnick:
Needs rebase

jonasnick (approved):
ACK d3ece5991953e16e287c5c988496bd6dd8cf861c

PR #249, opened 2020-10-16
Add regtest support

erikarvstedt:
Regtest allows testing services in a synced-blockchain scenario.

Run the test via `test/run-tests.sh -s regtest`.

Besides enabling in-depth service testing, this is also a great consistency check for our service modules. The number of service bugfixes included in this PR is testament to this.

`regtest` revealed that lightning-loop always requires `server.host` to be set. Should we add an option for it?

### DIFF
--- a/modules/bitcoind.nix
+++ b/modules/bitcoind.nix
@@ -11,7 +11,10 @@ let
     # We're already logging via journald
     nodebuglogfile=1
 
-    ${optionalString cfg.testnet "testnet=1"}
+    ${optionalString cfg.regtest ''
+      regtest=1
+      [regtest]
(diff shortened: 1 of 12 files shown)


erikarvstedt:
Rebased to master.

erikarvstedt:
Ah, `server.host` is missing a [default](https://github.com/lightninglabs/loop/blob/96f395e26309d2227eab4645e96e4ab276a92e15/loopd/config.go#L126) only on regtest, so an option is not needed.

jonasnick (approved):
ACK 9951f10e745948210e217d4b0f0cefc147ef8769

PR #251, opened 2020-10-18
Services: Auto-enable dependencies

erikarvstedt:
Also, add missing dependencies to bitcoind, because `bitcoind.enable` is `false` outside of `secure-node.nix`.

### Comparison: auto-enable vs. using assertions
Pros:
- Less code
- Adheres to the NixOS philosophy that services should automatically enable their dependencies
- Works like [cryptoanarchy-deb-repo-builder](https://github.com/Kixunil/cryptoanarchy-deb-repo-builder), where, e.g., `sudo apt install btcpayserver` automatically installs all deps, including `lnd`.
- Enables auto-scenarios for all services, like `run-tests.sh -s btcpayserver` or `run-tests.sh -s spark-wallet`

Cons:
- The user might implicitly enable a lightning service and forget to back up its data. This is taken care of by the backup module.

### Details
Note that we can't use the following elegant definition in btcpayserver:
```
services.${cfg.btcpayserver.lightningBackend}.enable = true;
```
This leads to infinite recursion due to the intricacies of module evaluation.
Instead, we use
```
services.clightning.enable = mkIf (cfg.btcpayserver.lightningBackend == "clightning") true;
services.lnd.enable = mkIf (cfg.btcpayserver.lightningBackend == "lnd") true;
```

### DIFF
--- a/examples/configuration.nix
+++ b/examples/configuration.nix
@@ -65,7 +65,7 @@
   ### SPARK WALLET
   # Enable this module to use spark-wallet, a minimalistic wallet GUI for
   # c-lightning, accessible over the web or through mobile and desktop apps.
-  # Only available if clightning is enabled.
+  # Automatically enables clightning.
   # services.spark-wallet.enable = true;
 
   ### ELECTRS
@@ -87,6 +87,7 @@
   # services.btcpayserver.lightningBackend = "clightning";
   # Enable this option to connect BTCPayServert to lnd.
   # services.btcpayserver.lightningBackend = "lnd";
+  # The lightning backend service automatically enabled.
   # Afterwards you need to go into Store > General Settings > Lightning Nodes
   # and click to use "the internal lightning node of this BTCPay Server".
 
@@ -99,13 +100,14 @@
 
   ### LIGHTNING CHARGE
   # Enable this module to use lightning-charge, a simple drop-in solution for
-  # accepting lightning payments. Only available if clightning is enabled.
+  # accepting lightning payments.
+  # Automatically enables clightning.
   # services.lightning-charge.enable = true;
 
   ### NANOPOS
   # Enable this module to use nanopos, a simple Lightning point-of-sale
-  # system, powered by Lightning Charge. Only available if clightning and
-  # lightning-charge are enabled.
+  # system, powered by Lightning Charge.
+  # Automatically enables lightning-charge.
   # services.nanopos.enable = true;
 
   ### WEBINDEX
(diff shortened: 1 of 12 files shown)


erikarvstedt:
Travis just had a [spurious failure](https://travis-ci.org/github/fort-nix/nix-bitcoin/jobs/736796742): The CI build just stopped while running the VM test, but the test itself didn't fail.
Solved by re-pushing the commit with a changed date.

nixbitcoin:
If this is the NixOS way to do it, I think we should be consistent with NixOS user expectations. So concept ACK.

@jonasnick stated in https://github.com/fort-nix/nix-bitcoin/pull/189#issuecomment-643805100 referring to modules implicitly enabling other modules, "I'd prefer to not do that to allow users to easily see and opt-in to what software they actually use on their node (for now at least... the number of modules is still manageable)." Has your opinion changed?

nixbitcoin:
We should also probably removed the `Only available if` comments in `examples/configuration.nix`

erikarvstedt:
Thanks, I've updated `examples/configuration.nix`.
I've left the webindex section unchanged. Fixing webindex to better adapt to the enabled services is on my to-do list.

jonasnick:
It's fine either way. I'd personally still prefer being reminded of the dependencies but I can see that this is generally unnecessary. And if everyone has to manually enable the dependencies that's a lot of code duplication in a sense. Also, it's well documented now.

nixbitcoin (approved):
```
Hash: SHA512

ACK 67e49fe4154ef188f920f1a7cd2f16bdf4ed39bb
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCgAdFiEEV3o0Un8+KoXoD+Fk3RH5rVMIs7oFAl+Ov1kACgkQ3RH5rVMI
s7rhhxAAlW5XJY3OvjCbKb8uFpKp6CM1RqKslrUR2YT8PYSOZL1hyHltAJHV7fPr
z0MK018CmuSCpj3xzpkb8pXNjDibduXxkCxyaBn64KbxvRs5huxrI/nFyZaoGrtM
MPrIdmBAvezt2fXOUqLCPPfdUM7YKTnxRnwzivadsLtJ3lO7tuXSMphffnQtkOKj
W6shf+VzkzHwHO1g8xoJZ30whY9i+x674rMG8uDQrLbHP90bS3LfiE522UWfXh4O
8EgJrXUDZanFEkl9iLjhNtM1VuMYvWQXEoQUCCOeOCeWNxMLP2LCQ7XZtPvuOxYU
PFTmH4SE7hu9J//9KXLNddAVBcz1ywdJjC8Ju3W1TEuH4624LwWQ3VxHBhL0dKIh
u0Iyd/CWZeOtOnftMhxk35DmrS5SGwcF8W1w7RxuSdExRu22WguwbLcUzpMwAGvn
dZe7c57QEy6Qe/H5Ve0fOAxRxYYI25ND92IFAESdoDhDSsSSLd1oqgixE/F7DGyK
mTiiHhpqNjhA6v/LFAkAfqMTbtAcwm87aw2eNXzo9Fd7CwJMwwms/5Gn9a2VmGZ5
Qt9NTsx8MyK6lCY5Dof0dDnv3vYYjbw/i4tM6uXyMgcmsEDCwmGynfVfUyCvGeRG
zN80hrFzhCINpo+8gHksLc8wlN6RGBQD2BEbUt/NjqAFWeVRQBk=
=1XtD
-----END PGP SIGNATURE-----
```

PR #254, opened 2020-10-19
shell.nix: fix failing generate-secrets

### LINKED ISSUE #253: rpcauth script fails
```
$ cd examples
$ rm -rf ./secrets
$ nix-shell
Traceback (most recent call last):
  File "/nix/store/g7d8s8kywpigdbzp8r5n2gzqy5m903wb-rpcauth.py", line 6, in <module>
    from argparse import ArgumentParser
  File "/nix/store/r94aa2gj4drkhfvkm2p4ab6cblb6kxlq-python3-3.7.6/lib/python3.7/argparse.py", line 87, in <module>
    import re as _re
  File "/nix/store/r94aa2gj4drkhfvkm2p4ab6cblb6kxlq-python3-3.7.6/lib/python3.7/re.py", line 143, in <module>
    class RegexFlag(enum.IntFlag):
AttributeError: module 'enum' has no attribute 'IntFlag'
```
As a result, the HMAC files in the secrets directory are empty.

erikarvstedt:
generate-secrets failed due to the `PYTHONPATH` env var that was set through the nixops19_09 buildInput.

Fixes:

- Don't pull in the build environments of binaries that should be available in `PATH`. Only extend `PATH` instead.
- Run generate-secrets in an empty environment

Fixes #253

### DIFF
--- a/examples/shell.nix
+++ b/examples/shell.nix
@@ -20,10 +20,12 @@ with nixpkgs;
 stdenv.mkDerivation rec {
   name = "nix-bitcoin-environment";
 
-  buildInputs = [ nix-bitcoin.nixops19_09 nix-bitcoin.extra-container figlet ];
+  path = lib.makeBinPath [ nix-bitcoin.nixops19_09 nix-bitcoin.extra-container figlet ];
 
   shellHook = ''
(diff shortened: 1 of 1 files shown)


jonasnick (approved):
Thanks! I wouldn't have suspected the buildInputs. 

ACK d2dbad256f8aa7c8870672cf72408deaec584175

PR #255, opened 2020-10-20
Improve netns-isolation and Tor config

erikarvstedt:
## Topic: Allow RPC access from the main netns
This is implemented by commits starting with `netns-bitcoind: allow RPC access from main netns`.

Pros:
- Much reduced implementation complexity: No extra CLI, no multiple service listen addresses.
- Improved usability: The `service.cli` binary will always work from all allowed namespaces, including the service's own netns.\
  The previous CLI separation led to a bug in `regtestBase` because I inadvertently used `bitcoind.cli` instead of `bitcoind.cliBase` in `bitcoind.postStart` for generating regtest blocks.

Cons:
- Insignificantly decreased security for bitcoind and liquidd (not for lnd):\
  Previously, attackers from the main netns had to pass the allowip and the rpcauth
  checks to gain RPC access.\
  Now, it's just rpcauth, like for connections from all other allowed netns.

### DIFF
--- a/modules/bitcoind.nix
+++ b/modules/bitcoind.nix
@@ -182,7 +182,7 @@ in {
       };
       proxy = mkOption {
         type = types.nullOr types.str;
-        default = null;
+        default = if cfg.enforceTor then config.services.tor.client.socksListenAddress else null;
         description = "Connect through SOCKS5 proxy";
       };
       listen = mkOption {
--- a/modules/bitcoind.nix
+++ b/modules/bitcoind.nix
@@ -39,8 +39,8 @@ let
           "rpcwhitelist=${user.name}:${lib.strings.concatStringsSep "," user.rpcwhitelist}"}
       '') (builtins.attrValues cfg.rpc.users)
     }
-    ${lib.concatMapStrings (rpcbind: "rpcbind=${rpcbind}\n") cfg.rpcbind}
-    rpcconnect=${builtins.elemAt cfg.rpcbind 0}
+    rpcbind=${cfg.rpcbind}
+    rpcconnect=${cfg.rpcbind}
     ${lib.concatMapStrings (rpcallowip: "rpcallowip=${rpcallowip}\n") cfg.rpcallowip}
 
     # Wallet options
@@ -150,8 +150,8 @@ in {
         description = "Set the number of threads to service RPC calls";
       };
       rpcbind = mkOption {
-        type = types.listOf types.str;
-        default = [ "127.0.0.1" ];
+        type = types.str;
+        default = "127.0.0.1";
         description = ''
           Bind to given address to listen for JSON-RPC connections.
         '';
--- a/modules/bitcoind.nix
+++ b/modules/bitcoind.nix
@@ -40,6 +40,7 @@ let
       '') (builtins.attrValues cfg.rpc.users)
     }
     ${lib.concatMapStrings (rpcbind: "rpcbind=${rpcbind}\n") cfg.rpcbind}
+    rpcconnect=${builtins.elemAt cfg.rpcbind 0}
     ${lib.concatMapStrings (rpcallowip: "rpcallowip=${rpcallowip}\n") cfg.rpcallowip}
 
     # Wallet options
@@ -275,17 +276,12 @@ in {
         description = "What type of addresses to use";
       };
       cli = mkOption {
-        type = types.package;
-        # Overriden on netns-isolation
-        default = cfg.cliBase;
-        description = "Binary to connect with the bitcoind instance.";
-      };
-      cliBase = mkOption {
         readOnly = true;
         type = types.package;
         default = pkgs.writeScriptBin "bitcoin-cli" ''
           exec ${cfg.package}/bin/bitcoin-cli -datadir='${cfg.dataDir}' "$@"
         '';
+        description = "Binary to connect with the bitcoind instance.";
       };
       enforceTor =  nix-bitcoin-services.enforceTor;
     };
@@ -341,9 +337,8 @@ in {
         fi
       '';
       postStart = ''
-        cd ${cfg.cliBase}/bin
         # Poll until bitcoind accepts commands. This can take a long time.
-        while ! ./bitcoin-cli getnetworkinfo &> /dev/null; do
+        while ! ${cfg.cli}/bin/bitcoin-cli getnetworkinfo &> /dev/null; do
           sleep 1
         done
       '';
@@ -368,7 +363,7 @@ in {
       bindsTo = [ "bitcoind.service" ];
       after = [ "bitcoind.service" ];
       script = ''
-        cd ${cfg.cliBase}/bin
+        cd ${cfg.cli}/bin
         echo "Importing node banlist..."
         cat ${./banlist.cli.txt} | while read line; do
             if ! err=$(eval "$line" 2>&1) && [[ $err != *already\ banned* ]]; then
--- a/modules/netns-isolation.nix
+++ b/modules/netns-isolation.nix
@@ -141,6 +141,7 @@ in {
         inherit (v) netnsName;
         ipNetns = "${ip} -n ${netnsName}";
         netnsIptables = "${ip} netns exec ${netnsName} ${config.networking.firewall.package}/bin/iptables";
+        allowedAddresses = concatMapStringsSep "," (available: netns.${available}.address) v.availableNetns;
       in {
         "${n}".serviceConfig.NetworkNamespacePath = "/var/run/netns/${netnsName}";
 
@@ -165,15 +166,13 @@ in {
             ${netnsIptables} -w -A INPUT -s 127.0.0.1,${bridgeIp},${v.address} -j ACCEPT
             # allow return traffic to outgoing connections initiated by the service itself
             ${netnsIptables} -w -A INPUT -m conntrack --ctstate ESTABLISHED -j ACCEPT
-          '' + (optionalString (config.services.${n}.enforceTor or false)) ''
+          '' + optionalString (config.services.${n}.enforceTor or false) ''
             ${netnsIptables} -w -P OUTPUT DROP
             ${netnsIptables} -w -A OUTPUT -d 127.0.0.1,${bridgeIp},${v.address} -j ACCEPT
-          '' + concatMapStrings (otherNetns: let
-            other = netns.${otherNetns};
-          in ''
-            ${netnsIptables} -w -A INPUT -s ${other.address} -j ACCEPT
-            ${netnsIptables} -w -A OUTPUT -d ${other.address} -j ACCEPT
-          '') v.availableNetns;
+          '' + optionalString (v.availableNetns != []) ''
+            ${netnsIptables} -w -A INPUT -s ${allowedAddresses} -j ACCEPT
+            ${netnsIptables} -w -A OUTPUT -d ${allowedAddresses} -j ACCEPT
+          '';
           preStop = ''
             ${ip} netns delete ${netnsName}
           '';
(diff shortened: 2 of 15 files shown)


tuananh11091992 (approved):

erikarvstedt:
I'd love to welcome @tuananh11091992 to this repo, but, unfortunately, it's a [spam](https://github.com/lnbook/lnbook/issues/534) [account](https://github.com/bitcoin/bitcoin/issues/20134). :disappointed: Blocked and reported.

modules/bitcoind.nix
```diff
@@ -182,7 +183,7 @@ in {
       };
       proxy = mkOption {
         type = types.nullOr types.str;
-        default = null;
+        default = if cfg.enforceTor then config.services.tor.client.socksListenAddress else null;
```
nixbitcoin, line 186: I don't agree with linking the proxy setting with `cfg.enforceTor`. There are situations in which a user might want to use a Tor proxy but still not enforceTor, for example when he wants Bitcoind to communicate out over Tor, but accept incoming clearnet connections.

erikarvstedt: This use case is still possible, as before.
This setting just automatically sets `proxy` when `enforceTor == true`. For the mentioned use case the user would just set the `proxy` manually and leave `enforceTor` disabled.

modules/netns-isolation.nix
```diff
@@ -252,46 +252,29 @@ in {
...
       listen = netns.lnd.address;
-      rpclisten = [
-        "${netns.lnd.address}"
-        "127.0.0.1"
-      ];
-      restlisten = [
-        "${netns.lnd.address}"
-        "127.0.0.1"
-      ];
-      cliExec = mkCliExec "lnd";
```
nixbitcoin, line 282: Is there a reason why you left lightning-loop's `cliExec`?

nixbitcoin:
I think we should keep `netns-exec` for JoinMarket because it's useful to enforceTor even on the client scripts. Otherwise, we can remove all the `allowed_netns` from ` nix-bitcoin/pkgs/netns-exec/src/main.c `

modules/netns-isolation.nix
erikarvstedt: For lightning-loop we would have to expose the RPC port to the main netns, which would theoretically add a new class of attack vectors. So the security argument isn't as clear-cut as for the other services.
Personally, I'd definitely favor the "main netns" approach for `lightning-loop`, now that macaroon-auth is supported. So if you agree, I'd happily add it.

erikarvstedt:
Yes, I'd favor leaving JoinMarket as it is, too.

modules/bitcoind.nix
nixbitcoin: Makes sense

modules/netns-isolation.nix
jonasnick: Isn't the RPC port already exposed to the main netns in the same way bitcoind's RPC port is exposed? It seems you can remove netns-exec from `mkCliExec` and the tests still pass.

jonasnick (commented):
Concept ACK. This is indeed an insignificant change in terms of security given that it's already possible to communicate to the service netns from the main netns.

On the other hand, _should_ we try to prevent connections from the main netns on the network layer? Looking at the complexity and the miniscule improvement in security, this makes only sense for services that do not have authentication, like the old lightning-loop. But would be even better to avoid adding those services in the future.

erikarvstedt:
I've added main netns RPC access for lightning-loop. Here's the [unsquashed version](https://github.com/erikarvstedt/nix-bitcoin/commits/netns-and-tor-unsquashed) of this branch with full internal history.

erikarvstedt:
> this makes only sense for services that do not have authentication

Yes, I fully agree.

modules/netns-isolation.nix
erikarvstedt: No, the RPC server listens on the netns localhost.
RPC connections to loop can only be checked with regtest because the server doesn't start before bitcoind has finished syncing.
Scenario `netnsRegtest` (with its [latest fix](https://github.com/erikarvstedt/nix-bitcoin/commit/be546abdb970815ab52514f230ce0ab242558f20)) fails when removing `cliExec`.

test/tests.py
```diff
@@ -237,50 +237,34 @@ def _():
...
-        + "%s %s ||" % (ping_nanopos, ip("lnd"))
-        + "%s %s ||" % (ping_nanopos, ip("lightning-loop"))
-        + "%s %s ||" % (ping_nanopos, ip("liquidd"))
-        + "%s %s ||" % (ping_nanopos, ip("electrs"))
-        + "%s %s ||" % (ping_nanopos, ip("spark-wallet"))
-        + "%s %s" % (ping_nanopos, ip("btcpayserver"))
-    )
+    # These service lists are non-exhaustive
+    reachable_from_bitcoind = get_ips("clightning", "lnd", "liquidd", "nbxplorer")
+    unreachable_from_bitcoind = get_ips("lightning-loop", "btcpayserver", "spark-wallet", "nginx")
```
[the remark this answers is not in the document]
erikarvstedt: `fping` makes no difference here, because we have to run a separate ping process in each tested netns. But in general, positive ping tests are indeed cheap.

I'd still vote for keeping the test unchanged:
1. `availableNetns` is symmetrised [here](https://github.com/fort-nix/nix-bitcoin/blob/c0d04b908128f40bac41b995bd00a44ee5c45016/modules/netns-isolation.nix#L32).
2. Even if a netns connection should fail, the related positive ping test that would uncover the failure will never be executed:
   We're implicitly testing all netns connections in the service tests, so these tests would fail before the ping test.
   Only the negative ping tests can meaningfully help exposing defects.

test/tests.py
```diff
@@ -237,50 +237,34 @@ def _():
 # (and their corresponding network namespaces).
```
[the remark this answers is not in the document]
erikarvstedt: I don't see a way how we could accidentally assign these service addresses in the main netns so they could be bindable.
But we could add a rationale comment to `netns-isolation.nix` that mentions this use case.
You decide: 1. add test 2. add comment later.

modules/netns-isolation.nix
jonasnick: Right.

test/tests.py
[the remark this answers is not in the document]
jonasnick: Seems like we're talking past each other. I'm suggesting a negative ping test from lightning-loop to bitcoind, similar to how it worked before (nanopos to bitcoind, etc). But you're right that this would cost 100ms in every netns from which we'd run this.

test/tests.py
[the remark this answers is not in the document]
jonasnick: Agree that it's quite a stretch that we'd add an interface with that ip to the main netns. Sometimes tests can be pretty useful documentation too. And yeah, we should add rationale as to what netns is supposed to achieve in any case.

jonasnick (commented):
new changes look good to me

test/tests.py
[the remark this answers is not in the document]
nixbitcoin: I would also favor symmetrizing the tests. There could always be a bug that exposes the netns from one side.

test/tests.py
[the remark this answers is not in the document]
nixbitcoin: I'd favor adding the test.

nixbitcoin:
`netns-isolation` goals:
- [x] Minimize nix-bitcoin services' network exposure between eachother
- [x] Additional Tor enforcement on a netns level
- [x] Limit access to RPC interface of unauthenticated services (JoinMarket atm)
- [ ] Minimize nix-bitcoin services' network exposure from system services running in the main netns

I think goal 4 is desirable and should be implemented. Nonetheless, I think we can go forward with merging this PR because it doesn't open any new holes, just simplifies what we already have.

nixbitcoin:
`lightning-loop` tested. Open points are
* symmetrizing negative ping tests
* `machine.fail("nc -l %s 8332" % ip(bitcoind))`

erikarvstedt:
To reproduce the error fixed by commit `netns-isolation: fix error when stopping netns`, run

```bash
git checkout master

patch -p1 <<'EOF'
--- a/modules/netns-isolation.nix
+++ b/modules/netns-isolation.nix
@@ -175,5 +175,6 @@ in {
           preStop = ''
             ${ip} netns delete ${netnsName}
+            sleep 0.1
             ${ip} link del ${peer}
           '';
           serviceConfig = {
EOF

test/run-tests.sh -s netns container --run c bash -c '
  systemctl stop netns-electrs
  journalctl -u netns-electrs
'
```
Output:
```
systemd[1]: Starting netns-electrs.service...
ip[379]: Cannot remove namespace file "/var/run/netns/nb-electrs": No such file or directory
systemd[1]: Started netns-electrs.service.
systemd[1]: Stopping netns-electrs.service...
9sn3dqfhzb009bd3f06sqxx38fl4hbs7-unit-script-netns-electrs-pre-stop[10318]: Cannot find device "nb-veth-br-16"
systemd[1]: netns-electrs.service: Control process exited, code=exited, status=1/FAILURE
systemd[1]: netns-electrs.service: Failed with result 'exit-code'.
```
The `Cannot remove namespace file` message is fixed by `netns: fix error msg when starting netns`.

erikarvstedt:
As always, I'm keeping the [unsquashed branch](https://github.com/erikarvstedt/nix-bitcoin/commits/netns-and-tor-unsquashed) up-to-date.

jonasnick (approved):
Nice!

ACK b4b607dfa56eaa25a7af5eca55b3149322335db6

nixbitcoin (approved):
```
Hash: SHA512

ACK b4b607dfa56eaa25a7af5eca55b3149322335db6
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCgAdFiEEV3o0Un8+KoXoD+Fk3RH5rVMIs7oFAl+gJKUACgkQ3RH5rVMI
s7qr4xAAntiGHvIS1gvnxnvFjDZo36HulgxtV3VQ7zVgdz97H7hG/66WRdRH9sPl
I8GIbUuCzNOSgVu4ayatBZ9HEzsXrX2S9U63l4XKRvVaF1dLkDUAAtFbKxBUFG8S
WgOSPWVhzY2JXLJPV9PDFUNl0hn7TprWJdx/SW65LAGh0EZ1ymsOcVF1wc8VUHQ1
MjcYVOkQ6t0DBq8zgLG34cIu3KQ3rzovrowRecKMwE8IHAgelOroJfMdqJf+N1uh
BblYMyaZIqhFAu07EmqJRSLMF0yrnoaWivXLPpiWvrzLENheUrmyT3N+3Kgu9cgn
+j9U0/ysN3erFJ+r6oyTUSek8s9Vvjw4LRiuby/PplxILmbf7T7QlDpzjPJ9BGfL
rYsTmZWi5bcFtZuEzXmxE45EGjfOikueIKpv6FucD0cWM+aIa1QLR5GTH4aIP/iR
w/B3NnkfskzuIZj9cmZLCVILPXH2kOB/wAkGXPnxkiP+E+JZOD2/Pw/9KxvT52Zc
qh7I/o6NH1VQOpBgfuHBWdRi5i/uPkjK7xaEkxzsIYmCv4sDQW19TNbYx29NRBUm
P6NztxQg3lxZxtUHh8XZRh5f22ZVLcFeAOtL2vNzhFDYDakTbnyJi61ehN9f+wnK
b5F8J6LBLalUAltlTapV3HsQ1e+ABq6t9ubxyD81n7p1JFmSoJc=
=vfBH
-----END PGP SIGNATURE-----
```

PR #258, opened 2020-10-29
joinmarket 0.7.0 -> 0.7.2 & lightning-loop 0.10.0-beta -> 0.11.0-beta

nixbitcoin:
joinmarket 0.7.0 -> 0.7.2 & lightning-loop 0.10.0-beta -> 0.11.0-beta pkg updates with associated module changes.

Issue #256, opened 2020-10-23
nix-shell wipes secrets


nixbitcoin:
Running `nix-shell` in the `/examples` directory from master branch results in

```
/nix/store/kn2prc5sz8h9wb07z6h71q6qwbwlsdiw-generate-secrets: 6: /nix/store/nhvpypvj7bq30d5qn4390gfb1mq6xg4p-generate-secrets.sh: [[: not found
/nix/store/kn2prc5sz8h9wb07z6h71q6qwbwlsdiw-generate-secrets: 6: /nix/store/nhvpypvj7bq30d5qn4390gfb1mq6xg4p-generate-secrets.sh: [[: not found
/nix/store/kn2prc5sz8h9wb07z6h71q6qwbwlsdiw-generate-secrets: 6: /nix/store/nhvpypvj7bq30d5qn4390gfb1mq6xg4p-generate-secrets.sh: [[: not found
/nix/store/kn2prc5sz8h9wb07z6h71q6qwbwlsdiw-generate-secrets: 6: /nix/store/nhvpypvj7bq30d5qn4390gfb1mq6xg4p-generate-secrets.sh: [[: not found
/nix/store/kn2prc5sz8h9wb07z6h71q6qwbwlsdiw-generate-secrets: 6: /nix/store/nhvpypvj7bq30d5qn4390gfb1mq6xg4p-generate-secrets.sh: [[: not found
/nix/store/kn2prc5sz8h9wb07z6h71q6qwbwlsdiw-generate-secrets: 6: /nix/store/nhvpypvj7bq30d5qn4390gfb1mq6xg4p-generate-secrets.sh: [[: not found
/nix/store/kn2prc5sz8h9wb07z6h71q6qwbwlsdiw-generate-secrets: 6: /nix/store/nhvpypvj7bq30d5qn4390gfb1mq6xg4p-generate-secrets.sh: [[: not found
/nix/store/kn2prc5sz8h9wb07z6h71q6qwbwlsdiw-generate-secrets: 6: /nix/store/nhvpypvj7bq30d5qn4390gfb1mq6xg4p-generate-secrets.sh: [[: not found
/nix/store/kn2prc5sz8h9wb07z6h71q6qwbwlsdiw-generate-secrets: 6: /nix/store/nhvpypvj7bq30d5qn4390gfb1mq6xg4p-generate-secrets.sh: [[: not found
[...]

PR #261, opened 2020-11-03
Fix netns restart

erikarvstedt:
This fixes netns service restarts on Travis, which were [broken in the test](https://github.com/fort-nix/nix-bitcoin/pull/258#issuecomment-721039403) for #256.
Because I didn't succeed in locally reproducing the error and because of Travis' long queue times, I couldn't determine the true cause of the error with certainty.
But empirically this patch seems to work: Here's the successful [CI test](https://travis-ci.org/github/erikarvstedt/nix-bitcoin/builds/741241894) with #256 applied on top of this PR.
We should merge this PR before #256.

### DIFF
--- a/modules/netns-isolation.nix
+++ b/modules/netns-isolation.nix
@@ -152,7 +152,6 @@ in {
           requiredBy = bindsTo;
           before = bindsTo;
           script = ''
-            ${ip} netns delete ${netnsName} 2> /dev/null || true
             ${ip} netns add ${netnsName}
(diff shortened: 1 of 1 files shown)


jonasnick (approved):
utACK 0972af55f16815c61626359be71cacf650d6400d

jonasnick:
Fwiw I guess you mean #258 and not #256

erikarvstedt:
Yes, sorry, I meant #258.

PR #264, opened 2020-11-09
Fix containers for nixpkgs 20.09

erikarvstedt:
Nixpkgs 20.09 introduced a bug where nixpkgs overlays are ignored in containers. This breaks our `pkgs.nix-bitcoin` overlay:
```bash
$ test/run-tests.sh container
error: attribute 'nix-bitcoin' missing, at /home/main/d/nix-bitcoin/src/modules/bitcoind.nix:62:19
```
[My upstream bugfix PR](https://github.com/NixOS/nixpkgs/pull/98655) has not been merged yet.

Work around this by providing nix-bitcoin pkgs via an option instead of an overlay (in `modules.nix`).
This also has the advantage to slightly speed-up container builds because the `pkgs` of the host system can be reused in the container.

### DIFF
--- a/modules/bitcoind.nix
+++ b/modules/bitcoind.nix
@@ -59,7 +59,7 @@ in {
       enable = mkEnableOption "Bitcoin daemon";
       package = mkOption {
         type = types.package;
-        default = pkgs.nix-bitcoin.bitcoind;
+        default = config.nix-bitcoin.pkgs.bitcoind;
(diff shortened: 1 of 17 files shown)


jonasnick (approved):
Uh that's an annoying bug. Thanks for fixing

ACK f1681f5b45ebe2807503c2d8b00df701b4549ea8

PR #265, opened 2020-11-11
testing framework: re-add features removed in nixpkgs 20.09

erikarvstedt:
##### Copy of commit message
Since nixpks 20.09, the test output is just an empty directory.
Restore saving the log output and linking to the driver.

Without linking to the driver, the driver is eligible for garbage collection after running a test via `run-tests.sh --out-link-prefix`, which implies lengthy driver rebuilds.

### DIFF
--- a/test/lib/make-test-vm.nix
+++ b/test/lib/make-test-vm.nix
@@ -19,7 +19,19 @@ let
            '';
          };
        };
+       test' = test (args // { pkgs = pkgsFixed; });
     in
-      test (args // { pkgs = pkgsFixed; });
+      # See nixpkgs/nixos/lib/testing-python.nix for the original definition
+      test'.overrideAttrs (_: {
+        # 1. Save test output
+        # 2. Add link to driver so that a gcroot to a test prevents the driver from
+        #    being garbage-collected
+        buildCommand = ''
+          mkdir $out
+          LOGFILE=$out/output.xml tests='exec(os.environ["testScript"])' ${test'.driver}/bin/nixos-test-driver
+          ln -s ${test'.driver} $out/driver
+        '';
+      }) // { inherit (test') nodes driver; } ;
+
 in
   fixedTest


jonasnick:
How can I reproduce the driver rebuild? Running `run-tests.sh --out-link-prefix` on master doesn't rerun the tests.

erikarvstedt:
The following demonstrates the gcroots issue on a fresh system with only nix installed.
```bash
git clone https://github.com/fort-nix/nix-bitcoin
run-tests.sh --out-link-prefix /tmp
nix-store --gc
# Rebuilds the whole test, including slow-to-build pkgs like `elementsd`
run-tests.sh --out-link-prefix /tmp
```
If this doesn't trigger large rebuilds on your system, maybe you're using an extra binary cache or there are other gcroots pointing to (parts of) the test driver.

jonasnick:
Hm I can't reproduce it on my dev machine. I need to use the prefix `/tmp/` instead of `/tmp` because my user isn't allowed to create directories in `/` but that should be irrelevant.

erikarvstedt:
Sorry, I've fixed the prefix.
Can you print the output of the last two commands? 

To see that the test derivation has no dependencies, run: 
```
nix-store -q --referrers $(test/run-tests.sh -s default)
# No output
```
So setting a gcroot to the test won't prevent any dependencies (like the driver, pre 20.09) from being garbage-collected.

erikarvstedt:
Sorry, to show that the test driver is missing, we have of course to re-run the test.
Please retry my fixed example.

jonasnick (approved):
ACK 7e3d2965a5ff0cce057bfca2e7f88d04a5f72192

jonasnick:
Apparently I had something left that pointed to the test driver but I can reproduce it now, i.e., the tests are re-run, but at least the packages are not built again which is what happens on master.

erikarvstedt:
Thanks!

The [latest Travis failure](https://travis-ci.org/github/fort-nix/nix-bitcoin/jobs/743662209) is strange:
1. The test driver doesn't notice that bitcoind has started and hangs with `machine: waiting for unit: bitcoind`
2. electrs crashes with 
```
machine # [   58.477706] electrs[1468]: thread 'main' panicked at 'supplied instant is later than self', src/libstd/time.rs:263:9
```
I've never seen any of these happen in local tests. Maybe it's related to the latest nixpkgs updates. It's most certainly not caused by this PR. I'll keep an eye on it.

Issue #260, opened 2020-11-03
CI discussion


jonasnick:
From #258 

> To clarify this, I've added a debug helper and re-run the travis build, but it's still stalling, like in the last few days.
> @jonasnick, should I investigate a migration to Github Actions?

I'd be open to switching to a different CI system. We need to switch from travis.org to travis.com anyway because apparently they're moving over resources to travis.com. But travis.com [has unacceptable permissions](https://github.com/bitcoin/bitcoin/issues/17802#issuecomment-718857563). Apparently [github actions](https://github.com/bitcoin/bitcoin/pull/18031) had similar issues - not sure if that's still the case or what they're specifically referring to. @erikarvstedt Do you have experience with Github Actions? I haven't read up on this really, but short google search revealed that generally it seems to be preferred over travis.

jonasnick:
From https://github.com/bitcoin/bitcoin/issues/17802#issuecomment-724684116
> Travis for open source is effectively shutting down, so I don't think this is an option. See also: [...]

More on Github Actions https://github.com/bitcoin/bitcoin/issues/17803#issuecomment-725551444

nixbitcoin:
We're [now using](https://github.com/fort-nix/nix-bitcoin/pull/276) CirrusCI

PR #272, opened 2020-11-29
WIP: Add github action for CI

jonasnick:
Addresses #260

I tried translating our `travis.yml` to a github actions workflow. On the positive side it's way simpler than messing with a third party service and jobs start immediately. However, the nixos test framework fails with `No KVM available on VM Host.`. Apparently [nested virtualization is disabled on the runners](https://github.com/marketplace/actions/install-nix#is-it-possible-to-run-nixos-tests--qemukvm-with-github-actions) and there are [no plans to fix this soon](https://github.com/actions/virtual-environments/issues/183#issuecomment-706244929) (as of 2020-10-09). I wonder if it's possible to run the nixos tests without KVM or if that would be way too slow. Alternatively we can self-host the runner (which makes things rather complex again) or use something like cirrus-ci.org.

Another TODO is adding the `CACHIX_SIGNING_KEY`.

See the result of the action here: https://github.com/jonasnick/nix-bitcoin/actions/runs/389842729

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -1,0 +1,72 @@
+name: CI
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  setup:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - testmodules: 1
+            stable: 1
+            scenario: "default"
+          - pkg: "hwi"
+            stable: 1
+          - pkg: "hwi"
+            stable: 0
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v2
+      - name: Build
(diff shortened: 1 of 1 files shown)


erikarvstedt:
`KVM` can be disabled with `qemu` option `-accel tcg` (formerly `-no-kvm`).
The runtime increases by factor ~2.5 on `x86-64`, which is acceptable.
Unfortunately, this triggers a kernel page fault panic when `dotnet` is run ([bugtracker](https://bugs.launchpad.net/qemu/+bug/1866892)), so we can't test `btcpayserver` this way.

erikarvstedt:
I'm looking into setting up Cirrus now.

erikarvstedt:
Superseded by #276.

PR #275, opened 2020-12-01
electrs: 0.8.5 -> 0.8.6

### LINKED ISSUE #274: Update electrs version
I've tried to check if you've updated `electrs` version to 0.8.6 and it seems you didn't, close & ignore if you did.

The recent version of `electrs` is a quite important update as it fixes communication with Electrum (it was problematic before and completely broken for Electrum 4.0.5) and potential security vulnerability.
It also brings few other fixes and improvements of reliability etc.
We recommend you update it as soon as possible.

The new version is fully backwards-compatible. You only need to replace the binary without changing anything else. However as you're packaging system, I suggest you look at possible improvements available thanks to the recent changes:

* Dynamic linking is now supported and documented - it is also deterministic
* Cross compilation is supported and documented, however it was only tested for arm64 so far, armhf failed

The documentation can be found in see `docs/usage.md`.

I've noticed you had some trouble before (#143 ), I guess these changes will help you avoid those issues. Feel free to ask me if you need help with something Rust/build/electrs related.

erikarvstedt:
Fixes #274.

### DIFF
--- a/pkgs/electrs/default.nix
+++ b/pkgs/electrs/default.nix
@@ -1,24 +1,19 @@
 { lib, rustPlatform, llvmPackages, fetchurl, pkgs }:
 rustPlatform.buildRustPackage rec {
   pname = "electrs";
-  version = "0.8.5";
+  version = "0.8.6";
 
   src = fetchurl {
     url = "https://github.com/romanz/electrs/archive/v${version}.tar.gz";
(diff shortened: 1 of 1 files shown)


jonasnick (approved):
ACK a6346c25618ffbfb6b4fc4ba3a20d7b13626e0e7

PR #276, opened 2020-12-02
Use Cirrus CI

erikarvstedt:
Cirrus CI has explicit KVM support and is also used by https://github.com/bitcoin/bitcoin.

See it in action [at my fork](https://cirrus-ci.com/github/erikarvstedt/nix-bitcoin).

Before merging, we need to enable Cirrus for this repo and update the `CACHIX_SIGNING_KEY`:
- Visit https://github.com/marketplace/cirrus-ci and enable the Public Plan.
- The key is stored in `~/.config/cachix/cachix.dhall` as `secretKey`.
  Encrypt the key via https://cirrus-ci.com/github/fort-nix/nix-bitcoin -> Settings -> Secured Variables
  and post it here.


.cirrus.yml
```diff
@@ -0,0 +1,43 @@
+container:
+  image: nixos/nix
```
jonasnick, line 2: Which container registry does this use?

ci/build.sh
```diff
@@ -0,0 +1,74 @@
...
+printf "instantiated "; realpath $tmpDir/drv
+
+outPath=$(nix-store --query $tmpDir/drv)
+if nix path-info --store https://$cachixCache.cachix.org $outPath &>/dev/null; then
+    echo "$outPath" has already been built successfully.
+    exit 0
+fi
+
+# Cirrus doesn't expose secrets to pull-request builds,
+# so skip cache uploading in this case
```
jonasnick, line 62: Perhaps worth adding a comment why we're watching the store here if we're going to push later anyway.

jonasnick (commented):
What I've seen so far of cirrus CI seems nice: good documentation, fine-grained github app permissions etc.

> Make more economic use of the free CI resources

Generally, I'd be fine with spending a few bucks per month if that would be worth it.

jonasnick:
I added the cirrus app to the fort-nix organization. https://cirrus-ci.com/github/fort-nix/nix-bitcoin still shows "Repository not found!". Also, I can't login to cirrus, because github wants to give cirrus the permission to "act on my behalf" (wtf).
cirrus

erikarvstedt:
The permissions preview is indeed poorly designed. Here's how the actual permissions look like:

The `write access` part is the same as in Travis and other CI apps.

.cirrus.yml
erikarvstedt: https://hub.docker.com/r/nixos/nix/

jonasnick:
@erikarvstedt What your screenshot shows are the app permissions and I can see the same already in the fort-nix organization settings. Sorry, not sure how they relate to the "Sign in" permissions required for logging in to cirrus-ci.com.

However, according to the two github employees active [in this github thread](https://github.community/t/why-does-this-forum-need-permission-to-act-on-my-behalf/120453) this is "just" poor wording and essentially means that

> it has the ability to act on your behalf or know which resources you can access, but only within the scope of the permissions we’ve requested, in other words verifying your identity

Moreover, I don't think the github forums would require the "act on your behalf" permissions if it would really mean "act on your behalf". It's quite disappointing that the wording still hasn't been improved.

I'll try this again later today.

.cirrus.yml
jonasnick: Oh nice, that's quite a bit better than manually downloading nix as before.

jonasnick:
The `CACHIX_SIGNING_KEY` is:

```
ENCRYPTED[!cec502ed813cbcd0237697d2031f750186ff20eed5815b1ad950ad2f2d701702ae6ba2f0cb4cb1985687a696c8ee492c!]
```

jonasnick (approved):
ACK 1c0233c0a8d35f73b1f2ba2ebd98f33590a0fd15

PR #280, opened 2020-12-07
lnd: fix rpc/rest port configuration

mmilata:
Setting `lnd.restPort` currently causes lnd not to start because it does not get written into the config. Probably introduced in de23fdd37780406c9b09e9f0528d8f7ae7cf1a76. Did not test with netns-isolation.


erikarvstedt (approved):
Yes, I accidentally introduced this in de23fdd37780406c9b09e9f0528d8f7ae7cf1a76.
Thanks for the fix!

ACK 5feb9e3eaea77d8db159d1d084a22434d3177841

PR #281, opened 2020-12-09
lightning-loop: 0.11.1-beta -> 0.11.2-beta

nixbitcoin:
Simple update of lightning-loop allowing you to set custom swap size limits.


erikarvstedt (approved):
ACK 522b0000e6b4905d019d451f2651f7310ce8f690

PR #279, opened 2020-12-06
Update nixpkgs

jonasnick:
Includes clightning 0.9.1 -> 0.9.2.

PR #282, opened 2020-12-10
Improve tests

erikarvstedt:
Commits starting with `ci: extract build-to-cachix.sh` implement running the `pkgsUnstable` test locally.
This guarantees feature parity between local and CI tests and helps avoiding CI-only failures like in https://github.com/fort-nix/nix-bitcoin/pull/279.


jonasnick (approved):
Very nice! ACK 9977fa69afeb04e8c14fc8217bff5ae2699f32ae

PR #283, opened 2020-12-14
joinmarket: 0.7.4 -> 0.8.0-bcfa7eb

nixbitcoin:
This PR updates JoinMarket to 0.8.0 plus a hotfix for the genwallet script that I wrote myself. It includes a warning about the necessary migration steps. This warning isn't something for config versioning, right?

I had to temporarily change `get-sha256.sh` for the pinned commit. We can revert all this with version 0.8.1.

### DIFF
--- a/modules/versioning.nix
+++ b/modules/versioning.nix
@@ -6,14 +6,54 @@ let
 
   # Sorted by increasing version numbers
   changes = [
-    # None yet
-    # {
-    #   version = "0.1";
-    #   condition = config.services.foo.enabled;
-    #   message = ''
-    #     demo message
-    #   '';
-    # }
+    {
+      version = "0.0.26";
+      condition = config.services.joinmarket.enable;
+      message = let
+        inherit (config.services.joinmarket) dataDir;
+      in ''
+        JoinMarket 0.8.0 moves from wrapped segwit wallets to native segwit wallets.
+
+        If you have an existing wrapped segwit wallet, you have to manually migrate
+        your funds to a new native segwit wallet.
+
+        To migrate, you first have to deploy the new JoinMarket version:
+        1. Set `nix-bitcoin.configVersion = "0.0.26";` in your configuration.nix
+        2. Deploy the new configuration
+
+        Then run the following on your nix-bitcoin node:
+        1. Move your wallet:
+           mv ${dataDir}/wallets/wallet.jmdat ${dataDir}/wallets/old.jmdat
+        2. Autogenerate a new p2wpkh wallet:
+           systemctl restart joinmarket
+        3. Transfer your funds manually by doing sweeps for each mixdepth:
+           jm-sendpayment -m <mixdepth> -N 0 old.jmdat 0 <destaddr>
+
+           Run this command for every available mixdepth (`-m 0`, `-m 1`, ...).
+           IMPORTANT: Use a different <destaddr> for every run.
+
+           Explanation of the options:
+           -m <mixdepth>: spend from given mixdepth.
+           -N 0: don't coinjoin on this spend
+           old.jmdat: spend from old wallet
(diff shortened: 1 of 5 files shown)


erikarvstedt:
This is a perfect use case for config versioning, because this warning is only relevant for users of the previous version of the joinmarket module.

Here's a draft [here](https://github.com/erikarvstedt/nix-bitcoin/commits/joinmarket-0.8.0-ea).

Could you add a bit more detail about step `jm-wallet-tool bak.jmdat`?
Or is this really all that's necessary to transfer the funds to the new wallet?

nixbitcoin:
I squashed your commit and included thorough migrations steps.

I also had to point out that in this case, we first have to change the `configVersion` and deploy, then perform the migration steps. Any smarter way to do this?

erikarvstedt:
> Any smarter way to do this?

No, it's fine this way.

[Fixups](https://github.com/erikarvstedt/nix-bitcoin/commits/joinmarket-0.8.0-ea)

nixbitcoin:
Doing `cd ${dataDir}/wallets` is actually not necessary because we do `cd '${cfg.dataDir}' && ${cfg.cliExec} sudo -u ${cfg.user} $jm/$bin --datadir='${cfg.dataDir}` on every command.

Also I think you meant to say `0: set amount to zero to do a sweep, i.e. transfer all funds at given mixdepth` not `0: set amount to zero to do a sweep, i.e. transfer all funds at given`

modules/versioning.nix
```diff
@@ -6,14 +6,43 @@ let
...
+
+           Explanation of the options:
+           -m <mixdepth>: spend from given mixdepth.
+                          Repeat the command for every available mixdepth, e.g. `-m 0`, `-m 1`, ...
+           -N 0: don't coinjoin on this spend
+           old.jmdat: spend from old wallet
+           0: set amount to zero to do a sweep, i.e. transfer all funds at given mixdepth
+           <destaddr>: destination p2wpkh address from wallet.jmdat with mixdepth 0
+
+           You might want to time stagger these 4 transactions.
```
erikarvstedt, line 40:
```suggestion
           You might want to time stagger these transactions.
```
The number of max. mixdepths is configurable and the default is 5 (= 5 transactions).

modules/versioning.nix
```diff
@@ -6,14 +6,43 @@ let
...
+           mv ${dataDir}/wallets/wallet.jmdat ${dataDir}/wallets/old.jmdat
+        2. Autogenerate a new p2wpkh wallet:
+           systemctl restart joinmarket
+        3. Transfer your funds manually by doing sweeps for each mixdepth:
+           jm-sendpayment -m <mixdepth> -N 0 old.jmdat 0 <destaddr>
+
+           Explanation of the options:
+           -m <mixdepth>: spend from given mixdepth.
+                          Repeat the command for every available mixdepth, e.g. `-m 0`, `-m 1`, ...
+           -N 0: don't coinjoin on this spend
```
erikarvstedt, line 35: Doesn't this decrease privacy because all input transactions can be linked together?

erikarvstedt (changes requested):
We should also note that different destination addresses should be used for each sweep.

nixbitcoin:
> We should also note that different destination addresses should be used for each sweep.

Done. Also included possibility of coin-freezing to use individual coins and not connect them

erikarvstedt:
[Final fixups](https://github.com/erikarvstedt/nix-bitcoin/commits/joinmarket-0.8.0-ea).

nixbitcoin:
> [Final fixups](https://github.com/erikarvstedt/nix-bitcoin/commits/joinmarket-0.8.0-ea).

Squashed

erikarvstedt (approved):
ACK fdfafb2f40076ed053ccac288c2d33ffb117e313

jonasnick:
Man, that config versioning stuff really is super cool.

PR #286, opened 2020-12-23
Fix boot loader reference for UEFI

galderz:


erikarvstedt (approved):
ACK 792962bb32b085d4d88082c698c58985e2da2008

PR #287, opened 2020-12-23
Removed reference to file that does not exist

galderz:


jonasnick (commented):
The file should exist after being created in step 4.1.

erikarvstedt:
The file is copied from `examples/nixops/node-vbox.nix` in step 3.

erikarvstedt:
Retracted my incorrect comment. @jonasnick is right.
The manual would be clearer if the file were renamed to `node-target.nix`.

erikarvstedt:
Superseded by #296.

galderz:
I don't know what exactly happened, but when I copied over files, i ended up a `nixops/node.nix` that contained:

```
  bitcoin-node = { config, pkgs, lib, ... }: {
    imports = [
      ../configuration.nix
      <nix-bitcoin/modules/deployment/nixops.nix>
    ];
```

So, the actual configuration is in a `configuration.nix` file at the root of the node folder.

PR #290, opened 2020-12-30
JoinMarket Orderbook Watcher

nixbitcoin:
This PR adds the obwatcher script to the joinmarket package and adds a stand-alone joinmarket-obwatcher module.

I decided to make this module independent from the joinmarket module following the principle of least principle. It does not need to interact in anyway with the existing bitcoind or joinmarket services.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ NixOS modules
   * [liquid](https://github.com/elementsproject/elements)
   * [Lightning Loop](https://github.com/lightninglabs/loop)
   * [JoinMarket](https://github.com/joinmarket-org/joinmarket-clientserver)
+    * [JoinMarket Orderbook Watcher](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/docs/orderbook.md)
   * [recurring-donations](modules/recurring-donations.nix): for periodic lightning payments
   * [bitcoin-core-hwi](https://github.com/bitcoin-core/HWI)
 * Helper
--- a/examples/configuration.nix
+++ b/examples/configuration.nix
@@ -179,6 +179,8 @@
   # earn sats by providing CoinJoin liquidity. This makes it impossible to use other
   # scripts that access your wallet.
   # services.joinmarket.yieldgenerator.enable = true;
+  # Enable this option to enable the JoinMarket order book watcher.
+  # services.joinmarket-ob-watcher.enable = true;
 
   # FIXME: Define your hostname.
   networking.hostName = "nix-bitcoin";
--- a/modules/joinmarket-ob-watcher.nix
+++ b/modules/joinmarket-ob-watcher.nix
@@ -1,0 +1,99 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+let
+  cfg = config.services.joinmarket-ob-watcher;
+  inherit (config) nix-bitcoin-services;
+  nbPkgs = config.nix-bitcoin.pkgs;
+  torAddress = builtins.head (builtins.split ":" config.services.tor.client.socksListenAddress);
+  configFile = builtins.toFile "config" ''
+    [BLOCKCHAIN]
+    blockchain_source = no-blockchain
+
+    [MESSAGING:server1]
+    host = darksci3bfoka7tw.onion
+    channel = joinmarket-pit
+    port = 6697
+    usessl = true
+    socks5 = true
+    socks5_host = ${torAddress}
+    socks5_port = 9050
+
+    [MESSAGING:server2]
+    host = ncwkrwxpq2ikcngxq3dy2xctuheniggtqeibvgofixpzvrwpa77tozqd.onion
+    channel = joinmarket-pit
+    port = 6667
+    usessl = false
+    socks5 = true
+    socks5_host = ${torAddress}
+    socks5_port = 9050
+  '';
+in {
+  options.services.joinmarket-ob-watcher = {
+    enable = mkEnableOption "JoinMarket orderbook watcher";
+    address = mkOption {
+      type = types.str;
+      default = "127.0.0.1";
+      description = "HTTP server address.";
+    };
+    port = mkOption {
+      type = types.port;
+      default = 62601;
+      description = "HTTP server port.";
+    };
+    dataDir = mkOption {
+      readOnly = true;
+      default = "/var/lib/joinmarket-ob-watcher";
+      description = "The data directory for JoinMarket orderbook watcher.";
+    };
+    user = mkOption {
+      type = types.str;
+      default = "joinmarket-ob-watcher";
+      description = "The user as which to run JoinMarket orderbook watcher.";
+    };
+    group = mkOption {
+      type = types.str;
+      default = cfg.user;
+      description = "The group as which to run JoinMarket orderbook watcher.";
+    };
+    # This option is only used by netns-isolation
+    enforceTor = mkOption {
+      readOnly = true;
+      default = true;
+    };
+  };
+
+  config = mkIf cfg.enable {
+    services.tor = {
+      enable = true;
+      client.enable = true;
+    };
+
+    systemd.services.joinmarket-ob-watcher = {
+      wantedBy = [ "multi-user.target" ];
+      requires = [ "tor.service" ];
+      after = [ "tor.service" ];
+      preStart = ''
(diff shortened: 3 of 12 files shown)


erikarvstedt:
Could you rename the module to `joinmarket-ob-watcher`, so that we're consistent with the upstream name?

erikarvstedt:
By the way, you can almost always include changes to the tests in the commit that introduces functional changes. This doesn't increase commit complexity because the extra hunks will only affect test-specific files and are easily distinguished from the other changes. The advantage is that the test changes directly document the functional changes.
If you want to split this PR into multiple comits, I'd recommend to separate the pkg and the module. But a single commit is totally fine, foo.

erikarvstedt:
I'd recommend that we use subdomains instead of URL path prefixes for separating different webservices in nginx.

The corresponding nixos configuration looks like this:
```
services.nginx = {
  enable = true;
  virtualHosts."ob-watcher.*".locations."/" = {
    proxyPass = "http://${joinmarket-ob-watcher.address}:${toString joinmarket-ob-watcher.address.port}";
  };
  virtualHosts."nanopos.*".locations."/" = {
    proxyPass = "http://${nanopos.address}:${toString nanopos.address.port}";
  };
}
```
`ob-watcher` can now be accessed via `ob-watcher.localhost` or `ob-watcher.mynode.org`.
This is also compatible with Lets Encrypt, which supports wildcard certificates that cover all subdomains.

This approach avoids the inconvenience of configuring the URL path prefixes for each webservice, which sometimes isn't even possible.
Also, this allows adding dedicated onion services for each webservice without the need to add the redundant path prefix to the service-specific onion address. E.g., one can use `8a...c4.onion` instead of `8a...c4.onion/ob-watcher`.

In case of `ob-watcher`, this approach also allows us to use the default `orderbook.html`.

erikarvstedt:
Btw., containers are great for developing a service:
```
# scenarios.nix
{ testEnv, config, pkgs, lib }:
with testEnv;
with lib;
rec {
  ob-watcher = {
    services.bitcoind.enable = mkForce false;
    services.joinmarket-obwatcher.enable = true;
    test.container.exposeLocalhost = true;
  };
  # ...
}
```
```
scenarioOverridesFile=scenarios.nix run-tests.sh -s ob-watcher container
## In container shell
c curl -L localhost/obwatcher
# Access externally from host
curl -L $ip/obwatcher
sudo -u $(logname) firefox $ip/obwatcher
```

nixbitcoin:
> Also, this allows adding dedicated onion services for each webservice without the need to add the redundant path prefix to the service-specific onion address. E.g., one can use `8a...c4.onion` instead of `8a...c4.onion/ob-watcher`.

Could you still browse to ob-watcher from the nginx onion service? It's nice not calling up a new onion service for every link on a website.

erikarvstedt:
Sure, but nix-bitcoin shouldn't force a particular nginx configuration and path-prefix for using a webservice. It should be possible just host an ob-watcher with a node, without a path prefix.

erikarvstedt:
If you think that's worthwhile, you could add an option `urlPathPrefix` to `ob-watcher`, like `rootpath` for btcpayserver. But using a subdomain is much simpler.

erikarvstedt:
It just occured to me that there's a simple solution:
When converting the links from root relative to relative (`href="/ordersize"` -> `href="ordersize"`) `ob-watcher` works with arbitrary URL path prefixes.

```
		<nav class="navbar navbar-inverse navbar-fixed-top">
	      <div class="container">
	        <div class="navbar-header">
			<a class="navbar-brand" href=".">Join Market</a>
	        </div>
	        <div id="navbar" class="collapse navbar-collapse">
	          <ul class="nav navbar-nav">
	            <li class="active"><a href=".">Orders</a></li>
		    <li><a href="ordersize">Size Distribution</a></li>
	            <li><a href="depth">Depth</a></li>
	            <li><a href="orderbook.json">Export orders</a></li>
	            <li><a target="_blank" href="https://github.com/JoinMarket-Org/joinmarket-clientserver/releases">New segwit version</a></li>
	          </ul>
	        </div><!--/.nav-collapse -->
	      </div>
	    </nav>
```

nixbitcoin:
0dfbded
- combines all commits into one
- converts the links from root relative to relative
- renames the module
- adds a standalone Tor onion service (also to `nodeinfo`).

erikarvstedt:
[Fixups and one extra commit](https://github.com/erikarvstedt/nix-bitcoin/commits/joinmarket-obwatcher-ea).

erikarvstedt:
I've added an [upstream PR](https://github.com/JoinMarket-Org/joinmarket-clientserver/pull/771) for the orderbook patch.

nixbitcoin:
Squashed your fixups except, ce90becbf228a825ffc34db2da80700b751a7c0a because `netns-isolation` is using `enforceTor` to make outbound firewall rules for the `nb-joinmarket-ob-watcher` network namespace. Also had to correct a mistake in bd648a16430617116db8b524d3e43321bc241320 on the IRC server `port`.

nixbitcoin:
"JoinMarket Orderbook Watcher" added to `README.md` features.

erikarvstedt:
Thanks for spotting the mistakes. [One final fixup](https://github.com/erikarvstedt/nix-bitcoin/commits/joinmarket-obwatcher-ea-2).

Kixunil:
> I'd recommend that we use subdomains instead of URL path prefixes for separating different webservices in nginx.

This has consequences that led me to not do that in my Debian repository, so perhaps you want to know about it. The issue is that if the user wants to use clearnet and Letsencrypt, then either the subdomains need to be in the certificate, which leaks the list of web apps or wildcard certificate needs to be used which is annoying to work with because it requires manual update of TXT DNS record.

erikarvstedt:
@Kixunil, thanks for bringing up these caveats.
For automatically generating wildcard certs, NixOS's [security.acme](https://search.nixos.org/options?channel=unstable&from=0&size=30&sort=relevance&query=security.acme) module supports (via go-acme) [a wide range of common DNS providers](https://go-acme.github.io/lego/dns/). But still, the weakened security of keeping DNS API credentials on the node remains an issue.

Kixunil:
Yeah, that's also a possibility. I may be biased because I don't use any of those common providers. :)

jonasnick (commented):
Fwiw, generally I don't think this PR is a great fit for nix-bitcoin as there is very little benefit to having this data provided by more than single person. People who run proper joinmarket have access to the real orderbook and for everyone else the orderbook is just informational. Since this PR has already received some review I'm open to merging it nonetheless if it's ACK'd and doesn't get too complex.

nixbitcoin:
> Fwiw, generally I don't think this PR is a great fit for nix-bitcoin as there is very little benefit to having this data provided by more than single person. People who run proper joinmarket have access to the real orderbook and for everyone else the orderbook is just informational. Since this PR has already received some review I'm open to merging it nonetheless if it's ACK'd and doesn't get too complex.

I mainly wanted to upstream this feature and get review after implementing it for [nixbitcoin.org](https://nix-bitcoin.org). I don't think it causes too much added complexity since we rely on the same JoinMarket package and the orderbook watcher script is relatively stable. I think we can go ahead and merge.

erikarvstedt:
ACK 6edbeb317dc709a3fbbe672222ae45e7519c84e0

erikarvstedt (dismissed):
ACK 6edbeb317dc709a3fbbe672222ae45e7519c84e0

nixbitcoin:
Rebased with necessary patches

erikarvstedt:
[Fixed branch](https://github.com/erikarvstedt/nix-bitcoin/commits/joinmarket-obwatcher-ea3):
- Update to the current joinmarket master to include my [upstream ob-watcher fix](https://github.com/JoinMarket-Org/joinmarket-clientserver/commit/00be3a3748bc718faf8697376284177732503f81).
  Besides that, there are no significant changes in this version ([history](https://github.com/JoinMarket-Org/joinmarket-clientserver/commits/master)).
- Restore enabling tor, which is a strict requirement for ob-watcher.
  Please keep in mind that nix-bitcoin modules should work without the secure-node preset.

erikarvstedt (approved):
ACK 8c125ec48c3f3caabab65acd2d7c9f2dffe1a2d3

PR #291, opened 2020-12-30
btcpayserver: add rootpath option

nixbitcoin:
This PR adds the BTCPayServer rootpath option to our btcpayserver module. This allows you to set a custom root path like with `https://nixbitcoin.org/btcpayserver/`

### DIFF
--- a/modules/btcpayserver.nix
+++ b/modules/btcpayserver.nix
@@ -84,6 +84,12 @@ in {
         default = null;
         description = "The lightning node implementation to use.";
       };
+      rootpath = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        example = "btcpayserver";
+        description = "The prefix for root-relative btcpayserver URLs.";
+      };
       enforceTor = nix-bitcoin-services.enforceTor;
     };
   };
@@ -150,6 +156,7 @@ in {
         btcexplorerurl=http://${cfg.nbxplorer.bind}:${toString cfg.nbxplorer.port}/
         btcexplorercookiefile=${cfg.nbxplorer.dataDir}/${config.services.bitcoind.makeNetworkName "Main" "RegTest"}/.cookie
         bind=${cfg.btcpayserver.bind}
+        ${optionalString (cfg.btcpayserver.rootpath != null) "rootpath=${cfg.btcpayserver.rootpath}"}
         port=${toString cfg.btcpayserver.port}
       '' + optionalString (cfg.btcpayserver.lightningBackend == "clightning") ''
         btclightning=type=clightning;server=unix:///${cfg.clightning.dataDir}/bitcoin/lightning-rpc


modules/btcpayserver.nix
```diff
@@ -84,6 +84,11 @@ in {
         default = null;
         description = "The lightning node implementation to use.";
       };
+      rootpath = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        description = "The root path in the URL to access BTCPay.";
```
erikarvstedt, line 90:
```suggestion
        example = "https://mynode.org/btcpayserver/"
        description = "The root path prefix in the btcpayserver URL.";
```

nixbitcoin: It's actually just the `/btcpayserver` addendum to the URL. 

Fixed in 93f745382912c8b37d5a2649f8ee9e085d4b1cdd

modules/btcpayserver.nix
```diff
@@ -84,6 +84,12 @@ in {
         default = null;
         description = "The lightning node implementation to use.";
       };
+      rootpath = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        example = "/btcpayserver";
+        description = "The root path in the URL to access BTCPay.";
```
erikarvstedt, line 91:
```suggestion
        description = "The root path in the URL to access btcpayserver.";
```
We're using `btcpayserver` instead of `BTCPay` in the other descriptions.

erikarvstedt:
What's the default value in btcp for the root path?
Is the leading slash really needed?

modules/btcpayserver.nix
nixbitcoin: Fixed in a0446c8c11608b620391e819fb19f703a1ee7b13

nixbitcoin:
> What's the default value in btcp for the root path? Just an empty string?

The default value is `/`.

> Is the leading slash really needed?

I verified that the leading slash is in fact not needed.

erikarvstedt (approved):
ACK a0446c8c11608b620391e819fb19f703a1ee7b13

modules/btcpayserver.nix
```diff
@@ -84,6 +84,12 @@ in {
         default = null;
         description = "The lightning node implementation to use.";
       };
+      rootpath = mkOption {
+        type = types.nullOr types.str;
+        default = null;
+        example = "btcpayserver";
+        description = "The root path in the URL to access btcpayserver.";
```
erikarvstedt, line 91:
```suggestion
        description = "The prefix for root-relative btcpayserver URLs.";
```
Sorry for bringing this up again, but now we have a really clear description.

erikarvstedt (changes requested):

modules/btcpayserver.nix
nixbitcoin: No problem, I care about having accurate documentation.

Fixed in edc657d138e51970e0ea981fbbc2d3471d0cda41

erikarvstedt:
ACK edc657d138e51970e0ea981fbbc2d3471d0cda41

erikarvstedt (approved):
ACK edc657d138e51970e0ea981fbbc2d3471d0cda41

Issue #194, opened 2020-06-03
Example module


nixbitcoin:
**Time Horizon:** Mid Term

**Description:**  Provide example module, whose structure should be followed in all modules.

**Motivation:** This should streamline the development and review process.

**Implementation:** Create example module in `contributing/` and modify all existing and future modules to adhere to it.

nixbitcoin:
Modules were streamlined in https://github.com/fort-nix/nix-bitcoin/pull/312 and now all serve as "examples"

PR #292, opened 2020-12-30
joinmarket: always synchronize secrets.jm-wallet-password

nixbitcoin:
`secrets.jm-wallet-password` is always needed by joinmarket, not just when `joinmarket.yieldgenerator.enable`

### DIFF
--- a/modules/joinmarket.nix
+++ b/modules/joinmarket.nix
@@ -196,11 +196,11 @@ in {
         ReadWritePaths = "${cfg.dataDir}";
       } // nix-bitcoin-services.allowTor;
     };
+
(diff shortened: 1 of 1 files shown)


erikarvstedt:
[Fixup](https://github.com/erikarvstedt/nix-bitcoin/commits/fix-joinmarket-password-ea).
In other modules we're defining secrets at the very end. You can always use `bitcoind.nix` as a reference for service module layout. 
I'm planning to document this (see also https://github.com/fort-nix/nix-bitcoin/issues/194) and streamline the layout across all modules.

nixbitcoin:
Squashed

erikarvstedt (approved):
ACK ed636dd070cfa5f2df26d9bd2301c3aa049b0431

PR #293, opened 2020-12-30
Module refactorings, onionServices

erikarvstedt:
This PR has 4 topics that build on each other:

### 1. Add consistent address options for services
*Implemented in commits `bitcoind: add consistent address options` to `remove nix-bitcoin.pkgs.lib`.*

Rename address options to `address` and `port` (or `rpcAddress` and `rpcPort` for secondary features).
Previously, we used option names as diverse as `listen`, `bind`, `bind-addr` and `host`.

Also, give `port` options an explicit default value.
This improves clarity and ensures that port values are always accessible from within the modules.

### 2. Rename `onion-chef` -> `nix-bitcoin.onionAddresses`, improve implementation
*Implemented in commits `rename onion-chef -> nix-bitcoin.onionAddresses` to `onionAddresses: add readonly option 'dataDir'`.*

These are refactorings without new features.

### 3. Add `nix-bitcoin.onionServices` module
*Implemented in commits `add nix-bitcoin.onionServices` to `lnd: use onionServices for address announcing`.*

Onion services can now be defined in a concise, generic way:
```
nix-bitcoin.onionServices.clightning = {
  enable = true;
  announce = true;
};
```
This is enabled by the consistent address/port naming of step 1 and the optional service option `getAnnounceAddressCmd`.
Both form a generic reverse proxy interface that allows us to easily add other reverse proxies in the future, like (letsencrypt) TLS.

The `enforceTor` and `onionServices` config for services has been extracted to `presets/enable-tor.nix` and can now be used outside of the `secure-node` preset.
This also greatly improves the readability of `secure-node`, which now consists of just a screenful of essential settings.

Onion services now automatically work for services that bind to an INADDR_ANY (`0.0.0.0`) address.

The `operator` user is now automatically granted access to all onion services and doesn't need separate access definitions for each service.

### 4. Move features from the `secure-node` template to the other modules.
*Implemented in commits including and following `remove webindex`.*

This makes these features usable outside of `secure-node`.

To see the new `nodeinfo` output, run `test/run-tests.sh container --run c nodeinfo`.


erikarvstedt:
Rebased.

nixbitcoin:
Concept ACK. I think we should provide some migration instructions for users who have the obsolete configuration settings. I checked if it could be done with config versioning, but evaluation fails earlier.

erikarvstedt:
I've added error messages for obsolete `address`/`port` options.
We'll use this approach more thoroughly when we've reached a stable release.

nixbitcoin:
Can we also add warnings for the obsolete `announce-tor` options? I would say these are even more important than the `address`/`port` options since they are probably defined in existing `configuration.nix`'s.

I tried inserting them in `modules/obsolete-options.nix` but I'm getting
```
error: infinite recursion encountered, at /nix/store/7k1s2pcc3n2zh40pyj79743599n135wc-source/lib/modules.nix:345:19
```

erikarvstedt:
Added errors for `{clightning,lnd}.announce-tor` options.

nixbitcoin (dismissed):
```
Hash: SHA512

ACK 233570b0a3196195f7d5617fceebeb9e1f142554
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCgAdFiEEV3o0Un8+KoXoD+Fk3RH5rVMIs7oFAl/0aPUACgkQ3RH5rVMI
s7qtEBAAweJ5DTWuOxVZ5FX1TaQr0nqQrXT6qBKZ3Yz2MU+AB9JJr0zml8j9zUfh
1RLnL7xJJ4ZWShdcD1kY5pHIcatpKgGhgKmULjiWg8lPjJPk6xOCNd81hGWB1ptL
EhezF938c7nqSni6BsHI+/SM5qmDjlPTBqzJeJeYiiVnhmRyZTihepTtVmWAXNbU
i3dCe049v23K4r5PMY662Cb0knNQRDs8eq3V0PsM3vB/EB+yDqqU0zCGfnhufkx/
5XkP3P+oaXEo3/by4930k3eHbbi0Xo77/myz60NYfIuEXPqIGeKo5kkXcqtEvSHv
uR7DYXirB8kG41/0sSfi7f/QvIeQ4UY0LrMXp4qfVUuxn52T5HYyJsK/I7yxVxi2
CYAe8QLM5aZMUcYQa69HD9lvZMDVJxXqy4YrTPPNRg4bfvWgfKBKwTThAGjchBvz
Te81I00jAK7OaVuu1XLGAI6LW5nywN9GffkFerXq2SpdUU8BbKzmhiEvlEBT0YiT
cFSBSm3OZwrYavMyhqhgjXKF1Y5EAEi21HV45qT4n5DrTpRPnMgUi++E+r6mPABz
lNB6iXlnQHK5YqlFslJwcJILr8EluOA7PcTgVMa6jk9WUybkMF++hmcZgjLBdWMg
6RUEkj1S9ZwQKU9NVWW6U0o511s2xH/QAEa+mE7nW6vGtl/d3Wg=
=bWex
-----END PGP SIGNATURE-----
```

erikarvstedt:
Fixed a typo.

erikarvstedt:
I've squashed in a final patch that sets the `nix-bitcoin.onionServices` config in `enable-tor.nix` to priority `default`.
This allows users to disable specific onionServices without using `mkForce`.

erikarvstedt:
We should merge this PR before `JoinMarket Orderbook Watcher` so we don't have to subsequently fix the `onionServices` setup for `ob-watcher`.

nixbitcoin:
> We should merge this PR before `JoinMarket Orderbook Watcher` so we don't have to subsequently fix the `onionServices` setup for `ob-watcher`.

I agree with this procedure. I will implement the necessary changes in `JoinMarket Orderbook Watcher`

erikarvstedt:
> ... reduces the list of onionServices to only sshd

Can you try to find out what causes this in your config?
It's definitely working in our test configs:
```
read -d '' tmpstr <<'EOF'
  let cfg = ((import <nix-bitcoin/test/tests.nix> { scenario = "default"; }).vm {}).nodes.machine.config; in
  with cfg;
  services.tor.hiddenServices
EOF
nix eval "($tmpstr)"
```

erikarvstedt:
> ... reduces the list of onionServices to only sshd

Fixed. When used with an `attrs` parameter, `mkDefault` doesn't propagate the priority to the attr values.

nixbitcoin (approved):
```
Hash: SHA512

ACK 599851b30c3f8045732061a2d5160537c2fab67a
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCgAdFiEEV3o0Un8+KoXoD+Fk3RH5rVMIs7oFAl/14asACgkQ3RH5rVMI
s7qzzA//Ud3/UBFSxySuNjFEQF5fn3u8beBdp2C48WTj51PylIx65X0qF/K8RcPd
PY2yhHm96eDhYZqgac+aEuqubxr8OAzEvN2ntElqOEMPOzWbHmvpEWMlrfgWV0il
UFLai7XFn/M8AwVYVsNSaHwCfjlKTsB3wiS5i622sabYlRsAQK7mJrdb7m/BytOG
DXIooXJmm1Ub8Z1p7QtpPmQctiuv65xGWn7E6xa3L14UiBX+2M1DOgIgnGpQOOLv
1H7/ongPBIX1fh9SNpHn6lv2PCaAS8sEvkcBoRcbFtCRPR/4llohghZXNyJ8zQNQ
pe72B1j4pcj9W8WSOEtZBzgcqlFj9E8S5eojHdqlrxmFX3JGJSp8ZZbTJmdjKCu2
z5CK2a2hD4618phv3WGLUc22B156TtM47wscljOs6Ycb/dIg0jHW8dlkDrp09FQj
NBTiEP/oNt3iUK4xodOXL+H5D2+WNSNo9bdn51bg4gdztL9PiMLoAJbbSXZFVJQh
hGhY6PsOSqhWx7iPxUrmOoWCQ8RDMBIkmPZMC4QGMP6/F8D80VU/O2W/SVjT8cCy
M04p0do6ciaRf3SVi8pu36yHv35mwVRiHMGHx1ZLAr+gx9T7pz9rVNgYFSY9huf0
DF3FurlttOIXdfAM7nPCVQeWsnxuR0gFIffy4+oAmydUdV8RFB8=
=5ozb
-----END PGP SIGNATURE-----
```

jonasnick:
How are users supposed to figure out which option to set, for example bitcoind's `bind`?

erikarvstedt:
That's what this PR is trying to improve. The `listen/bind/address/host` option is always named `address` and it's defined right after the service's `enable` option.

jonasnick:
Users first need become aware of this. Many are probably already know some of the options that are renamed in this PR or consult `bitcoind -help` to get a list of the options or follow other documentation of Bitcoin Core.

We should avoid forcing users dig through the source code. Is there a way to list all nix-bitcoin options (like man configuration.nix), ideally in the examples/ nix-shell? We could add the original name to the corresponding option description such that users could search for it in the nix-bitcoin option list.

modules/spark-wallet.nix
```diff
@@ -5,18 +5,17 @@ with lib;
...
-  onion-chef-service = (if cfg.onion-service then [ "onion-chef.service" ] else []);
 
   # Use wasabi rate provider because the default (bitstamp) doesn't accept
   # connections through Tor
   torRateProvider = "--rate-provider wasabi --proxy socks5h://${config.services.tor.client.socksListenAddress}";
   startScript = ''
-    ${optionalString cfg.onion-service ''
-      publicURL="--public-url http://$(cat /var/lib/onion-chef/spark-wallet/spark-wallet)"
+    ${optionalString (cfg.getAnnounceAddressCmd != "") ''
+      publicURL="--public-url http://$(${cfg.getAnnounceAddressCmd})"
```
jonasnick, line 14: `getAnnounceAddressCmd` doesn't seem to be a very fitting in this context especially. Why not just `getOnionAddressCmd`?

docs/usage.md
```diff
@@ -8,7 +8,7 @@ fetch-release > nix-bitcoin-release.nix
 
 Nodeinfo
 ---
-Run `nodeinfo` to see your onion addresses for the webindex, spark, etc. if they are enabled.
+Run `nodeinfo` to see onion addresses and local addresses for enabled services.
```
jonasnick, line 11: There are two other lines that need to be changed in usage.md (`grep '{SSHD_ONION,ELECTRS_ONION}'` won't work)

modules/btcpayserver.nix
```diff
@@ -117,9 +117,9 @@ in {
       configFile = builtins.toFile "config" ''
         network=${config.services.bitcoind.network}
         btcrpcuser=${cfg.bitcoind.rpc.users.btcpayserver.name}
-        btcrpcurl=http://${config.services.bitcoind.rpcbind}:${toString cfg.bitcoind.rpc.port}
-        btcnodeendpoint=${config.services.bitcoind.bind}:8333
-        bind=${cfg.nbxplorer.bind}
+        btcrpcurl=http://${config.services.bitcoind.rpc.address}:${toString cfg.bitcoind.rpc.port}
+        btcnodeendpoint=${config.services.bitcoind.address}:8333
```
jonasnick, line 121: Is the static port 8333 intentional (instead of `bitcoind.port`)?

modules/onion-services.nix
```diff
@@ -0,0 +1,100 @@
...
+          }
+        );
+      };
+
+      # Enable announcing services to access their own onion addresses
+      nix-bitcoin.onionAddresses.access = (
+        genAttrs announcingServices singleton
+      ) // {
+        # The operator user can access onion addresses for all active services
+        operator = mkIf config.nix-bitcoin.operator.enable activeServices;
```
jonasnick, line 76: `nodeinfo` doesn't work if the operator gets a non-default name. Not sure if that's related to this line.

modules/nix-bitcoin-services.nix
```diff
@@ -51,6 +51,15 @@ let self = {
     '';
   };
 
+  getAnnounceAddressCmd = mkOption {
```
jonasnick, line 54: This seems to be the wrong commit for this option because it's already inherited by the spark module in the previous commit.

jonasnick (commented):
Very nice improvements overall. I hope there's a way to make the options more discoverable.

As a sidenote, the new release of bitcoin core should support v3 onion services, so we'll be able to use the `announce` to set `externalip`.

modules/presets/enable-tor.nix
```diff
@@ -0,0 +1,42 @@
...
+    nbxplorer.enforceTor = true;
+    spark-wallet.enforceTor = true;
+    recurring-donations.enforceTor = true;
+  };
+
+  # Add onion services for incoming connections
+  nix-bitcoin.onionServices = {
+    bitcoind.enable = defaultTrue;
+    clightning.enable = defaultTrue;
+    lnd.enable = defaultTrue;
```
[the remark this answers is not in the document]
erikarvstedt: Agreed. I had a fix queued up in another (unreleased PR) as I wanted to limit the scope of this PR, but I've added this now (`fixup! disable default onion services for clightning, lnd, btcpayserver`). 

The default onion service for btcpayserver is a downright security risk because the webinterface allows any visitor to register an admin account on a freshly setup node. (docs: `The first created account on a newly-deployed BTCPay Server is automatically admin.`)

modules/btcpayserver.nix
erikarvstedt: I had saved that for another PR (`Replace direct usage of 8332, 8333 with variables`), but you're right, this belongs here.(`fixup! use option value for btc port`)

modules/nix-bitcoin-services.nix
erikarvstedt: I'll fix this in the final squash.

modules/onion-services.nix
erikarvstedt: Yes, the bug is caused by this line. Thanks, fixed in `fixup! fix onion address access for custom operator names`.

docs/usage.md
erikarvstedt: Fixed in `fixup! update usage.md`.

erikarvstedt:
> getAnnounceAddressCmd doesn't seem to be a very fitting in this context especially

I've intended `getAnnounceAddressCmd` to be an universal interface for setting the public address of a service.
For example, a clearnet node could use the following config:
```
{
  systemd.services.lnd = {
    # A service that fetches and writes the node's public IP to /var/lib/public-ip
    requires = [ "get-ip.service" ];
    after = [ "get-ip.service" ];
  }
  services.lnd = {
    enable = true;
    getAnnounceAddressCmd = "cat /var/lib/public-ip"
  };
}
```
But you're right, `getAnnounceAddressCmd` is not an ideal fit, I've renamed it to `getPublicAddressCmd`.

> announce isn't accurate for spark since nothing is announced

`lnd`, `clightning` (and soon `bitcoind`) announce their public address to peers, while `spark` just displays it in its webinterface.
To better reflect these different ways of services handling their public address, I've renamed `nix.bitcoin.onionServices.*.announce` to the more general term `*.public`.

For each service, I've added details about how the public address is used to the description of `getPublicAddressCmd`.

(Implemented in `fixup! rename announce -> public`)

> (spark) "announce" behavior should just be default if the onion service is enabled.

Fixed in `fixup! set onion-service defaults outside of enable-tor`

> ... a way to make the options more discoverable

Users always had to refer to the source code to discover options and their descriptions, I don't think this problem is aggravated by this PR.
It's relatively simple to make a `man nix-bitcoin` that shows all `nix-bitcoin` options in the same format as `man configuration.nix`.
Can we defer this to another PR?

modules/presets/enable-tor.nix
```diff
@@ -0,0 +1,42 @@
...
+    liquidd.enable = defaultTrue;
+    electrs.enable = defaultTrue;
+    spark-wallet = {
+      enable = defaultTrue;
+      externalPort = 80;
+      announce = true;
```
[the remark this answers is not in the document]
erikarvstedt: See https://github.com/fort-nix/nix-bitcoin/pull/293#issuecomment-758131410.

modules/spark-wallet.nix
erikarvstedt: See https://github.com/fort-nix/nix-bitcoin/pull/293#issuecomment-758131410.

jonasnick (commented):
Looks all good.  Really like the approach of having a `getPublicAddressCmd` universal interface.

> It's relatively simple to make a man nix-bitcoin that shows all nix-bitcoin options 

That's good to hear. Sure, we can defer this.

examples/configuration.nix
```diff
@@ -203,5 +211,5 @@
   # The nix-bitcoin release version that your config is compatible with.
   # When upgrading to a backwards-incompatible release, nix-bitcoin will display an
   # an error and provide hints for migrating your config to the new release.
-  nix-bitcoin.configVersion = "0.0.26";
+  nix-bitcoin.configVersion = "0.0.29";
```
nixbitcoin, line 214: This should be `0.0.30` as `v0.0.29` was released in the meantime and this PR will be released with `v0.0.30`.

modules/versioning.nix
```diff
@@ -5,7 +5,19 @@ let
   version = config.nix-bitcoin.configVersion;
 
   # Sorted by increasing version numbers
-  changes = [
+  changes = let
+    mkOnionServiceChange = service: {
+      version = "0.0.29";
```
nixbitcoin, line 10: This should be `0.0.30` as `v0.0.29` was released in the meantime and this PR will be released with `v0.0.30`.

examples/configuration.nix
```diff
@@ -37,25 +37,32 @@
...
-  # offers outgoing functionality, but doesn't announce the Tor Hidden Service
-  # under which peers can reach us.
-  # services.clightning.announce-tor = true;
+  #
+  # Set this to create an onion service by which clightning can accept incoming
+  # connections via Tor.
+  # nix-bitcoin.onionServices.clightning.enable = true;
+  #
+  # Set this to enable clightning to announce the onion service to other peers.
+  # nix-bitcoin.onionServices.clightning.public = true;
```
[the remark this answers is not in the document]
nixbitcoin: Somebody might want to enable the onion service without announcing it though...

modules/spark-wallet.nix
```diff
@@ -31,24 +30,30 @@ in {
...
       type = types.separatedString " ";
       default = "";
       description = "Extra command line arguments passed to spark-wallet.";
     };
-    enforceTor =  nix-bitcoin-services.enforceTor;
+    getPublicAddressCmd = mkOption {
+      type = types.str;
+      default = "";
+      description = ''
+        Bash expression which outputs the public service address to display in the web interface.
```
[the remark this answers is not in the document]
nixbitcoin: µ-nit

examples/configuration.nix
```diff
@@ -93,6 +100,12 @@
   # The lightning backend service automatically enabled.
   # Afterwards you need to go into Store > General Settings > Lightning Nodes
   # and click to use "the internal lightning node of this BTCPay Server".
+  #
+  # Set this to create an onion service to make the btcpayserver web interface
+  # accessible via Tor.
+  # Security WARNING: Create a bitcpayserver administrator account before allowing
```
nixbitcoin, line 106: `btcpayserver` not `bitcpayserver`

examples/configuration.nix
```diff
@@ -93,6 +100,12 @@
...
+  # public access to the web interface.
+  # nix-bitcoin.onionServices.bitcpayserver.enable = true;
```
nixbitcoin, line 108: `btcpayserver` not `bitcpayserver`

examples/configuration.nix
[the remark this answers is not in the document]
jonasnick: Sure I'm only saying announce implies public, not the other way around. We wouldn't need to mention public in the example config, because public without announce is a rather advanced use case.

nixbitcoin:
Great improvements! I think we can merge this PR mod nits.

examples/configuration.nix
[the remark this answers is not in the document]
nixbitcoin: You're right. It shouldn't be possible to set `onionServices.service.public` when `onionServices.service.enable = false` anyway.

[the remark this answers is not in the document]
nixbitcoin: Also we need to change the wording in `README.md` to reflect that onionServices are no longer created for every service automatically. Ex:

`All applications use Tor for outbound connections and accept inbound connections via onion services.` -> `All applications use Tor for outbound connections and support accepting inbound connections via onion services.`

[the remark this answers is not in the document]
erikarvstedt: Fixed in 6495bf5aaa19c856f8514d1bdba335a233f7c5eb.

modules/spark-wallet.nix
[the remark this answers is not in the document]
erikarvstedt: [The](https://twitter.com/nixbitcoinorg/status/1346851008446668802) 27 year old nix-bitcoin developer: enthusiastically addressing µ-nits at 1 am.
Actually, Jonas' comment was very relevant — because my description was totally wrong. The QR code is never shown in the web interface. Fixed in 3b5bb802407b4188abbf3998b28aa70750e6a4a9.

examples/configuration.nix
erikarvstedt: Fixed.

modules/versioning.nix
erikarvstedt: Fixed.

examples/configuration.nix
erikarvstedt: Fixed.

examples/configuration.nix
erikarvstedt: Fixed.

erikarvstedt:
Squashed. The old [unsquashed version is here](https://github.com/erikarvstedt/nix-bitcoin/commits/onion-services-unsquashed).

modules/onion-services.nix
```diff
@@ -103,6 +103,14 @@ in {
     # Set sensible defaults for some services
     {
       nix-bitcoin.onionServices = {
+        spark-wallet = {
+          externalPort = 80;
+          # Enable 'public', but don't auto-enable the onion service.
```
nixbitcoin, line 108: What's the rationale here? When would a user want to publish an onion service that doesn't exist?

erikarvstedt: I've extended the comment. We want to set `public = true` as the default for spark-wallet, without enabling the onionService by default.

nixbitcoin: Why's that? One can only publish a hidden service that exists, right?

erikarvstedt: `onionServices` entries that are `public` but not `enabled` are ignored.

nixbitcoin: > We want to set `public = true` as the default for spark-wallet, without enabling the onionService by default.

Can you elaborate why?

erikarvstedt: To quote @jonasnick:
> The "announce" behavior should just be default if the onion service is enabled. 

`public` for spark-wallet has no disadvantages and should be enabled by default when a users sets `onionServices.spark-wallet.enable = true`.

nixbitcoin: Oh, I think I get it. The `enable` is the exposed option the user toggles, which will always include the `.onion` in the QR code through the `true` `public` option.

nixbitcoin (approved):
```
Hash: SHA512

ACK e2922eb4ce6b820fd1bf698c6aadce5d5f4d27c6
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCgAdFiEEV3o0Un8+KoXoD+Fk3RH5rVMIs7oFAmAAPVgACgkQ3RH5rVMI
s7pJcRAAsMfOzQTDju1slfPVm57CH5n2uiv8EqSRHoTFTTaLjwli7Y86TfuxLXI8
ixHZLiIx2zRbHWUElsP57sWFa8pn+dvgjQUpziNl2ZnJJ8IiPYk2LAgKhDNAyW8w
VVF1rWhSwsZHmn+uqstv1Igpcmc34FZZ0kyAAjgvNZOzP07xxYQvKgTjLO5dmL8h
xdSaeJExNIwW3C4tCu2a8hGJgcqoKRdo9C0yBml7/UI2ZSSOLV+6Qgg1Ym3elWBB
E2gA3VkfKBnvnUhAeB9RLgRebNM/o8sw8/RSg0uP2aCtDzG52+KBJQaV+ubcaxOl
XFFXvajwEMy5PHvGt+mEdN9tPlEN9g6aYDcMXSXB7cRF9NFDuHJStvZpTn3XGgw6
NoAflonSiB88zuPxDnVM2+mSaIgSnD8MIMTH4iyF59Luc/VoKNIf5HQVq2HNSnnK
tcaYm+xqbCt7MABu28D3KuB31Hh2iFPCGZB4f5d3lNtUERi5S4u0YZnjLyKSJHhQ
SGpshgGbfirO4dQ02KskC2wUpnSJRvFG1nGhHOFxJ9xI6KG40oN9l5cJJNL15Z4C
5G/yaTk7ybf569fql8N/PmcTNejyznIFOLR0aBA/BMC8lWnWgxWxEH0dl1cvsasj
tBtTXgtk45dxSHvO9qRt8ipRdwJRX9VENuoVsQCbp1s6xTL27rc=
=wqhU
-----END PGP SIGNATURE-----
```

jonasnick (commented):
ACK e2922eb4ce6b820fd1bf698c6aadce5d5f4d27c6

PR #296, opened 2021-01-03
Remove nixops examples and as recommended deployment method; replace with krops

### LINKED ISSUE #115: Consider replacing nixops with krops
I just learned about krops (https://tech.ingolf-wagner.de/nixos/krops/) which is simpler than nixops and sufficient for most applications.

* pro
   * native file encryption
   * build happens on target 
       * easier for macOS users
       *  easier to deploy to arm
       * using no substitutes won't slow down your personal computer
* con
    * no easy virtualbox tutorial anymore
        * but we could provide a build with [nixos-generate](https://github.com/nix-community/nixos-generators)

jonasnick:
Fixes #115

Just recently a new nix-bitcoin user told me about his problems when setting up nix-bitcoin, some of which were related to nixops. As far as I can tell, krops is much better suited for simple deployments such as a nix-bitcoin node. Krops should also make it easier to use nix-bitcoin without substitutes. And krops is somewhat future-proof because it can already deploy a machine with flakes by using a custom `writeCommand`.

As you can see, we have to chmod the secrets dir before nixops copies them over because it also copies the permissions from the local machine. That's not great. Ideally, one would be able to set the target permissions when specifying the krops `source`.

Existing nixops deployments should continue to work.

WIP because I need to write a migration from nixops guide and test it.


nixbitcoin:
The PR at this point breaks our secrets deployment mechanism. It's not enough to copy all secrets to a root owned directory with read-only permissions.

nixbitcoin:
[Initial Fixups](https://github.com/nixbitcoin/nix-bitcoin/commits/remove-nixops-nb) (permissioned secrets, install instructions, and migration guide)

jonasnick:
Oh, I see how I messed up. Should definitely have a test or "example" with krops to make sure this happens again. Haven't read your code yet but cherry-picked into my branch to simplify the review. Also cherry-picked the other commits, except the
- `fixup! install instructions: use latest-nixos-minimal iso` bec. it's nice to have nix-bitcoin provide the hash independently (but yeah we need to remember updating this)
- `fixup! install instructions: use dd instead of cp to burn to media` bec. iirc we changed this to cp at one point, let's just leave as is

examples/krops/node.nix
```diff
@@ -0,0 +1,27 @@
...
+  nixBitcoinPkgs = import <nix-bitcoin> {};
+  source = nixBitcoinPkgs.krops.lib.evalSource [{
+    nixpkgs.file = {
+      path = toString <nixpkgs>;
+      useChecksum = true;
+    };
+    nix-bitcoin.file = {
+      path = toString <nix-bitcoin>;
+      useChecksum = true;
+    };
```
nixbitcoin, line 15: I think we should exclude the `.git` folder for people deploying from git repos
```suggestion
    filters = [
      {
        type = "exclude";
        pattern = ".git";
      }
    ];
```

docs/install.md
```diff
@@ -276,60 +89,74 @@ This is borrowed from the [NixOS manual](https://nixos.org/nixos/manual/index.ht
     nano /mnt/etc/nixos/configuration.nix
     ```
 
-    Option 1: Edit NixOS configuration for UEFI
+    We now need to adjust the configuration to make sure that we can ssh into the system and that it boots correctly. We add some lines to set `services.openssh` such that the configuration looks as follows:
```
0xB10C, line 92: Ideally, a public key for ssh should be set here already (when setting up the root login before we deploy nix-bitcoin).  `krops-deploy` asks for the root password for each operation (easily 15-20).

0xB10C (commented):
Setup a fresh nix-bitcoin deployment following the new krops instructions. I highly prefer krops over NixOps.

docs/install.md
```diff
@@ -342,88 +169,143 @@ This is borrowed from the [NixOS manual](https://nixos.org/nixos/manual/index.ht
...
     ```
-    Option 2: Set grub device for Legacy Boot (MBR)
+
+    Locate the `FIXME` and set the target to the name of the ssh config entry created earlier, i.e. `bitcoin-node`.
+
+3. Copy `hardware-configuration.nix` from your node to the deployment directory.
+
     ```
-    boot.loader.grub.device = "/dev/sda";
+    scp /etc/nixos/hardware-configuration.nix ./
```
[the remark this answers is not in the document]
jonasnick: Thanks @0xB10C ... fixed.

docs/install.md
jonasnick: Hmm, I don't remember this. Will need to reproduce it. Not so easy to type your ssh key.

jonasnick:
Rebased and added a deployment example that can serve as a test. I still haven't really looked at the password module yet (but it appears to be working fine).

examples/krops/node.nix
jonasnick: Fixed.

nixbitcoin:
Some [fixups](https://github.com/nixbitcoin/nix-bitcoin/commits/remove-nixops-nb). I think we can remove "WIP" now.

examples/deploy-krops.sh
```diff
@@ -0,0 +1,149 @@
...
+
+identityFile=qemu-vm/id-vm
+chmod 0600 $identityFile
+
+echo "Building VM"
+nix-build --out-link $tmpDir/vm - <<EOF
+(import <nixpkgs/nixos> {
+  configuration = {
+    virtualisation.graphics = false;
+    virtualisation.diskSize = 8192;
```
nixbitcoin, line 41: `qemu-vm.nix` makes the Nix store in the VM writable by layering an overlay filesystem on top of the host's Nix store. This overlay has a very limited size when using Tmpfs. Instead we should use `virtualisation.writableStoreUseTmpfs = false;` which makes the VM write to its own filesystem (8192MB) so we can use the writableStore with more featured deployments.

erikarvstedt:
[Fixed branch](https://github.com/erikarvstedt/nix-bitcoin/commits/remove-nixops-ea).

Last year I published [a rough draft](https://github.com/erikarvstedt/nix-bitcoin/commit/6a4c97e1e53b77af02632b86c82137647108d59f) for a krops deployment example.
I've adapted its main features to this PR:
- Build the nix-bitcoin node config on the main host and not inside the VM.
  This significantly speeds up the build and allows caching the build output in the main host's nix store. (This also obviates the need for `virtualisation.writableStoreUseTmpfs = false`.)
- Parallelize VM startup and the nix build tasks.
- Deploy the nixpkgs source via a symlink instead of copying.
- Share common implementation details with `deploy-qemu-vm.sh`.

More changes:
- `examples: add krops deployment method`
  - Misc. refactorings
  - `deployment/krops.nix` is greatly simplified by using the existing `setup-secrets` mechanism.
  - Add `krops-configuration.nix` that imports `configuration.nix` so that `configuration.nix` isn't used to define deployment-specific settings.
    This allows the same `configuration.nix`to be used with different deployment methods.
    This also allows removing step `7. Enable krops in your configuration.nix` in `nixops-krops-migration.md`
  - Explain why `useChecksum` is used in `krops/sources.nix`.
  - Rename `node.nix` -> `deploy.nix`. `deploy.nix` generates a krops deploy script, so that's more fitting.
  - `krops-deploy` (`shell.nix`):
    Only `chmod 700 secrets/`. Don't chmod the containing secrets files because they are protected by the parent dir permissions. These are also rsynced when deploying via krops.
  - I haven't tested the `krops-deploy` shell.nix function, could you please do that?

I've updated the `docs` commits to reflect the above changes.

The hardened profile is now disabled for the example VMs (`examples/qemu: disable hardened profile`).
For future reference: The hardened profile can be deployed to a VM without modification when `apparmor` is enabled in the initial VM config:
```
security.apparmor.enable = true;
```

jonasnick:
Pushed the squashed #336 and added a few doc fixups.

erikarvstedt:
ACK.
Could you rebase so that the `nixops19_09` build in `nixops-krops-migration.md`, step 10 succeeds?

jonasnick:
rebased & squashed

nixbitcoin (approved):
```
Hash: SHA512

ACK 0433a0a767f414378fbcec3b465b26e6ab04586c
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCgAdFiEEiDOlIkszv4ojhh+OtgROy6La5dAFAmBPdqIACgkQtgROy6La
5dB4VBAAgers25to3S2TGuCO8XCwGDB0JGZkoTJZ3hC23gRqo7s/7hPtzc09xoDG
USnz2swSdX9Z5TY5Ud9Q/mK8Jh47disi35uAaHc1a7zHomq8YHhmzN2u3kBJg7aO
sCMwMBfaJOhe9dAvBaLr2h5K0N90tlIODByHBU1U7a5cO4YUArDYuV3v9UNvZKEb
dlTwxBDWFL2zuZ2ax9eeyspOwv/iRtyGviT2zE7uouGkXG0UWF0H+9T1Z+901tHi
cxe1NWc22jrSAY44fIeKgwV3O9Vb/uqHJ1JJrmTHqt+Sk2sxR95RcAXfhVqfAclO
kSbHT2UMcUHyx8Ke8HDpr+nD87ZUPeU8hf7iNhy2+KdM71u1tdVj56UjTiZCuC+Q
acQ46zigWED8hFpC5r4zEXUcaLVtpyS4epRewzWM8+BZTQ57Y+/xsm5icj9B5AEf
Xnl8Co265QYRZ+ihBprLm8GTriUM/5c6s8iS7ENhKEuGJ//U3sYnRhP9BAXM5oKX
jpLsCwQLm4yopalqfg/u30aXylBwzEEZSnJ7lL8vT8pq8haJ3C1sD2L3bfe+ifep
cBO74Se5eeWa7P/o/g4kyWoW5/yCegZF82nCQfbk8ToBr25L1aFwhFvvqHRYWRjB
PA8yl0Dyk9mlTcwFD1QtQm7UscUj2qOSzbtcxCYh73MHMb+eJLg=
=2PEL
-----END PGP SIGNATURE-----
```

nixbitcoin:
Tested
* deployment with (legacy) nixops
* migration of nixops node

erikarvstedt:
[Fixups](https://github.com/erikarvstedt/nix-bitcoin/commits/remove-nixops-ea2).
[Squashed version](https://github.com/erikarvstedt/nix-bitcoin/commits/remove-nixops-ea2-squashed).

jonasnick:
Pushed squashed branch.

erikarvstedt (approved):
ACK 520a983862728320ca6fec3677e5fcbf57d05217

erikarvstedt:
Maybe remove the `WIP` in the PR title before merging.

galderz:
So, what happens with those who have setup their nodes already via `nixops`? Could you add some migration information for us?

galderz:
Or, is https://github.com/fort-nix/nix-bitcoin/blob/master/docs/nixops-krops-migration.md what we need to follow to migrate our nodes to be managed by `krops`?

jonasnick:
@galderz deployments via `nixops` continue to work. See also the release notes: https://github.com/fort-nix/nix-bitcoin/releases/tag/v0.0.41. If you want to switch to `krops`, you can follow the migration guide you linked to.

erikarvstedt:
@jonasnick, I think we should change the release title
`Supported deployment tool changed from NixOps to krops`
->
`Recommended deployment tool changed from NixOps to krops`

galderz:
@jonasnick I had a fair bit of issues with `nixops` so very happy to kiss that goodbye. I upgraded to `0.0.41` and then did the migration to `krops` and everything appears to have gone smoothly :)

jonasnick:
@erikarvstedt yeah that was a bit misleading, fixed

@galderz happy to hear that :)

PR #299, opened 2021-01-10
joinmarket: add rpcWalletFile option

jonasnick:
The [joinmarket docs](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/docs/USAGE.md#setting-a-core-wallet-recommended) recommend using a separate wallet to avoid mixing up jm and
default wallet.

In the future, ideally joinmarket would create the wallet by default, by using some kind of `wallet` option in the bitcoind module.

### DIFF
--- a/modules/joinmarket.nix
+++ b/modules/joinmarket.nix
@@ -25,6 +25,7 @@ let
     rpc_port = ${toString bitcoind.rpc.port}
     rpc_user = ${bitcoind.rpc.users.privileged.name}
     @@RPC_PASSWORD@@
+    ${optionalString (cfg.rpcWalletFile != null) "rpc_wallet_file=${cfg.rpcWalletFile}"}
 
     [MESSAGING:server1]
     host = darksci3bfoka7tw.onion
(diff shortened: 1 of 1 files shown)


modules/joinmarket.nix
```diff
@@ -25,6 +25,7 @@ let
     rpc_port = ${toString bitcoind.rpc.port}
     rpc_user = ${bitcoind.rpc.users.privileged.name}
     @@RPC_PASSWORD@@
+    ${optionalString (cfg.rpcWalletFile != null) "rpc_wallet_file=${toString cfg.rpcWalletFile}"}
```
erikarvstedt, line 28:
```suggestion
    ${optionalString (cfg.rpcWalletFile != null) "rpc_wallet_file=${cfg.rpcWalletFile}"}
```

modules/joinmarket.nix
```diff
@@ -121,6 +122,14 @@ in {
       default = cfg.user;
       description = "The group as which to run JoinMarket.";
     };
+    rpcWalletFile = mkOption {
+      type = types.nullOr types.str;
+      default = null;
+      description = ''
+        Name of the watch-only bitcoind wallet the JoinMarket addresses are imported
+        to.
```
erikarvstedt, line 130:
```suggestion
        Name of the watch-only bitcoind wallet the JoinMarket addresses are imported to.
```

erikarvstedt (changes requested):

erikarvstedt (approved):
ACK e3a45fcc0c007f1ad3bda1b72ba53d8c55114dd7

PR #306, opened 2021-01-27
Update nixpkgs

nixbitcoin:
Includes CVE-2021-3156 patch

Only `nixpkgs` is updated for now, because my local test suite was failing with the `nixpkgs-unstable` update and speed is crucial on this one. We would also need to enable sharing our bitcoind tor v3 onion service and c-lightning 0.9.3 isn't in `nixpkgs-unstable` anyway.


jonasnick (approved):
ACK f96591c0305f89333e5e233894449147bc0e7662

erikarvstedt:
I'll look into the bitcoind update later today or tomorrow.

PR #307, opened 2021-01-29
Madaidan hardening

nixbitcoin:
This PR implements a number of hardening measures from @madaidan's [Linux Hardening Guide](https://madaidans-insecurities.github.io/guides/linux-hardening.html).

It also replaces sudo with doas because of the recent CVE-2021-3156. Sudo is simply [too complex](https://lobste.rs/s/efsvqu/heap_based_buffer_overflow_sudo_cve_2021#c_c6fcfa) and includes features that a SETUID program should never have. Furthermore, it lets the operator user control selected services so we can discourage users from ssh'ing into the root user at all and instead recommend using the operator user for all normal system management tasks.

I have saved the remaining improvements from @madaidan's [Linux Hardening Guide](https://madaidans-insecurities.github.io/guides/linux-hardening.html), like system-wide AppArmor and LibreSSL, for future PRs because of their complexity.

### DIFF
--- a/modules/presets/hardened-extended.nix
+++ b/modules/presets/hardened-extended.nix
@@ -1,0 +1,141 @@
+# This preset adds additional hardening settings on top of the
+# default ./hardened.nix preset.
+# These settings trade even more functionality and performance for increased security.
+
+# This preset enables usbguard. Use `services.usbguard.rules` to whitelist
+# select devices.
+#
+# See madaidan's Linux Hardening Guide for detailed explanations:
+# https://madaidans-insecurities.github.io/guides/linux-hardening.html
+
+{
+  imports = [
+    # Build on standard hardened preset
+    ./hardened.nix
+  ];
+
+  boot.kernel.sysctl = {
+    # Prevent boot console kernel log information leaks
+    "kernel.printk" = "3 3 3 3";
+
+    # Restrict loading TTY line disciplines to the CAP_SYS_MODULE capability to
+    # prevent unprivileged attackers from loading vulnerable line disciplines with
+    # the TIOCSETD ioctl
+    "dev.tty.ldisc_autoload" = "0";
+
+    # The SysRq key exposes a lot of potentially dangerous debugging functionality
+    # to unprivileged users
+    "kernel.sysrq" = "4";
+
+    # Protect against time-wait assassination by dropping RST packets for sockets
+    # in the time-wait state
+    "net.ipv4.tcp_rfc1337" = "1";
+
+    # Disable accepting IPv6 router advertisements
+    "net.ipv6.conf.all.accept_ra" = "0";
+    "net.ipv6.default.accept_ra" = "0";
+
+    # Disable TCP SACK. SACK is commonly exploited and unnecessary for many
+    # circumstances so it should be disabled if you don't require it
+    "net.ipv4.tcp_sack" = "0";
+    "net.ipv4.tcp_dsack" = "0";
+
+    # Restrict usage of ptrace to only processes with the CAP_SYS_PTRACE
+    # capability
+    "kernel.yama.ptrace_scope" = "2";
+
+    # Prevent creating files in potentially attacker-controlled environments such
+    # as world-writable directories to make data spoofing attacks more difficult
+    "fs.protected_fifos" = "2";
+    "fs.protected_regular" = "2";
+
+    # Avoid leaking system time with TCP timestamps
+    "net.ipv4.tcp_timestamps" = "0";
+
+    # Disable core dumps
+    "syskernel.core_pattern" = "|/bin/false";
+    "fs.suid_dumpable" = "0";
+
+    # Only swap when absolutely necessary
+    "vm.swappiness" = "1";
+  };
+
+  boot.kernelParams = [
+    # Disable slab merging which significantly increases the difficulty of heap
+    # exploitation by preventing overwriting objects from merged caches and by
+    # making it harder to influence slab cache layout
+    "slab_nomerge"
+
+    # Disable vsyscalls as they are obsolete and have been replaced with vDSO.
+    # vsyscalls are also at fixed addresses in memory, making them a potential
+    # target for ROP attacks
+    "vsyscall=none"
+
+    # Disable debugfs which exposes a lot of sensitive information about the
+    # kernel
+    "debugfs=off"
+
+    # Sometimes certain kernel exploits will cause what is known as an "oops".
+    # This parameter will cause the kernel to panic on such oopses, thereby
+    # preventing those exploits
+    "oops=panic"
+
+    # Only allow kernel modules that have been signed with a valid key to be
+    # loaded, which increases security by making it much harder to load a
+    # malicious kernel module
+    "module.sig_enforce=1"
+
+    # The kernel lockdown LSM can eliminate many methods that user space code
+    # could abuse to escalate to kernel privileges and extract sensitive
+    # information. This LSM is necessary to implement a clear security boundary
+    # between user space and the kernel
+    "lockdown=confidentiality"
+
+    # These parameters prevent information leaks during boot and must be used
+    # in combination with the kernel.printk
+    "quiet loglevel=0"
+  ];
+
+  boot.blacklistedKernelModules = [
+    # Obscure networking protocols
+    "dccp"
+    "sctp"
+    "rds"
+    "tipc"
+    "n-hdlc"
+    "x25"
+    "decnet"
+    "econet"
+    "af_802154"
+    "ipx"
+    "appletalk"
+    "psnap"
+    "p8023"
+    "p8022"
+    "can"
+    "atm"
+    # Various rare filesystems
+    "jffs2"
+    "hfsplus"
+    "squashfs"
+    "udf"
+    "cifs"
+    "nfs"
+    "nfsv3"
+    "nfsv4"
+    "gfs2"
+    # vivid driver is only useful for testing purposes and has been the cause
+    # of privilege escalation vulnerabilities
+    "vivid"
+    # Disable Bluetooth
+    "bluetooth"
+    "btusb"
+    # Disable webcam
+    "uvcvideo"
+    # Disable Thunderbolt and FireWire to prevent DMA attacks
+    "thunderbolt"
+    "firewire-core"
+  ];
+
+  services.usbguard.enable = true;
+}
--- a/pkgs/lib.nix
+++ b/pkgs/lib.nix
@@ -15,6 +15,11 @@ let self = {
       MemoryDenyWriteExecute = "true";
       ProtectKernelTunables = "true";
       ProtectKernelModules = "true";
+      ProtectKernelLogs = "true";
+      ProtectClock = "true";
+      # Test and enable these when systemd v247 is available
+      # ProtectProc = "invisible";
+      # ProcSubset = "pid";
       ProtectControlGroups = "true";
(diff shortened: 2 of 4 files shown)


erikarvstedt:
Great work, thank you!

[Fixed branch](https://github.com/erikarvstedt/nix-bitcoin/tree/madaidan-hardening-ea/).

## Changes

### Squashes
Squash all sysctl-related commits into `madaidan: sysctl hardening` and group all sysctl options into one section.
Improve some comments.

Squash `madaidan: disable Thunderbolt and FireWire to prevent DMA attacks` into `madaidan: blacklist kernel modules` and move the commit message to a comment in the source file.

### `doas`
There were some issues with the `doas` implementation:
- joinmarket was broken if operator (and therefore doas) was disabled
- enabling operator forced the whole system to disable `sudo` and to use `doas`, which breaks existing systems

This is fixed in commits `use runuser for dropping privileges` and `services: use doas if enabled`:
`doas` is now automatically used instead of `sudo` when `security.doas.enable == true`.
`secure-node.nix` disables `sudo` and enables `doas` by default.

### Operator service control
Add option `allowServiceControl` for enabling service control.
It's disabled by default, but enabled in `secure-node`.

Optimize the polkit config:
Instead of adding a rule for each service, add a single rule that contains a simple set lookup.

Add a test case.

Allow using the operator module within a reduced nix-bitcoin module set. (A trivial change, see comment "This allows using...".)
I'll explain this further in a future PR.

nixbitcoin:
Pulled in @erikarvstedt's fixes and added a comment in install.md to discourage users from ssh'ing into the root user.

modules/presets/hardened.nix
```diff
@@ -11,4 +11,138 @@
...
+    "kernel.sysrq" = "4";
+
+    # Protect against time-wait assassination by dropping RST packets for sockets
+    # in the time-wait state
+    "net.ipv4.tcp_rfc1337" = "1";
+
+    # Ignore all ICMP requests to avoid Smurf attacks, make the device more
+    # difficult to enumerate on the network and prevent clock fingerprinting
+    # through ICMP timestamps
+    "net.ipv4.icmp_echo_ignore_all" = "1";
```
jonasnick, line 39: In terms of use cases,  there's a difference between a) using a hardening measures with a small performance penalty to increase the cost of a weaponized exploit against clightning or privilege escalation and b) willfully sabotaging your own system because you want to protect against an attack that was hypothetical even back in the 90s. I personally would like to enable a) without also doing b). So it would be best to have an option that allows to choose between the two.  I haven't looked at anything else in this PR yet, just commenting now because it stood out to me.

erikarvstedt: I think we're already providing most of a) by default through systemd `defaultHardening`.
I'm not completely familiar with all features of `hardened.nix` (the NixOS one and the madaidan extensions) but I think they fall mostly, if not exclusively, into category b).
As I've said before, and as [stated by its maintainer](https://github.com/NixOS/nixpkgs/issues/100799#issuecomment-728935064), `hardened.nix` was "never intended for normal deployments". I'd recommend disabling it by default in `configuration.nix`.
What's left to be done is to evaluate which features, if any, of `hardened.nix` could be moved into a separate a) preset or could be enabled by default. We might defer this to another PR.

nixbitcoin: > I haven't looked at anything else in this PR yet, just commenting now because it stood out to me.

If you do, I'm sure you will find many settings that induce no penalty at all but reduce attack surface and make exploitation harder.

I disagree that most of these hardening options fall into category b). Disabling kernel modules/features you don't need, and reducing attack-surface is generally a good thing. Eventhough I take @jonasnick 's point that preemptively disabling ICMP requests is too much, I still see the overall value of leaning towards security by default.

I would propose taking out all the hardening options that induce large performance, functionality and stability penalties and putting them into a separate `hardened-paranoia` profile while keeping `hardened` enabled by default.

Either way I'm happy about the feedback and think we're moving in the right direction on this.

erikarvstedt:
[One simple fixup](https://github.com/erikarvstedt/nix-bitcoin/commits/madaidan-hardening-ea-2).

@nixbitcoin, @jonasnick, does it happen often that you need to manually start or stop a service?
I'm not sure if this use case is worth the rather complex implementation.
In case of a failure, restarting can be done via redeployment, but fixing will require root access anyway in most cases.
Stopping can be done by deploying a changed config.

nixbitcoin:
Rebased and rearranged to make https://github.com/fort-nix/nix-bitcoin/pull/315 possible.

nixbitcoin:
> @nixbitcoin, @jonasnick, does it happen often that you need to manually start or stop a nix-bitcoin related service?

For me yes, especially with `joinmarket-yieldgenerator` which is started manually.

nixbitcoin:
Rebased, placed madaidan hardenings into a separate `hardened-paranoia` profile while keeping hardened enabled by default, and deleted 9501566bef3875b991c309891b078c60f171799c (`let the operator user control selected services`).

madaidan:
@nixbitcoin Blacklisting the ioctl syscall will break almost everything. It'd be much better to whitelist specific ioctl calls by using seccomp to filter syscall arguments like https://github.com/Whonix/sandbox-app-launcher/blob/460c55c0ff8147a61d81f2c10ebed468e516c3ae/usr/share/sandbox-app-launcher/seccomp-whitelist#L118-L131 but I don't think systemd supports that. Alternatively, SELinux could be used: https://kernsec.org/files/lss2015/vanderstoep.pdf

nixbitcoin:
> @nixbitcoin Blacklisting the ioctl syscall will break almost everything. It'd be much better to whitelist specific ioctl calls by using seccomp to filter syscall arguments like https://github.com/Whonix/sandbox-app-launcher/blob/460c55c0ff8147a61d81f2c10ebed468e516c3ae/usr/share/sandbox-app-launcher/seccomp-whitelist#L118-L131 but I don't think systemd supports that. Alternatively, SELinux could be used: https://kernsec.org/files/lss2015/vanderstoep.pdf

I've removed the commit that added `ioctl` to the blacklist. SystemCall argument filtering isn't supported by systemd, SELinux isn't supported by NixOS, and `bwrap` (used in https://github.com/Whonix/sandbox-app-launcher) can't really be used to confine systemd services.

I'm working on integrating system-wide AppArmor, which integrates nicely with systemd. Maybe it will be possible to restrict system call arguments with that.

erikarvstedt:
[Fixed branch](https://github.com/erikarvstedt/nix-bitcoin/commits/madaidan-hardening-ea-3).

Adding the `hardened-paranoia` preset is a self-contained change, so a single commit simplifies reviewing.
I think `hardened-extended` would be a better name because it is more neutral and precise.

nixbitcoin:
Reset onto your branch and changed name to `hardened-extended`

> Because the `hardened` test is very slow, only test the `hardened-paranoia` profile.

I think you forgot to add this change

erikarvstedt:
[Last fix](https://github.com/erikarvstedt/nix-bitcoin/commits/madaidan-hardening-ea-3). I've shortened the overlong comment in `configuration.nix` a bit.

> I think you forgot to add this change

No, the test is still named `hardened`, but it's using the extended preset.

Kixunil:
A friendly warning: going too far with "security" can lead to bad outcomes. Especially things like denying access to various otherwise-public resources.

Two interesting examples:

When adding support for Tor, I've experienced a lot of pain because, in Debian, Tor uses apparmor which blocks access to `/etc/tor`, except for `/etc/tor/torrc`. Yep, Tor is denied from accessing its own directory and I wanted it to include `/etc/tor/torrc.d` directory. This led to huge frustration and extremely difficult debugging. I couldn't figure it out for months and had various symptoms later when I "fixed" it. Multiple instances of `systemctl reload tor@default` had to be rewritten to use `restart` instead, which is slower and God knows what effect it has on privacy.

There's a mandatory password protection in `lnd`. Seemingly good idea, but the consequences are that it can protect you from incredibly unlikely edge-case where attacker has access to LND database but **not** to macaroons nor `tls.key` and can **not** write to the `lnd` binary. This design created **two** security risks instead. Risk of `lnd` being down when old state of channel is published and the risk of attacker stealing if social engineering succeeds or risk of DoS.

I don't know where's the line, what's the right thing to do. I just know that security is not one-sided thing and there can be "too much security".

nixbitcoin:
> [Last fix](https://github.com/erikarvstedt/nix-bitcoin/commits/madaidan-hardening-ea-3). I've shortened the overlong comment in `configuration.nix` a bit.

Squashed

> No, the test is still named `hardened`, but it's using the extended preset.

I added this. It wasn't included in your fixup afa53b630e1 AFAIK.

nixbitcoin:
> A friendly warning: going too far with "security" can lead to bad outcomes. Especially things like denying access to various otherwise-public resources.

I fully agree with you, but these changes simply reduce the kernel's attack surface and systematically implement the principle of least privilege. With NixOS, when something breaks we can easily readjust privilege system-wide and for all users, predictably and deterministically. This is one strength NixOS has over Debian.

erikarvstedt:
Sorry, I made a typo which is fixed [here](https://github.com/erikarvstedt/nix-bitcoin/commits/madaidan-hardening-ea-3).

modules/presets/hardened-extended.nix
```diff
@@ -0,0 +1,141 @@
...
+
+    # The SysRq key exposes a lot of potentially dangerous debugging functionality
+    # to unprivileged users
+    "kernel.sysrq" = "4";
+
+    # Protect against time-wait assassination by dropping RST packets for sockets
+    # in the time-wait state
+    "net.ipv4.tcp_rfc1337" = "1";
+
+    # Disable malicious IPv6 router advertisements
```
jonasnick, line 34: nit: s/malicious/accepting/ because it applies to _all_ router advertisments, not just the malicious ones.

pkgs/lib.nix
```diff
@@ -15,6 +15,11 @@ let self = {
       MemoryDenyWriteExecute = "true";
       ProtectKernelTunables = "true";
       ProtectKernelModules = "true";
+      ProtectKernelLogs = "true";
+      ProtectClock = "true";
+      # Enable these when systemd v247 is available
+      # ProtectProc = "invisible";
+      # ProcSubset = "pid";
```
jonasnick, line 22: We may not actually want to set this because the systemd.exec manpage says for `ProcSubset`
> Note that Linux exposes various kernel APIs via /proc/, which are made unavailable with this setting. Since these APIs are used frequently this option is useful only in a few, specific cases, and is not suitable for most non-trivial programs.

jonasnick (commented):
Enjoyed reading @madaidan's guide. While I don't consider leaking my system time a threat and I don't want to mess with my network, I think this is a useful preset and is easily modifiable.

concept ACK

modules/presets/hardened-extended.nix
nixbitcoin: Fixed in e9b1710

pkgs/lib.nix
nixbitcoin: Fixed in 477684c

Changed wording to specify that we need to carefully test these before we enable them

erikarvstedt:
[Fix typos](https://github.com/erikarvstedt/nix-bitcoin/commits/madaidan-hardening-ea-4).

erikarvstedt (approved):
ACK 1ff5f8f01b52ead9b86e794e7a1c9e8de3f3e40f

Issue #303, opened 2021-01-20
Bitcoind 0.21 problem heads up


Kixunil:
Just a heads up that the new version of bitcoind sometimes gets stuck at start. I discovered it when trying to create a Debian package and [what looks like the same problem was found in bitcoind CI](https://github.com/bitcoin/bitcoin/issues/20654).

It *is* possible that I made some mistake, so you may decide to try it anyway with heavy testing. If you happen to find out please let me know.

I will try to be helpful if anyone wants to investigate but probably will not have a huge amount of time in the upcoming days/weeks.

nixbitcoin:
Thanks for the heads up, but it didn't come up in our testing

Kixunil:
Great that you tried, I will retry myself.

PR #308, opened 2021-01-30
Update nixpkgs-unstable, bitcoind

erikarvstedt:
I couldn't reproduce #303, neither within our test framework nor on my production server.

Use the following for a live test of the new v3 onion service:
```
test/run-tests.sh -s '{
  imports = [ <nix-bitcoin/modules/presets/secure-node.nix> ];
  nix-bitcoin.onionServices.bitcoind.public = true;
}' container

### Run these commands in the container shell

## check that onion service is reachable
onionAddr=$(c nodeinfo | jq -r .bitcoind.onion_address | tr ':' ' ')
echo $onionAddr
# Requires torsocks
torsocks nc -zv $onionAddr

# check config
c cat /var/lib/bitcoind/bitcoin.conf

# check network
c bitcoin-cli getnetworkinfo
```

Commit `improve backup test` fixes the cryptic `AssertionError` that probably happened to @nixbitcoin after updating nixpkgs-unstable.

### DIFF
--- a/examples/configuration.nix
+++ b/examples/configuration.nix
@@ -28,6 +28,10 @@
   # LND and electrs are not compatible with pruning.
   # services.bitcoind.prune = 100000;
   #
+  # Set this to accounce the onion service address to peers.
+  # The onion service allows accepting incoming connections via Tor.
+  # nix-bitcoin.onionServices.bitcoind.public = true;
+  #
   # You can add options that are not defined in modules/bitcoind.nix as follows
   # services.bitcoind.extraConfig = ''
   #   maxorphantx=110
--- a/modules/bitcoind.nix
+++ b/modules/bitcoind.nix
@@ -22,7 +22,7 @@ let
     ${optionalString (cfg.assumevalid != null) "assumevalid=${cfg.assumevalid}"}
 
     # Connection options
-    ${optionalString cfg.listen "bind=${cfg.address}"}
+    ${optionalString cfg.listen "bind=${cfg.address}${optionalString cfg.enforceTor "=onion"}"}
     port=${toString cfg.port}
     ${optionalString (cfg.proxy != null) "proxy=${cfg.proxy}"}
     listen=${if cfg.listen then "1" else "0"}
--- a/modules/bitcoind.nix
+++ b/modules/bitcoind.nix
@@ -67,6 +67,14 @@ in {
         default = 8333;
         description = "Port to listen for peer connections.";
       };
+      getPublicAddressCmd = mkOption {
+        type = types.str;
+        default = "";
+        description = ''
+          Bash expression which outputs the public service address to announce to peers.
+          If left empty, no address is announced.
+        '';
+      };
       package = mkOption {
         type = types.package;
         default = config.nix-bitcoin.pkgs.bitcoind;
@@ -328,6 +336,10 @@ in {
           ${extraRpcauth}
           ${/* Enable bitcoin-cli for group 'bitcoin' */ ""}
           printf "rpcuser=${cfg.rpc.users.privileged.name}\nrpcpassword="; cat "${secretsDir}/bitcoin-rpcpassword-privileged";
+          echo
+          ${optionalString (cfg.getPublicAddressCmd != "") ''
+            echo "externalip=$(${cfg.getPublicAddressCmd})"
+          ''}
         )
         confFile='${cfg.dataDir}/bitcoin.conf'
         if [[ ! -e $confFile || $cfg != $(cat $confFile) ]]; then
--- a/modules/bitcoind.nix
+++ b/modules/bitcoind.nix
@@ -11,6 +11,8 @@ let
     # We're already logging via journald
     nodebuglogfile=1
 
+    startupnotify=/run/current-system/systemd/bin/systemd-notify --ready
+
     ${optionalString cfg.regtest ''
       regtest=1
       [regtest]
@@ -346,15 +348,12 @@ in {
           install -o '${cfg.user}' -g '${cfg.group}' -m 640 <(echo "$cfg") $confFile
         fi
       '';
-      postStart = ''
-        # Poll until bitcoind accepts commands. This can take a long time.
-        while ! ${cfg.cli}/bin/bitcoin-cli getnetworkinfo &> /dev/null; do
-          sleep 1
-        done
-      '';
       serviceConfig = nix-bitcoin-services.defaultHardening // {
+        Type = "notify";
+        NotifyAccess = "all";
         User = "${cfg.user}";
         Group = "${cfg.group}";
+        TimeoutStartSec = 300;
         ExecStart = "${cfg.package}/bin/bitcoind -datadir='${cfg.dataDir}'";
         Restart = "on-failure";
         UMask = mkIf cfg.dataDirReadableByGroup "0027";
--- a/test/tests.py
+++ b/test/tests.py
@@ -296,12 +296,18 @@ def _():
         "0 differences found",
     )
     # Backup should include important files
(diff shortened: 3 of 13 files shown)


examples/configuration.nix
```diff
@@ -38,9 +42,8 @@
   # in C.
   services.clightning.enable = true;
   #
-  # Set this to create an onion service by which clightning can accept incoming connections
-  # via Tor.
-  # The onion service is automatically announced to peers.
```
jonasnick, line 43: Wasn't this correct? I thought we had fixed this in the onionServices PR. `onionService.clightning.enable` defaults to the setting of public and we don't set `onionService.clightning` in `onion-services.nix` nor in `enable-tor.nix`.

modules/bitcoind.nix
```diff
@@ -328,21 +338,22 @@ in {
...
       '';
-      postStart = ''
-        # Poll until bitcoind accepts commands. This can take a long time.
-        while ! ${cfg.cli}/bin/bitcoin-cli getnetworkinfo &> /dev/null; do
-          sleep 1
-        done
-      '';
       serviceConfig = nix-bitcoin-services.defaultHardening // {
+        Type = "notify";
+        NotifyAccess = "all";
```
jonasnick, line 353: Just for my understanding, why isn't the default "main" enough?

modules/bitcoind.nix
```diff
@@ -328,21 +338,22 @@ in {
...
         User = "${cfg.user}";
         Group = "${cfg.group}";
+        TimeoutStartSec = 300;
```
jonasnick, line 356: IIUC if the node is reindexing (i.e. started with `reindex=1`), we'll only get a notification after that is done. Reindexing is the same as initial sync without block download, so it usually takes hours.

modules/bitcoind.nix
erikarvstedt: I've also wondered why that's the case, because bitcoind's call to `notify` should happen from the main cgroup.
But experimentally only value `all` works, and I didn't feel like digging into this.

examples/configuration.nix
erikarvstedt: Sorry, you're totally correct. I've reverted this commit.

modules/bitcoind.nix
erikarvstedt: I guess we should set`TimeoutStartSec ="infinity"` then.

modules/bitcoind.nix
jonasnick: Okay

jonasnick:
Looks good to me. My machine is still building the packages. Will report back when it's finished.

modules/bitcoind.nix
Kixunil: I believe reindexing doesn't count as a part of startup sequence - the RPC is available during reindexing (you can check status at least). So `startupnotify` command should fire soon enough.

But will check it and reattempt my integration. Thanks for heads up!

modules/bitcoind.nix
Kixunil: FTR I had troubles with it as well and I had to move `systemd-notify` from a separate script to the config directly.

If you are willing to give it one more attempt maybe put `exec` before `systemd-notify` to replace the `sh` process created by `bitcoind` instead of creating yet another child. I didn't try yet, just suspicion.

erikarvstedt: Hah, it seems we've all misread the docs.

Here's the full story:
```
Value | Process(es) allowed to notify
------+--------------------------------------------------------------
main  | The process launched by ExecStart
exec  | The process launched by ExecStart{,Pre,Post,...}
all   | Processes in the cgroups launched by ExecStart{,Pre,Post,...}
```
So when a service uses systemd's `notify` helper binary instead of directly writing to the notify socket, it will always need to use value `all`.

modules/bitcoind.nix
erikarvstedt: @Kixunil, you're right. I've checked the code; [AppInitMain()](https://github.com/bitcoin/bitcoin/blob/16b784d953365bb2d7ae65acd2b20a79ef8ba7b6/src/init.cpp#L1267) doesn't block on reindexing.
[Step 7: load block chain](https://github.com/bitcoin/bitcoin/blob/16b784d953365bb2d7ae65acd2b20a79ef8ba7b6/src/init.cpp#L1530), can take a few minutes, but `TimeoutStartSec = 300` should be sufficient. I've reverted the last change.

Kixunil: I was thinking about checking the code but didn't have enough time, thanks a lot for doing it!

jonasnick (approved):
ACK 506c1cd066949c40ec26f722dcb31cd3e3b320dd

test/tests.py
```diff
@@ -297,7 +297,6 @@ def _():
     )
     # Backup should include important files
     files = {
-        "bitcoind": "var/lib/bitcoind/wallet.dat",
```
nixbitcoin: Shouldn't we still test that the wallet created by `$cli createwallet "test"` is being backed up?

erikarvstedt: Originally, I wanted avoid adding the costly `backups` test to the `regtest` scenario. But there's actually a much simpler and faster solution. Done via [force push](https://github.com/fort-nix/nix-bitcoin/compare/506c1cd066949c40ec26f722dcb31cd3e3b320dd..b114d0c3b129f64ea58ba9c97c7fb91abe3c2057).

nixbitcoin:
I suggest bumping the `nixos-unstable` rev so we include the lnd and clightning updates all in one.

jonasnick (approved):
ACK b114d0c3b129f64ea58ba9c97c7fb91abe3c2057

nixbitcoin (approved):
```
Hash: SHA512

ACK b114d0c3b129f64ea58ba9c97c7fb91abe3c2057
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCgAdFiEEV3o0Un8+KoXoD+Fk3RH5rVMIs7oFAmAX0jwACgkQ3RH5rVMI
s7oeJA//XmQ7aYJefaKTIrQJerEDhHugOCrZtK1v7RjhxXeWTGLUElv7MDjxufu0
8KVHkbUap2vP1fDol+4MPa2ARUSTkZ+Nz4kVd0Pt0yyom/K8m48QxrbxTI3dH060
+uxFLheDLf6rrj45aP7n7P4tPQUTwAyS8Ui1+s9CiJ03GT/h3qXbS7LkxvX7D32Z
cHQCj4eJ2xLZDfvVZ4+tOIDoQX75ICFqTzMfTD1Pd9b3lOG/I/jJA6aER1Xa65tt
UQFIeOwuBMh00utCD30dMS+Yp8Tk4Yjt+hJ4qykunYLIwHa0wPVjos2Ossxt2VXK
XrMRptcPPJ1pvdr3IuuDhLZklsvT0lBbNdZszB/iv9UQ+a1qKeuSM9Z6XQ1YrTQS
qV/YV/Tv/OLZnLHdsEpID00iaeuqB1ZnaTOvBqt7rUkGjVILv5jrCoTFWdFpY1tH
HTrQUGNoh4rF9zLQCuyLsx8bhvpvj3ClMSnCKRCP5faUC7xK09YezZ7ww2GJZzvC
okKFSHOh2dSqYxB8e2E2lQDY5GZEP4VRKqfF1EMDxunINKiAt+HUehK09Qk9dTFa
lsd/n6V+yNBYQO3eN8k4WcP2pbtOJVPieaQR7w48C5NqLHM9mhNSEksHe0Iv86FY
vQSPbKyWxYj8BPuwX5ffxhIWWAJUjNulR7mGf91HYV9tErTBkY4=
=JOPb
-----END PGP SIGNATURE-----
```

PR #309, opened 2021-01-30
fix build-to-cachix

erikarvstedt:
Edit: I've included this PR in #308.

This fixes the cause of the CI build failure in #307.

## Copy of commit message
- Don't fail on `kill $cachixPid` when cachix has already exited.
  This fixes some failing CI builds.

- Stop the cachix background worker before the final `cachix push`.
  This can avoid unneeded reuploads.


erikarvstedt:
I've included this PR in #308 to fix the CI build.

PR #312, opened 2021-02-01
Refactorings, cleanups

erikarvstedt:
Some (mostly minor) refactorings and cleanups that I had piled up over the recent months.

To review commit `services: use consistent layout`, run `test/run-tests.sh eval` on this commit and its parent to see that the system derivation is unchanged.

### DIFF
--- a/modules/joinmarket.nix
+++ b/modules/joinmarket.nix
@@ -185,18 +185,19 @@ in {
         # Generating wallets (jmclient/wallet.py) is only supported for mainnet or testnet
         ExecStartPost = mkIf (bitcoind.network == "mainnet") (nbLib.privileged ''
           walletname=wallet.jmdat
-          pw=$(cat "${secretsDir}"/jm-wallet-password)
-          mnemonic=${secretsDir}/jm-wallet-seed
-          if [[ ! -f ${cfg.dataDir}/wallets/$walletname ]]; then
-            echo Create joinmarket wallet
-            # Use bash variables so commands don't proceed on previous failures
-            # (like with pipes)
-            cd ${cfg.dataDir} && \
-              out=$(sudo -u ${cfg.user} \
-              ${nbPkgs.joinmarket}/bin/jm-genwallet \
-              --datadir=${cfg.dataDir} $walletname $pw)
-            recoveryseed=$(echo "$out" | grep 'recovery_seed')
-            echo "$recoveryseed" | cut -d ':' -f2 > $mnemonic
+          wallet=${cfg.dataDir}/wallets/$walletname
+          if [[ ! -f $wallet ]]; then
+            echo "Create wallet"
+            pw=$(cat "${secretsDir}"/jm-wallet-password)
+            cd ${cfg.dataDir}
+            if ! sudo -u ${cfg.user} ${nbPkgs.joinmarket}/bin/jm-genwallet --datadir=${cfg.dataDir} $walletname $pw \
+                   | grep 'recovery_seed' \
+                   | cut -d ':' -f2 \
+                   | (umask u=r,go=; cat > "${secretsDir}/jm-wallet-seed"); then
+              echo "wallet creation failed"
+              rm -f "$wallet" "${secretsDir}/jm-wallet-seed"
+              exit 1
+            fi
           fi
         '');
         ExecStart = "${nbPkgs.joinmarket}/bin/joinmarketd";
--- a/modules/joinmarket.nix
+++ b/modules/joinmarket.nix
@@ -148,7 +148,7 @@ in {
     ];
     users.users.${cfg.user} = {
         description = "joinmarket User";
-        group = "${cfg.group}";
+        group = cfg.group;
         home = cfg.dataDir;
         extraGroups = [ "tor" ];
     };
@@ -202,11 +202,11 @@ in {
           fi
         '');
         ExecStart = "${nbPkgs.joinmarket}/bin/joinmarketd";
-        WorkingDirectory = "${cfg.dataDir}"; # The service creates 'commitmentlist' in the working dir
-        User = "${cfg.user}";
+        WorkingDirectory = cfg.dataDir; # The service creates 'commitmentlist' in the working dir
+        User = cfg.user;
         Restart = "on-failure";
         RestartSec = "10s";
-        ReadWritePaths = "${cfg.dataDir}";
+        ReadWritePaths = cfg.dataDir;
       } // nbLib.allowTor;
     };
 
@@ -242,10 +242,10 @@ in {
       serviceConfig = nbLib.defaultHardening // rec {
         RuntimeDirectory = "joinmarket-yieldgenerator"; # Only used to create start script
         RuntimeDirectoryMode = "700";
-        WorkingDirectory = "${cfg.dataDir}"; # The service creates dir 'logs' in the working dir
+        WorkingDirectory = cfg.dataDir; # The service creates dir 'logs' in the working dir
         ExecStart = "${pkgs.bash}/bin/bash /run/${RuntimeDirectory}/start";
-        User = "${cfg.user}";
-        ReadWritePaths = "${cfg.dataDir}";
+        User = cfg.user;
+        ReadWritePaths = cfg.dataDir;
       } // nbLib.allowTor;
     };
   })
--- a/modules/joinmarket.nix
+++ b/modules/joinmarket.nix
@@ -168,30 +168,31 @@ in {
       after = [ "bitcoind.service" ];
       path = [ pkgs.sudo ];
       serviceConfig = nbLib.defaultHardening // {
-        ExecStartPre = nbLib.privileged ''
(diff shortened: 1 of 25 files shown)


modules/joinmarket.nix
```diff
@@ -187,29 +177,36 @@ in {
...
-              out=$(sudo -u ${cfg.user} \
-              ${nbPkgs.joinmarket}/bin/jm-genwallet \
-              --datadir=${cfg.dataDir} $walletname $pw)
-            recoveryseed=$(echo "$out" | grep 'recovery_seed')
-            echo "$recoveryseed" | cut -d ':' -f2 > $mnemonic
+            echo "Create wallet"
+            pw=$(cat "${secretsDir}"/jm-wallet-password)
+            cd ${cfg.dataDir}
+            sudo -u ${cfg.user} ${nbPkgs.joinmarket}/bin/jm-genwallet --datadir=${cfg.dataDir} $walletname $pw \
+              | grep 'recovery_seed' \
```
nixbitcoin, line 185: Try replacing with `grep 'recoveryy_seed'` and you will see that the command succeeds eventhough the `grep` fails. That is why I initially added the comment and the bash variables.

erikarvstedt: I've checked that before, it's failing as expected.
Try it yourself via `run-tests.sh -s joinmarket container -d --run c journalctl -u joinmarket`.

nixbitcoin: I deployed with `deploy-qemu-vm.sh` and verified that the `joinmarket` service starts without failure but the `/secrets/jm-wallet-seed` file is empty, when testing with `recoveryy_seed`.

modules/joinmarket.nix
```diff
@@ -141,75 +141,78 @@ in {
...
+            if [[ ! -f ${cfg.dataDir}/wallets/$walletname ]]; then
+              echo "Create wallet"
+              pw=$(cat "${secretsDir}"/jm-wallet-password)
+              cd ${cfg.dataDir}
+              # Instead of redirecting the pipe output to the seed file, store it in a variable.
+              # This prevents writing an empty seed file if the pipe fails
+              walletSeed=$(
+                sudo -u ${cfg.user} ${nbPkgs.joinmarket}/bin/jm-genwallet --datadir=${cfg.dataDir} $walletname $pw \
+                  | grep 'recovery_seed' \
+                  | cut -d ':' -f2
```
nixbitcoin, line 190: I tried again, this time replacing with `cut -d '--' -f2` and the script again succeeded without producing a seed. Can we add `set -u` to make extra sure this script doesn't fail?

erikarvstedt: The initial version of my refactored script behaved in exactly the same way as your version (including failures), except
that my script wrote an empty seed file when the pipe failed (this was fixed by the last squash).
The reason that you saw joinmarket running on your node after a pipe failure is that joinmarket was automatically restarted.
But on the second run it skipped the wallet generation because the wallet file from the failed previous run still existed.
The same thing happened with your original script.
[The fix is](https://github.com/erikarvstedt/nix-bitcoin/commit/6e7b951) to remove the wallet when seed extraction fails.

erikarvstedt:
There's something interesting going on with this CI bug:
The underlying problem is that btcpayserver depends on postgresql, but this dependency is not defined in its systemd service.
When btcpayserver starts before postgresql, it fails with a connection error.
In all CI tests so far and all our local testing this had never happened, postgresql always ran before btcpayserver.
But for this PR, where the modules test on the CI node is preceded by a build of elementsd (caused by [this commit](https://github.com/erikarvstedt/nix-bitcoin/commit/2ac7a92)), the bug is triggered consistently, without exception. This consistency makes it all the more baffling.
[I've now fixed](https://github.com/erikarvstedt/nix-bitcoin/commit/ef9d127) the btcpayserver bug and the test succeeds, but the mystery of why this bug was triggered under these circumstances remains unsolved for now. Any guesses?

nixbitcoin:
Truly baffling, I never had any problems on the local test suite. Great improvements overall!

nixbitcoin (dismissed):
```
Hash: SHA512

ACK 3e6912ad3b2969b33c4cfc9c66dcf06a1fabb766
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCgAdFiEEV3o0Un8+KoXoD+Fk3RH5rVMIs7oFAmAZS2sACgkQ3RH5rVMI
s7rfgA//Z+lPoWOsCxeYaJ8lng/mx5gCaWVkQ9Mpf18/miDPC1BJMUG+cwnpQV0U
kKZyRywwubY/6y9NBEzeiCpvVbfCdIEla2Xgtxfti765HuFyneHgqaBP6vpdBoNB
/172pOyXVshlIzt3ZB7rLhG0pnbqia6OwMUwz4DpmzSySEAREQTSknV5Dfp2K+HY
9Ph906RlkfZ1w0DMrHJpcpGskcgelGv7EvLBELCrfoly6fAs1PqKlWlDI9TTJbHk
CEV3mLlDzle8HAd+pvvRNIC0jBtYjuvGJE0ZGjZcP4ENfBfAeCzOUvHwmKkm7zQl
0gxhFhs5WdbfPwvmgvd5cZTcw6t6n8T1lNv2+W6y47h8ClWIKi0LJFyjtjn128Dx
owqWaVERp19C6loKTp3vSgypgpQgcrQ2Df1Jf1MDwCTQI0y556dEge7EOafn8bO5
C8psEr2+2C3dq2SRNDlbMIhj68m5l2+fhLhgCpOar6nFwuKe/w5MdDsMruxlSOp+
IhBIaKxR3uym/ukljFUZOegKOaXGepvajFNOa2WEIE9cf9osOwMroP0vfiCsRFmo
YQbdGzm4Vh2plCZlPw/yKOXJfrPNxOfWXAouOj6BLlGf9Uj2FHfZT1FIWv/WCmIf
JYZ4UhwwostIdT5CZgiT6WWNA4bCJ5X/dkJJ4vaBVkWh9fWQCvw=
=FzxE
-----END PGP SIGNATURE-----
```

erikarvstedt:
I've pushed a [minor fixup](https://github.com/fort-nix/nix-bitcoin/compare/3e6912ad3b2969b33c4cfc9c66dcf06a1fabb766..97cb2800e04678e2ce38f78d16749b52de94d527).
Explanation: The error message will also be printed if `jm-genwallet` fails, not just on seed extraction errors.

nixbitcoin (approved):
```
Hash: SHA512

ACK 97cb2800e04678e2ce38f78d16749b52de94d527
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCgAdFiEEV3o0Un8+KoXoD+Fk3RH5rVMIs7oFAmAai9QACgkQ3RH5rVMI
s7rLzw/+NkppTJbn6tX/ff1Ezgu1+x/KpiBH2KoxxHOmSmg2hYOCs1SZ47Huvq5G
pk/0MCZEpDq9DkSMf/0lr+QuirD6mnG1HJPEnmcLiy/tS5ZaGAKZo1+3XqP+/YqT
SpuNS30vT2UuQeTq564FPKMKM5H1b51EBkfk+IjMPdTJyWoCnGPqeUETOr3SkiuG
B7FPf5x/piqXsEoFjlRFPpobTFecmi2KnGNOmQG0O0TwgksrEiTcv1TVSRBiVExr
gIO2e7QdfUfO2tQkrc4IwJKnyjhMG6oJMV7AbCMguUkxsg47xESyEBN9XaP4o5dj
xgnwEG4N9MHYFwdAAmD92W5WEE4tLLGCJW+4ijiaPseIHAG/hFDcEIUfnyvKajO+
mrzeMxmUPIIBVFnIMRJPwyxok5avXkpZjqeIDWOIPPXFbZebx24gmd2fk0OwVzs+
6w+ByoB8n9FU5XygXXyYH46H6cmn2vzJDK0QnDJ9w8SVQrNkO91L6b6s+nxhY4q3
gkQAH0oYaUF31qgcB7qcp2byD22Ew7rcCeRlGE6IjYZLJW9DEwjq0tRSIj4oh8br
YYCh5QbqLXDFtiWmxZ1LIzdc9wk6+SzLmi7KDUzn7k+3/4xEyeD9B9CJMQ2XzmRD
/h85e/FI8sPc5jzCQe43Vsw2O8beUuBNPYyJqS8E2oZASF/342c=
=9eDW
-----END PGP SIGNATURE-----
```

jonasnick:
Oops, missed this somehow. Needs rebase.

erikarvstedt:
97cb280 -> 803584a: Rebased to master.

erikarvstedt:
Be sure to not miss commit [`run-tests: add option --copy-src`](https://github.com/fort-nix/nix-bitcoin/pull/312/commits/83622e11b0f4aa7f2ac37e48c339836375bddd00), this greatly improved the dev experience for me.

jonasnick (approved):
Nice! Some option descriptions were directly taken from the help pages of the corresponding applications themselves, but it's fine to rephrase them if we want.

ACK 0a2c8e4864dc30b6d1ed86a16793d37699707650

PR #321, opened 2021-02-12
Update nixpkgs

nixbitcoin:
Includes [CVE-2019-25016](http://cve.circl.lu/cve/CVE-2019-25016) patch

> In OpenDoas from 6.6 to 6.8 the users PATH variable was incorrectly inherited by authenticated executions if the authenticating rule allowed the user to execute any command. Rules that only allowed to authenticated user to execute specific commands were not affected by this issue.

AFAICT this vulnerability wasn't very serious for us because it allowed the `operator` user to execute any command in the `PATH` as the `joinmarket` or `lnd` user, which is the intended behavior. It used to be the default behavior for the software.

### DIFF
--- a/pkgs/nixpkgs-pinned.nix
+++ b/pkgs/nixpkgs-pinned.nix
@@ -8,8 +8,8 @@ in
 {
   # To update, run ../helper/fetch-channel REV
   nixpkgs = fetch {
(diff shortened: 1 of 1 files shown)


erikarvstedt (approved):
ACK 47e54429105ef99514bd4f40f19d990dde34d4a2

PR #324, opened 2021-02-16
Fix lnd onion

nixbitcoin:
This PR fixes an issue with LND reported by @mmilata in the Freenode `#nix-bitcoin` channel.

If both `nix-bitcoin.onionServices.lnd.public` & `services.lnd.restOnionService.enable` were enabled, one would try to create a file named `lnd` and the other would try to create a directory named `lnd` with a file named `lnd-rest` inside it. This would obiously cause an error and fail the LND service. 

A bit of a yak shave, but I had to use `${config.services.${service}.user}` in order to avoid the infinite recursion of `${config.systemd.services.${service}.serviceConfig.User}`.

I've tried my best to locate all uses of hardcoded usernames, but its not guaranteed that all have been found/fixed. Please double-check.

### DIFF
--- a/modules/bitcoind.nix
+++ b/modules/bitcoind.nix
@@ -385,13 +385,13 @@ in {
     users.groups.bitcoinrpc = {};
     nix-bitcoin.operator.groups = [ cfg.group ];
 
-    nix-bitcoin.secrets.bitcoin-rpcpassword-privileged.user = "bitcoin";
+    nix-bitcoin.secrets.bitcoin-rpcpassword-privileged.user = cfg.user;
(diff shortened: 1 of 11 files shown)


mmilata (approved):
Verified that this fixes the problem for me. Thanks!

erikarvstedt:
[Fixups](https://github.com/erikarvstedt/nix-bitcoin/commits/fix-lnd-onion-ea).

erikarvstedt (approved):
ACK ecc601a6d6654f782a9d6b5229ab712e9600b4fe

PR #325, opened 2021-02-17
bitcoind: enable cookie-based authentication

nixbitcoin:
This PR adds an option to enable bitcoind cookie-file based RPC authenticaiton. Programs like [revaultd](https://github.com/re-vault/revaultd/blob/25bb6922986709c642caec2f3ebb6cd750224c6f/src/common/config.rs#L13) only allow bitcoind cookie-file based RPC authentication. Users of these programs should be able to run them under the `bitcoin` user with cookie-file based RPC authentication.

Sadly this only allows `bitcoin` and `root` users to make `bitcoin-cli` calls without manually adding the `privileged` or `public` password.

EDIT: Feature requested by @danielabrozzoni

### DIFF
--- a/modules/bitcoind.nix
+++ b/modules/bitcoind.nix
@@ -327,8 +327,6 @@ in {
         cfg=$(
           cat ${configFile}
           ${extraRpcauth}
-          ${/* Enable bitcoin-cli for group 'bitcoin' */ ""}
-          printf "rpcuser=${cfg.rpc.users.privileged.name}\nrpcpassword="; cat "${secretsDir}/bitcoin-rpcpassword-privileged"
           echo
           ${optionalString (cfg.getPublicAddressCmd != "") ''
             echo "externalip=$(${cfg.getPublicAddressCmd})"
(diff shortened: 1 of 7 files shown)


erikarvstedt:
We could add `chmod g+r ${cfg.dataDir}/.cookie` to `postStart` to enable group access.

nixbitcoin:
Seems like a bad idea, as that would completely void the rpcwhitelisting.

erikarvstedt:
Yeah, but then you should change the `cookieAuth` description. The `restrictive cookie-file permissions` is not the reason we can't provide group access with `cookieAuth`.

nixbitcoin:
> Yeah, but then you should change the `cookieAuth` description. The `restrictive cookie-file permissions` is not the reason we can't provide group access with `cookieAuth`.

What do you suggest?

erikarvstedt:
Why would it `completely void the rpcwhitelisting`?
Currently, group `bitcoin` already has privileged access.

nixbitcoin:
You're completely correct, I had an error in my thinking. Rewrote the PR to enable cookie authentication (with group access) by default and replace `rpcuser`/`rpcpassword`.

nixbitcoin:
Fixed regtest

erikarvstedt:
That's a great solution!
[Fixups and one extra commit](https://github.com/erikarvstedt/nix-bitcoin/commits/bitcoind-cookie-auth-ea).
Regarding `rename group bitcoinrpc -> bitcoinrpc-public`: The linux group name length limit [is 32 byte](https://linux.die.net/man/8/useradd).

nixbitcoin:
Squashed and rebased

erikarvstedt (approved):
ACK 4e9059dc072f09e3d44c0f2a7a841981062c1f72

danielabrozzoni:
Works good, thanks a lot :)

PR #326, opened 2021-02-18
spark-wallet: fix util-linux evaluation error

jb55:
I noticed node-env imports nixpkgs from the local env so `nix build -f . pinned.stable.spark-wallet` doesn't work here. I couldn't think of a fix for this so here's a quick fix for the evaluation issue.


jb55:
hmm maybe my local nixpkgs is just outdated or something...

Issue #323, opened 2021-02-14
Package ‘libvirt-5.9.0’ is marked as insecure, refusing to evaluate.


nixbitcoin:
Steps to reproduce

1. Execute `./deploy-qemu.sh` from master

erikarvstedt:
I can't reproduce this from NixOS. Are you running Arch?
Does the following work when run from the repo root?
```
export NIX_PATH=nixpkgs=$(nix eval --raw -f "pkgs/nixpkgs-pinned.nix" nixpkgs):nix-bitcoin=$(realpath ".")
nix-build --no-out-link - <<EOF
(import <nixpkgs/nixos> {
  configuration = {
    imports = [
      <nix-bitcoin/examples/configuration.nix>
      <nix-bitcoin/modules/secrets/generate-secrets.nix>
    ];
    virtualisation.graphics = false;
    services.mingetty.autologinUser = "root";
  };
}).vm
EOF
```

jonasnick:
I can reproduce the problem. It doesn't have anything to do with qemu per se but with nixops19_09 (which is built when invoking the example nix-shell in deploy-qemu.sh).

```
export NIX_PATH=nixpkgs=$(nix eval --raw -f "pkgs/nixpkgs-pinned.nix" nixpkgs)
nix-build --no-out-link pkgs -A nixops19_09
```

nixbitcoin:
> I can't reproduce this on NixOS. Are you running Arch?
> Does the following script succeed when run from the repo root?
> 
> ```shell
> export NIX_PATH=nixpkgs=$(nix eval --raw -f "pkgs/nixpkgs-pinned.nix" nixpkgs):nix-bitcoin=$(realpath ".")
> nix-build --no-out-link - <<EOF
> (import <nixpkgs/nixos> {
>   configuration = {
>     imports = [
>       <nix-bitcoin/examples/configuration.nix>
>       <nix-bitcoin/modules/secrets/generate-secrets.nix>
>     ];
>     virtualisation.graphics = false;
>     services.mingetty.autologinUser = "root";
>   };
> }).vm
> EOF
> ```

This script does succeed

mmilata:
Would this be a good time to update to newer nixops than what's in 19.09? Not sure whether it has working libvirtd backend but willing to look into it.

galderz:
I'm seeing this. How do I permanently fix this?

galderz:
I see this when doing `nix-shell`

[...]

PR #331, opened 2021-02-23
nixops: remove libvirtd plugin

nixbitcoin:
Simple fix that doesn't force users to edit their `shell.nix`. AFAIK `libvirtd` is not currently needed in nix-bitcoin.

Fixes https://github.com/fort-nix/nix-bitcoin/issues/323

### DIFF
--- a/pkgs/nixops/default.nix
+++ b/pkgs/nixops/default.nix
@@ -24,12 +24,6 @@ let
       rev = "bff6054ce9e7f5f9aa830617577f1a511a461063";
       sha256 = "0j0lbi8rqmw17ji367zh94lvlb062iiyavl4l7m851v40wqr8a5i";
     };
-    libvirtd = {
-      owner = "nix-community";
-      repo = "nixops-libvirtd";
-      rev = "v1.0.0";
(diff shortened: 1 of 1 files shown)


mmilata (approved):
Nixops won't be able to deploy https://github.com/fort-nix/nix-bitcoin/blob/master/examples/nixops/node-libvirtd.nix but I believe it fixes the problem for everyone else.

erikarvstedt:
Superseded by #296.

jonasnick:
Hm, not clear to me in what way #296 supersedes this. We're still going to have a nixops19_09 package, and people may want to continue to use it until they decide to switch. In fact I was under the impression that #323 only affects the qemu example, but of course it breaks any deployment that uses nixops - including mine. So we should fix this as soon as possible.

ACK f1064761d75fa34a6b04de40a0bf7aa6a7d3f78f

erikarvstedt (approved):
ACK f1064761d75fa34a6b04de40a0bf7aa6a7d3f78f

Sorry, I thought that the nixops pkg was deleted in #296.

PR #337, opened 2021-03-09
joinmarket: 0.8.1 -> 0.8.2

nixbitcoin:
Urgent security update, see [release notes](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master/docs/release-notes/release-notes-0.8.2.md)

### DIFF
--- a/modules/joinmarket.nix
+++ b/modules/joinmarket.nix
@@ -87,6 +87,12 @@ let
     minsize = ${toString yg.minsize}
     size_factor = ${toString yg.size_factor}
     gaplimit = 6
+
+    [SNICKER]
+    enabled = false
+    lowest_net_gain = 0
+    servers = cn5lfwvrswicuxn3gjsxoved6l2gu5hdvwy5l3ev7kg6j7lbji2k7hqd.onion,
+    polling_interval_minutes = 60
   '';
 
    # The jm scripts create a 'logs' dir in the working dir,
--- a/pkgs/joinmarket/default.nix
+++ b/pkgs/joinmarket/default.nix
@@ -1,10 +1,10 @@
 { stdenv, lib, fetchurl, python3, nbPython3Packages, pkgs }:
 
 let
-  version = "0.8.1";
+  version = "0.8.2";
   src = fetchurl {
     url = "https://github.com/JoinMarket-Org/joinmarket-clientserver/archive/v${version}.tar.gz";
-    sha256 = "1q3x1x0a78v6apwvbyhl7yh4dgr7xpikd8j07gi3by004ns3789d";
+    sha256 = "0bi1d49kn57b0775cd8gzsc13dbiivvnhrc61d1xb4z1cr3ih8q2";
   };
 
   runtimePackages = with nbPython3Packages; [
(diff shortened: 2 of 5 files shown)


erikarvstedt:
Would joinmarket work with the current `cryptography` version from unstable?

nixbitcoin:
> Would joinmarket work with the current `cryptography` version from unstable?

No, its pinned to 3.3.2 because of [this](https://github.com/pyca/cryptography/issues/5771).

erikarvstedt:
Could you add a comment to `python-packages/cryptography/default.nix` to explain why it was copied/pinned?

nixbitcoin:
> Could you add a comment to `python-packages/cryptography/default.nix` to explain why it was copied/pinned?

Sure, but what do you think about patching the pinning out of `joinmarket` and using the newest rust-based version instead?

erikarvstedt:
>  using the newest rust-based version instead
Yeah, great idea!

nixbitcoin:
This is how far I got, I'm having some issues pulling in `cryptography` from unstable (which on our pinned revision is 3.3.2 - so no patch necessary). Maybe you have some ideas.

erikarvstedt:
[Fix](https://github.com/erikarvstedt/nix-bitcoin/commits/joinmarket-0.8.2-ea).
This pins `cryptography` to the exact required version `3.3.2`.

erikarvstedt:
I think my latest fixup is fine (see the updated post).
To use the unstable version, copy the two cryptography sources from unstable and add a `patchPhase` to `nix-bitcoin/pkgs/joinmarket/default.nix` to patch the version in `jmdaemon/setup.py`.

nixbitcoin:
> I think my latest fixup is fine (see the updated post).
> To use the unstable version, copy the two cryptography sources from unstable and add a `patchPhase` to `nix-bitcoin/pkgs/joinmarket/default.nix` to patch the version in `jmdaemon/setup.py`.

Can't we just pull in the `cryptography` version from our pinned nixpkgs-unstable? That would forego adding another package that we need to actively maintain.

erikarvstedt:
No, because `unstable.python3Packages.cryptography` has conflicting python pkg dependencies (`propagatedBuildInputs`). 
We have to evaluate the cryptography derivation in the context of our own Python package set.

erikarvstedt:
We could also track the cryptography version of `pinned.nixpkgs-unstable`. This would avoid copying the sources, 
but the cryptography version may diverge further from the required version in joinmarket on `unstable` updates. This might introduce bugs.
You can implement this approach like this, but using `pinned.nixpkgs-unstable` instead of `nixpkgsCryptography`:
https://github.com/erikarvstedt/nix-bitcoin/commit/a9cd597123820c0352626124220eda0e29d045a0

nixbitcoin:
> You can implement this approach like this, but using `pinned.nixpkgs-unstable` instead of `nixpkgsCryptography`:
> [erikarvstedt@a9cd597](https://github.com/erikarvstedt/nix-bitcoin/commit/a9cd597123820c0352626124220eda0e29d045a0)

Good idea, but I'm getting `error: undefined variable 'pinned'`

erikarvstedt:
`pinned` was just a shorthand to explain the concept.
Use `unstable = (import ../nixpkgs-pinned.nix).nixpkgs-unstable;` and `callPackage "${unstable}/pkgs/development/...`.

erikarvstedt (approved):
ACK 5ead2a707555d922e0251b2f102dd6df0dab6095

PR #339, opened 2021-03-16
Misc improvements

erikarvstedt:
Commit `examples: improve robustness of deploy scripts` fixes the reason why I couldn't reproduce #323.

Commit `spark-wallet: add flakes compatibility` is a requisite for #327.

I recently ran into some spurious lnd test failures. The [log output](https://gist.github.com/7ad10d8ca985b7b799ec66d3fc59a8b7) `lnd.service: Control process exited, code=exited, status=7/NOTRUNNING` points to curl failing with error 7 (`Failed to connect() to host or proxy`) in the `lnd-create-macaroons` postStart script. I've added commit `lnd: show curl error messages` to help debugging this.

@nixbitcoin: Commit `tests: optimize building multiple tests at once` enables running the nixpkgs btcpayserver update tests in a [reproducible nix-only build](https://gist.github.com/c3d9ce21d91ae350b188bad63602076a).


examples/deploy-container.sh
```diff
@@ -8,10 +8,12 @@ set -euo pipefail
 # Run with option `--interactive` or `-i` to start a shell for interacting with
 # the node.
 
-if [[ ! -v IN_NIX_SHELL ]]; then
+if [[ ! -v NIX_BITCOIN_EXAMPLES_DIR ]]; then
     echo "Running script in nix shell env..."
     cd "${BASH_SOURCE[0]%/*}"
     exec nix-shell --run "./${BASH_SOURCE[0]##*/} $*"
+else
+    cd "$NIX_BITCOIN_EXAMPLES_DIR"
```
nixbitcoin, line 16: Maybe include a comment explaining why, for people who are writing their "own experiments" from this example.

erikarvstedt: I think this is unnecessary because there's only one possible reason for changing the working directory: Because we're using paths relative to this dir. And that's what happens below.
How would you phrase the comment?

nixbitcoin: It's not strictly necessary, but maybe something like

```
# Move into `examples` directory from arbitrary working dir
```

nixbitcoin:
Concept ACK

Tested
* running `examples/deploy-qemu-vm.sh` from `./shell.nix` nix-shell
* `nix-shell --run fetch-release`
* logo printing on non-interactive shells
* building spark-wallet
* LND postStart scripts
* local test suite
* running nixpkgs btcpayserver update tests in a reproducible nix-only build
* tested netns tor enforcement on joinmarket scripts
* tested systemd services tor enforcement

examples/deploy-container.sh
erikarvstedt: In this case, the comment repeats WHAT the code does. 
But I really think the WHAT and also the WHY (see above) of `cd ...`  are self-explanatory and it's okay to leave it as is.

erikarvstedt:
I've pushed a short fixup that actually unlocks the intended perfomance improvements when evaluating multiple tests.

nixbitcoin (approved):
```
Hash: SHA512

ACK 4dc24774bd910959a39eadc111695163ce398ed9
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCgAdFiEEiDOlIkszv4ojhh+OtgROy6La5dAFAmBTLwUACgkQtgROy6La
5dDDkw//V/P2x/yic0hhXRlirxMHCTa0VUQPOP9uVFXCHOqJ6DmNoZpHHlovPDgS
ayHktPSU0An9HdiaPwo+wW+JL5wO7BrwQT6/5bmkJJLQpxsOR96INvhFSPyW8hQF
AY4Qyxa5rloGdt0RIUk655bC/53c2q4EUKwHIhf0rvpPwux3raeyjlgog9/MHOpM
61uDGiDeIV6mclffskBUNyi7UoMv1mgPT9AdhpVA5qJRDzc8bUR7hYjI77h2TCIe
4AmBXiQ+7kH6BH8SGZH34PjdwL/EWz4+IKQRKG1eei6DLeMTinQowcpucy9zXgFB
/eM5VqsrjQHAh8+c86MsR+Y/4weBT0Cao6eW73bgJEYx/NBaGka52BTmBADsYbYP
x1pH1EcdlBSNi6s1qAukkLu7rDLrLEQJx2ItH4BN2vMMPZcs132b5ZhtsT4xEyZj
oItWm9io/IgA+/UQ6pUpy+FxYX3cS/xQ2yuCOaFE0y79gdKFQf5TtixUm7X/EFl4
umgi8DcgKDw1IfbVVLftaqRK3LFXdP3c2y0T1D825uP+/XT2OYKsdOzL4YW0vpxu
vlBPCG6nnIwkKo/881hHco34RoWEpP6F/SgpYF8ny0nHQwdSs+f6+gQWEJaLnboc
8MO6ktgrbvgT02cWh0QCg2cbk6bI3t3SoMHxtf9CVMxlq0QdMo4=
=eeeU
-----END PGP SIGNATURE-----
```

erikarvstedt:
Another short fixup. Avoid copying the whole nixpkgs dir into the store.

erikarvstedt:
[Porting nix-bitcoin](https://github.com/erikarvstedt/nix-bitcoin/commits/nixos-unstable) to NixOS unstable revealed that `RestrictAddressFamilies` has no effect on NixOS 20.09.
I've updated the relevant commit in this PR and renamed it to `services/hardening: allow AF_NETLINK`.

pkgs/lib.nix
```diff
@@ -16,7 +16,8 @@ let self = {
       ProtectKernelTunables = "true";
       ProtectKernelModules = "true";
       ProtectControlGroups = "true";
-      RestrictAddressFamilies = "AF_UNIX AF_INET AF_INET6";
+      # AF_NETLINK is required by network libraries used in bitcoind and lightning-pool
+      RestrictAddressFamilies = "AF_UNIX AF_INET AF_INET6 AF_NETLINK";
```
jonasnick, line 20: You noticed that `RestrictAddressFamilies` has no effect on stable because when porting to NixOS unstable you lightning-pool didn't work and bitcoind continued to work because we had set it to `allowAnyProtocol`, correct? Then, why not add `AF_NETLINK` to lightning-pool only?

> Allowing AF_NETLINK for unprivileged services poses no security risk.

A quick google search tells me otherwise.

Concept ACK to the rest of the PR.

erikarvstedt: `Bitcoind+zmq` without `AF_NETLINK` breaks right at startup.
`lightning-pool` starts successfully but breaks when `bitcoind` finishes IBD. This is visible only in the `regtest` scenario.
I speculated that other services might break in other situations which are not covered by our tests, so I whitelisted `AF_NETLINK` for all services. 
A common use of `AF_NETLINK` by unprivileged processes is to read routing information via the kernel's `rnetlink` socket. 

> A quick google search tells me otherwise.

AFAIK, there have been kernel bugs in the past, but there's no structural risk given the kernel functions correctly.
I'd be okay with service-specific whitelisting, but it might lead to delayed runtime failures with obscure error messages. What do you prefer?

nixbitcoin: > What do you prefer?

I prefer service-specific whitelisting with the possibility of delayed runtime failures with obscure error messages

erikarvstedt: Fixed in `services: add finer-grained address family restrictions`.
I've also added `AF_NETLINK` to the zmq clightning plugin, which failed with no error message at all.

jonasnick: Thanks @erikarvstedt. A bit annoying indeed that all this functionality is untestable in master until NixOS stable gets whatever is necessary from unstable. I noticed you don't give `AF_NETLINK` to `lnd` which had it before. Is that intentional?

jonasnick: Ah it's mentioned in the commit message.

jonasnick (commented):
concept ACK 8edd6c4ccdffd236b4e430d6d68f2dd04046bff1

pkgs/lib.nix
erikarvstedt: > Ah it's mentioned in the commit message.
Yes, a few seconds before your post I forced pushed the following addition to the commit msg:
```
lnd: lnd only runs a zmq client, not a server, therefore it requires
no additional address families.
```

This PR is currently being tested on unstable via https://github.com/fort-nix/nix-bitcoin/pull/344.

erikarvstedt:
I've pulled in two trivial commits from the unstable branch.

nixbitcoin (approved):
```
Hash: SHA512

ACK 4cddf284e96adf8e2e70e228df55ff1efd9d9649
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCgAdFiEEiDOlIkszv4ojhh+OtgROy6La5dAFAmBYutAACgkQtgROy6La
5dAzbQ/+MmQuC0cz/x3hS0uzexkf3nZKMHtPVntDhMJwYZyZXCn71Cf1IPTS0a3J
1Kx6N2CJi8qwPsMzMkB3vLds19ccttR1ETLo9PFtgnv2A+CEa6BFvo6+v//X/ZdS
NndYxbfiOGC11Xio0cXGT/h+JPRi/knb33MRV+dSGyhMkVoK3TnoQnjnmJTewIue
UUnsFT3azYqvW02Xoh9hJ0lYXCXMRYuyb2BBIoyKxGeyOiRzq6QiCFs476uKJanQ
4LeBkZebBsauY1bq3nh8JL1uJvoZiCXRXWdJ4zCBWUCck1UnNF8BwKCq5DWTW8Fp
peljQBgit+6aGx/bRn3o3ItaZJJ78IRTwAH10wmIrz2h42RL/m9YvBApEVfpe0L3
8dLuVUwsw9SgjonMQlyFantB42FGdyakT6Nqld/KmPUH3TwKrMWziJnZKP5SKBko
FJDwguaV6NbTOxC7Sj1WJ1ZVjfjAPl9xkxYT4SRdlpu5uDIBbDBxsp2uKKAsOjlr
lZWsheTo7vwnRIZhfmkIHm/Fp8IrZRUf+9kJA7emb8yJQVX4S3otQK9my+D+usFr
DN3X46SaPmwY8moDacG6evL2lcST5vi5ruj6g+E0hRvW5GDVrnXUR70MjHRgIiBM
cqjot2Tqs1i4hJS307ELvtLmjuKvj20ORiKmkwsx3WUD7EH0Ycc=
=x+og
-----END PGP SIGNATURE-----
```

PR #346, opened 2021-03-24
nixbitcoin.org

nixbitcoin:
I decided to open source the code of the [nixbitcoin.org](https://nixbitcoin.org) website/demo node. I hope that open sourcing it will allow it to serve as a model nix-bitcoin deployment, clean up the code and decentralize control.

I currently have the secrets on a separate unpushed branch, but I could work on converting the secrets directory to a password store, encrypt it to @erikarvstedt's, @jonasnick's, my PGP keys, and add it to this branch. I can also add your SSH keys to the root `openssh.authorizedKeys.keys`.

Also note, that the webdesign isn't in its final stage and could be completely changed.


erikarvstedt:
Concept ACK. But I think this should be added as a separate repo.
This also allows us to maintain a less strict review policy, so you can quickly commit changes as you see fit.

nixbitcoin:
PR moved to https://github.com/fort-nix/nixbitcoin.org/pull/1

PR #347, opened 2021-03-26
lightning-loop: 0.12.0 -> 0.12.1

nixbitcoin:


Kixunil:
Perhaps you could also make use of the new regtest server to implement tests. :)

nixbitcoin:
We already are https://github.com/fort-nix/nix-bitcoin/blob/master/test/tests.py#L394

Kixunil:
Yeah, that's a basic smoke test. I meant to actually try to execute loop in/out on regtest.

erikarvstedt (approved):
ACK cde9597fc44a89f1c58134dbeb9b7783d625d52c

PR #350, opened 2021-04-07
Update nixos unstable

nixbitcoin:
Includes
- btcpayserver: 1.0.7.0 -> 1.0.7.2
- clightning: 0.9.3 -> 0.10.0

Had to pin cryptography to exactly version 3.3.2

Also, update clightning-plugins

### DIFF
--- a/pkgs/clightning-plugins/default.nix
+++ b/pkgs/clightning-plugins/default.nix
@@ -6,8 +6,8 @@ let
   src = pkgs.fetchFromGitHub {
     owner = "lightningd";
     repo = "plugins";
-    rev = "6cd472636926f05a9c472139fabe1ff11c90aa6a";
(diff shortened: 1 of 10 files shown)


erikarvstedt:
As explained previously, we should avoid eval-from-derivation where possible. Hence it's better to copy the cryptography sources, [like so](https://github.com/erikarvstedt/nix-bitcoin/commits/update-nixos-unstable-ea).

nixbitcoin:
Yes, you're right. Reset to your HEAD.

erikarvstedt:
[Final update](https://github.com/erikarvstedt/nix-bitcoin/commits/update-nixos-unstable-ea-2). I've just noticed that you copied the whole recommonmark pkg. We can simply override the version instead.

erikarvstedt (approved):
ACK daeedda8251d8ad8c0bd75c594460b51df21a15c

jonasnick (commented):
utACK

PR #351, opened 2021-04-11
Update CLBOSS, add test, add option for custom pkg

jonasnick:
CLBOSS 0.10 pays an enormous amount in rebalancing fees, which appears to be fixed in CLBOSS 0.11A. It is only a pre-release and apparently has a bug [that prevents rebalancing](https://github.com/ZmnSCPxj/clboss/blob/master/ChangeLog). Hopefully we don't have to wait very long for a new release soon (ping @zmnscpxj :]). The new module option allowing custom packages will help users to run CLBOSS master or test specific PRs.

### DIFF
--- a/modules/clightning-plugins/clboss.nix
+++ b/modules/clightning-plugins/clboss.nix
@@ -12,11 +12,16 @@ let cfg = config.services.clightning.plugins.clboss; in
         Specify target amount (in satoshi) that CLBOSS will leave onchain.
       '';
     };
+    package = mkOption {
+      type = types.package;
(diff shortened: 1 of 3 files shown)


erikarvstedt:
[Minor fixup](https://github.com/erikarvstedt/nix-bitcoin/commits/update-clboss-ea).

nixbitcoin:
ACK with @erikarvstedt fixup

jonasnick:
Simpler indeed... Added fixup. Thanks.

erikarvstedt (approved):
ACK 7914405cede7acd6dab53f3e7965557b6d6c8636

PR #353, opened 2021-04-17
examples: add krops-eval to discover mistakes before deploying

jonasnick:
TODO:
- [ ] add flag to `krops-deploy` to disable `krops-eval`
- [ ] do `krops-eval` in `krops-deploy` in the background and abort deploying if eval fails


erikarvstedt:
Implemented in #385.

PR #363, opened 2021-06-20
Update nixpkgs, fix lnd module to work with 0.13.0

mmilata:
* btcpayserver 1.1.1
* nbxplorer 2.1.51
* lnd 0.13.0
* hwi 2.0.2

New lnd version has changed wallet unlocking API ([release notes#RPC](https://github.com/lightningnetwork/lnd/releases/tag/v0.13.0-beta)) so that it now starts sooner but returns `"waiting to start, RPC services not available"` until lnd boots up. There's also a new parameter/config option to [pass path to file with wallet password](https://github.com/lightningnetwork/lnd/blob/master/docs/wallet.md#auto-unlocking-a-wallet). Perhaps it could be used to simplify `lnd-create-wallet` script, for now retrying the RPC with exponential backoff seems to work though. Related: https://github.com/lightningnetwork/lnd/pull/5150

### DIFF
--- a/modules/lnd.nix
+++ b/modules/lnd.nix
@@ -198,7 +198,7 @@ in {
         RestartSec = "10s";
         ReadWritePaths = cfg.dataDir;
         ExecStartPost = let
-          curl = "${pkgs.curl}/bin/curl -s --show-error";
+          curl = "${pkgs.curl}/bin/curl -s --show-error --retry 10";
           restUrl = "https://${cfg.restAddress}:${toString cfg.restPort}/v1";
         in [
           (nbLib.script "lnd-create-wallet" ''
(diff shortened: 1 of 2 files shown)


jonasnick (commented):
Thanks. If I'm reading the release notes correctly, I think we can now remove the loop that waits "until the RPC port is open".

The actual error that happens when we don't `--retry` in curl is that macaroon creation fails because the wallet isn't fully unlocked yet. So, the correct thing to do before macaroon creation is to call the new State RPC and wait until it's ready. I think `--retry` is ok for now. A retry value as low as 2 worked in my test scripts.

And yes, would be great to unlock with file path but that can happen in a separate PR indeed.

jonasnick:
superseded by #366

PR #371, opened 2021-07-16
secp256k1: replace local version with nixpkgs

prusnak:
nixpkgs contains newer version, so we can upgrade


jonasnick (approved):
ACK f00061f900ce69ebb9e6c003cec96175b947aac0

nixbitcoin:
We use this for JoinMarket because JoinMarket uses rev [0d9540b13ffcd7cd44cc361b8744b93d88aa76ba](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/8c57852b2ae3c5a6ca0ec39724495d391982cfc2/docs/INSTALL.md#installation-on-linux). Could using a different secp256k1 version lead to problems @kristapsk @AdamISZ?

kristapsk:
@nixbitcoin Likely not, but somebody needs to go through changes and test it carefully.

AdamISZ:
(Background: I'll have to go back and check but as I recall fixing that version because coincurve did, and in the absence of secp256k1 releases there was no simple step-change after that. So we should double check if coincurve ever changed it in an update. But complicating this issue considerably: we've almost completely removed the coincurve dependency in Joinmarket (replaced with Simplexum/python-bitcointx), but unfortunately not yet 100%. The main thing that *could* go wrong is in that dependency; our use of the secp256k1 API is pretty simple.
Honestly if I had more time I think I could both answer your question, and remove the coincurve dependency which over complicates our back end, as concrete steps but it'd require some work and I'd love someone else to do it). 

In short: I agree with @kristapsk . It needs to be tested. As well as checking/updating the install script, the test suite is probably sufficient for this though (since it's low level, manual user-level testing isn't needed imo).

kristapsk:
Coincurve has updated secp256k1 to [`f2d9aeae6d5a7c7fbbba8bbb38b1849b784beef7`](https://github.com/bitcoin-core/secp256k1/commit/f2d9aeae6d5a7c7fbbba8bbb38b1849b784beef7) [in January](https://github.com/ofek/coincurve/commit/871fbbdf82190b0dfcc5d21259ad4c104f5348a1), so guess that should be safe to use.

prusnak:
Related: the library maintainers are considering providing tagged versions soon (https://github.com/bitcoin-core/secp256k1/issues/286 + https://github.com/bitcoin-core/secp256k1/issues/856), so the situation hopefully gets better soon.

jonasnick:
@nixbitcoin makes a good point. Since libsecp is unreleased, the libraries should come with their own, vendored version of libsecp. I think coincurve does and python-bitcointx doesn't. I hear that python-bitcointx doesn't recommend a specific commit of libsecp and it's test coverage is only "decent". So, I think at this point it's best is to go with whatever joinmarket suggests using. Thanks for your input @kristapsk @AdamISZ and thanks for opening this PR @prusnak.

I hope the libsecp maintainers get their act together.

prusnak:
I agree with the decision to close this PR and I will re-open the PR when libsecp256k1 has a proper versioned release.

nixbitcoin:
> I think coincurve does and python-bitcointx doesn't

Do you know how coincurve gets its libsecp?

jonasnick:
@nixbitcoin see here https://github.com/ofek/coincurve/blob/master/setup.py#L57

PR #372, opened 2021-07-23
CVE-2021-33909 patch

nixbitcoin:
Includes fix for [CVE-2021-33909](https://lwn.net/Articles/863586/), a local root vulnerability.


nixbitcoin:
Oops, fix is only available on nixos-21.05. We should really update @erikarvstedt

PR #374, opened 2021-08-04
Update to NixOS 21.05

erikarvstedt:
### Commit `security: update /proc restriction mechanism`

@nixbitcoin, you might want to update https://github.com/systemd/systemd/issues/16825:
`systemctl status` still leaks process infos to services running with `ProtectProc`.

### Commit `joinmarket-ob-watcher: allow required 'mbind' system call`

We're disabling `mbind` as suggested by the [docker seccomp blacklist](https://docs.docker.com/engine/security/seccomp/).
But [`mbind`](https://man7.org/linux/man-pages/man2/mbind.2.html) seems to be safe for use in unprivileged processes.
Shouldn't we just allow it in general?

### extra-container

You can use the following to audit the `extra-container` update:

```
tmp=$(mktemp -d -p /tmp)
mkdir -p $tmp/{old,new}
fetchAndCheck() {
    local url=https://github.com/erikarvstedt/extra-container/archive/$1.tar.gz
    curl -sSL $url | tar xz --strip-components=1 -C $2;
    echo "sha256 for $1 is $(nix hash-path --base32 "$2")"
}
fetchAndCheck 0.6 $tmp/old
fetchAndCheck 0.7 $tmp/new
git diff --no-index $tmp/{old,new}
rm -rf $tmp
```

### DIFF
--- a/pkgs/pinned.nix
+++ b/pkgs/pinned.nix
@@ -21,6 +21,8 @@ in
     lightning-loop
     lightning-pool;
 
+  inherit nixpkgsStable nixpkgsUnstable;
+
   stable = nixBitcoinPkgsStable;
   unstable = nixBitcoinPkgsUnstable;
 }
--- a/examples/configuration.nix
+++ b/examples/configuration.nix
@@ -240,7 +240,7 @@
   # compatible, in order to avoid breaking some software such as database
   # servers. You should change this only after NixOS release notes say you
   # should.
-  system.stateVersion = "18.09"; # Did you read the comment?
+  system.stateVersion = "21.05"; # Did you read the comment?
 
   # The nix-bitcoin release version that your config is compatible with.
   # When upgrading to a backwards-incompatible release, nix-bitcoin will display an
--- a/examples/qemu-vm/vm-config.nix
+++ b/examples/qemu-vm/vm-config.nix
@@ -4,7 +4,7 @@
 
   config = {
     virtualisation.graphics = false;
-    services.mingetty.autologinUser = "root";
+    services.getty.autologinUser = "root";
     users.users.root = {
       openssh.authorizedKeys.keyFiles = [ ./id-vm.pub ];
     };
--- a/modules/backups.nix
+++ b/modules/backups.nix
@@ -82,7 +82,6 @@ in {
       services.postgresqlBackup = {
         enable = true;
         databases = [ "btcpaydb" ];
-        startAt = [];
       };
       systemd.services.duplicity = rec {
         wants = [ "postgresqlBackup-btcpaydb.service" ];
--- a/modules/bitcoind.nix
+++ b/modules/bitcoind.nix
@@ -193,7 +193,7 @@ in {
       };
       proxy = mkOption {
         type = types.nullOr types.str;
-        default = if cfg.enforceTor then config.services.tor.client.socksListenAddress else null;
+        default = if cfg.enforceTor then config.nix-bitcoin.torClientAddressWithPort else null;
         description = "Connect through SOCKS5 proxy";
       };
       listen = mkOption {
--- a/modules/bitcoind.nix
+++ b/modules/bitcoind.nix
@@ -388,7 +388,10 @@ in {
       } // nbLib.allowLocalIPAddresses;
     };
 
-    users.users.${cfg.user}.group = cfg.group;
+    users.users.${cfg.user} = {
+      isSystemUser = true;
+      group = cfg.group;
+    };
     users.groups.${cfg.group} = {};
     users.groups.bitcoinrpc-public = {};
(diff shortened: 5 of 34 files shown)


nixbitcoin:
Thanks for taking out the `environment.memoryAllocator.provider = "libc";` It broke a test upgrade to NixOS 21.05 and took down nixbitcoin.org recently. Probably related to https://github.com/NixOS/nixpkgs/issues/127070.

nixbitcoin:
> ### Commit `security: update /proc restriction mechanism`
> 
> @nixbitcoin, you might want to update [systemd/systemd#16825](https://github.com/systemd/systemd/issues/16825):
> `systemctl status` still leaks process infos to services running with `ProtectProc`.

How do you see that?

> ### Commit `joinmarket-ob-watcher: allow required 'mbind' system call`
> 
> We're disabling `mbind` as suggested by the [docker seccomp blacklist](https://docs.docker.com/engine/security/seccomp/).
> But [`mbind`](https://man7.org/linux/man-pages/man2/mbind.2.html) seems to be safe for use in unprivileged processes.
> Shouldn't we just allow it in general?

IMO we should follow the principle of least privilege and only allow it in the services where it is required, like `joinmarket-ob-watcher`. Just curious, how did you figure out `mbind` was the syscall causing issues?

erikarvstedt:
> How do you see that?

Here's a [demo VM](https://gist.githubusercontent.com/erikarvstedt/84338647697bff9d57dcbc036074bdfc/raw/d53447f9f554650bbefaaedbcc1218bc9d109c1c/show-hideproc-bug.sh).

> IMO we should follow the principle of least privilege...

Okay, then we can leave it as is.

erikarvstedt:
Rebased to master.

nixbitcoin:
ACK mod nit:

When I log in as root and execute `doas -u operator bash` commands like `lncli` fail with `doas: not installed setuid`. This wasn't the case with NixOS 20.09. It's ok to ignore for me because users should be ssh'ing in as `operator` anyway to run regular commands.

erikarvstedt:
I've removed `helper/fetch-channel`, because it will be superseded by my upcoming flakes PR.

erikarvstedt:
@nixbitcoin the `doas` error is a NixOS bug. I've made an [upstream PR](https://github.com/NixOS/nixpkgs/pull/133622) and added a workaround to this PR.

nixbitcoin:
ACK c33e3287aebf4e590e8cf3cfd1cc0edf67d1fad8

nixbitcoin (dismissed):
ACK c33e3287aebf4e590e8cf3cfd1cc0edf67d1fad8

pkgs/pinned.nix
```diff
@@ -6,13 +7,18 @@ let
   nixBitcoinPkgsUnstable = import ./. { pkgs = nixpkgsUnstable; };
 in
 {
-  inherit (nixpkgsUnstable)
+  # Pkgs from nixpkgs stable.
+  # Inherit from pkgs instead of nixpkgsStable to avoid evaluating another instance
+  # of nixpkgs
```
jonasnick, line 12: So that means that every time we update nixpkgs we need to check if the version in unstable differs from stable? Not sure if that's worth the hassle since divergence may happen quickly for some packages. Somewhat independendently,
  1. For updating nixpkgs, I have local tooling now to diff the relevant nix expressions which helps writing release notes (and may help finding anomalies).
  2. I think it makes sense in the python tests to additionally check the versions of the packages (e.g. by running `bitcoind -version`) against some hardcoded strings in the python script. That way we can see from a nixpkgs update commit which of the relevant packages were updated.

jonasnick (commented):
Nice! Looks good to me besides that one comment.

Maybe it's just me, but with `ProtectProc=invisible` I wouldn't have expected that a simple `systemctl status` can circumvent that. The option is called PROTECT. The doc says "processes owned by other users are hidden". Well, great that we noticed that. If I'm not mistaken, the  removal of hidepid does not have significant consequences for us. Before, other users could not read bitcoinds rpcpassword from clightning. Now they can, but only if they are not contained in a systemd service. Is that right?

erikarvstedt:
> Is that right?

Yes, exactly. clightning should really switch to directly connecting to the RPC socket, skipping `bitcoin-cli`.

pkgs/pinned.nix
erikarvstedt: I've removed the commit which changes the pkgs pinning. I'll address this in a flakes PR which adds tooling to automate all aspects of nixpkgs updates.

jonasnick: Ok, looking forward to your flakes PR. Sounds exciting!

jonasnick:
ACK a2454975a58ce3d3773f4ec5b2c0cd50738cbcac

jonasnick (approved):

PR #378, opened 2021-08-15
Misc. improvements

erikarvstedt:
Mostly minor cleanups and refactorings that have piled up recently.

The only major functional change is in `backups: backup NixOS uid, gid mappings`.

### DIFF
--- a/modules/nix-bitcoin.nix
+++ b/modules/nix-bitcoin.nix
@@ -1,9 +1,56 @@
-# This file exists only for backwards compatibility
+{ config, pkgs, lib, ... }:
 
-{ lib, ... }:
+with lib;
 {
-   imports = [
-     ./presets/secure-node.nix
-     (lib.mkRemovedOptionModule [ "services" "nix-bitcoin" "enable" ] "Please directly import ./presets/secure-node.nix")
-   ];
+  options = {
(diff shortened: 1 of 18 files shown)


modules/nix-bitcoin.nix
```diff
@@ -1,9 +1,44 @@
-# This file exists only for backwards compatibility
+{ config, pkgs, lib, ... }:
 
-{ lib, ... }:
+with lib;
 {
-   imports = [
-     ./presets/secure-node.nix
```
jonasnick, line 6: We need to change `examples/minimal-configuration.nix` to import `modules.nix` instead of `nix-bitcoin.nix`. Fwiw the minimal container was buggy even before your commit because it imported `secure-node.nix` instead of being minimal.

jonasnick (commented):
Appreciate the cleaning! Looks good to me besides minimal-configuration.nix.

modules/nix-bitcoin.nix
erikarvstedt: Thanks, fixed!

jonasnick (approved):
ACK b0c66c41e13c5def8c31e3ff9ded577fa3e0f65b

PR #380, opened 2021-08-25
joinmarket: 0.8.3 -> 0.9.1

nixbitcoin:


modules/joinmarket.nix
```diff
@@ -235,27 +236,44 @@ in {
...
-          if ! ${nbPkgs.joinmarket}/bin/jm-genwallet --datadir=${cfg.dataDir} $walletname $pw \
-                 | grep 'recovery_seed' \
-                 | cut -d ':' -f2 \
-                 | (umask u=r,go=; cat > jm-wallet-seed); then
+          # Strip trailing newline from password file
+          if ! tr -d "\n" <"${secretsDir}/jm-wallet-password" \
+               | ${nbPkgs.joinmarket}/bin/jm-genwallet \
+                   --datadir=${cfg.dataDir} --wallet-password-stdin $seed $walletname \
+               | grep 'recovery_seed' \
+               | cut -d ':' -f2 \
```
nixbitcoin, line 273: Getting

```
joinmarket-post-start[550448]: cut: write error: Broken pipe`
```

on both `--recovery-seed-file` and normal generation. Can't immediately identify where this is coming from.

nixbitcoin:
Merged fixups and extra commit, see comments for issues.

Maybe we should also note that recreating from seed is the default behavior when a seed is present but no wallet file. Or is the edge case just to esoteric to be applicable to any real person?

modules/joinmarket.nix
erikarvstedt: This error is caused by the subshell containing the `if` statement exiting early when the condition is false. This error never occured in my local testing or in CI.
Fixed [here](https://github.com/erikarvstedt/nix-bitcoin/commits/joinmarket-0.9.1-ea-2).

erikarvstedt:
[Fixed branch](https://github.com/erikarvstedt/nix-bitcoin/commits/joinmarket-0.9.1-ea-2).

- Fix pipe error
- Document that wallets are restored from seed
- Remove obsolete commit `joinmarket: document optional fidelity bond wallet for makers`.

nixbitcoin:
Reset to your HEAD.

> Remove obsolete commit `joinmarket: document optional fidelity bond wallet for makers`.

Oh yeah, I had removed that commit on a previous merge but then forgot when I had to redo some of the squashing. Good catch!

erikarvstedt (approved):
ACK 9730be9282a96ee32305384fcf8749a21acf1f6f

PR #381, opened 2021-09-02
run-tests: Fix 'eval' command for newer versions of nix

erikarvstedt:
#### Copy of main commit  msg
There's no common `nix` command argument syntax for eval'ing a nix expression that supports both older and newer (flake support) versions of nix.
So fall back to nix-instantiate.


jonasnick (approved):
ACK 32ce9d0ff40dda8670a111d9b16dabf4857741d0

Confirmed that this works for Nix 2.3 and 2.4.

Issue #395, opened 2021-09-23
Possible to use nix-bitcoin to run bitcoin services on NixOS?


casey:
I have a nixos server that I configure with `/etc/nixos/configuration.nix`. I would like to run an LND node, but LND isn't yet packaged in nixpkgs, nor configurable in nixos. I was hoping to use nix-bitcoin to run LND, but was unable to figure out how to integrate the configuration into my `/etc/nixos/configuration.nix`. Is this possible, and is there documentation for it? Thank you!

erikarvstedt:
Yes, nix-bitcoin has explicit support for integration into existing configs.
We have a `deploy-container-minimal.sh` [example](https://github.com/fort-nix/nix-bitcoin/tree/master/examples) to cover this use case. I've now improved this by [adding](https://github.com/fort-nix/nix-bitcoin/pull/396) a [standalone-configuration.nix](https://github.com/fort-nix/nix-bitcoin/blob/2f2d0a2203edf70dde58d4bfb8b433558ff7a0f4/examples/standalone-configuration.nix) example file which you can directly copy to your system config.

prusnak:
> LND isn't yet packaged in nixpkgs, nor configurable in nixos

LND is packaged in nixpkgs - see https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/blockchains/lnd/default.nix

However, I see there is indeed no module similar to https://github.com/fort-nix/nix-bitcoin/blob/master/modules/lnd.nix in nixpkgs.

jonasnick:
Added more importable config and documentation in #396.

PR #396, opened 2021-09-23
examples: add importable-configuration.nix

erikarvstedt:
Motivated by #395.


nixbitcoin (commented):
Concept ACK. @casey does this fulfill your use-case?

examples/README.md
```diff
@@ -17,19 +17,19 @@ By default, [`configuration.nix`](configuration.nix) enables `bitcoind` and `cli
   This is the fastest way to set up a node.\
   Requires: [Nix](https://nixos.org/), a systemd-based Linux distro and root privileges
 
+- [`./deploy-container-minimal.sh`](deploy-container-minimal.sh) creates a
+  container defined by [standalone-configuration.nix](standalone-configuration.nix).\
+  You can copy and import this file to use nix-bitcoin in an existing NixOS configuration.\
+  The configuration doesn't use the [secure-node.nix](../modules/presets/secure-node.nix) preset.\
+  Requires: [Nix](https://nixos.org/), a systemd-based Linux distro and root privileges
```
casey, line 24: I think that if I were skimming paragraph, I might read the first sentence, think it only applied if you want to run `nix-bitcoin` in a container, and miss the rest.

So maybe it could be broken out into its own section:

```suggestion

-  You can copy and import this [standalone-configuration.nix](standalone-configuration.nix) to use nix-bitcoin in an existing NixOS configuration.\
  The configuration doesn't use the [secure-node.nix](../modules/presets/secure-node.nix) preset.\
  Requires: [Nix](https://nixos.org/), a systemd-based Linux distro and root privileges
```

casey (commented):
I added a comment suggestion breaking the item in `examples/README.md` in two to make it easier to notice while skimming.

It's not totally clear to me what "standalone" in `standalone-configuration.nix` refers to, maybe `nixos-configuration.nix` would be clearer?

casey:
> Concept ACK. @casey does this fulfill your use-case?

Yes, this definitely works for me.

jonasnick:
Looks good to me.

> It's not totally clear to me what "standalone" in standalone-configuration.nix refers to, maybe nixos-configuration.nix would be clearer?

Agree on the first point. Isn't it the opposite of "standalone" because it's supposed to be imported? Not sure if `nixos-configuration.nix` is more understandable because all configuration.nix files are NixOS configuration files. Alternatively `simple-configuration.nix`, `minimal-configuration.nix`, `importable-configuration.nix`?

erikarvstedt:
`importable-configuration.nix` is much better! I've renamed it.

Also, I've added a section `Get started` README section which directly points to the main use cases.

jonasnick (approved):
Liking the the README simplification, except that initally I would have preferred that the definition of a node had stayed (in particular since wallet does _not_ imply verification, unfortunately), but nix-bitcoin is more flexible now than "just" being a Bitcoin node.

ACK 2a16db69198d7a1b1143bf64f0480c5677ae5f75

PR #406, opened 2021-10-10
bitcoind: one-option i2p support

nixbitcoin:


erikarvstedt:
[Fixup](https://github.com/erikarvstedt/nix-bitcoin/commits/i2p-ea-2) to fix i2p with netns-isolation.

With the fixup applied, you can test i2p+netns like so:
```bash
run-tests.sh -s "{
  imports = [ scenarios.netnsBase ];
  services.bitcoind.i2p = true;
  services.bitcoind.extraConfig = ''onlynet=i2p'';
}" container --run c bash -c "
  # wait for i2p SAM
  while ! nc -z -w2 169.254.1.10 7656; do sleep 0.1; done
  echo \"adding orionwl's i2p node. this might take up to 60s...\"
  bitcoin-cli addnode h3r6bkn46qxftwja53pxiykntegfyfjqtnzbm6iv6r5mungmqgmq.b32.i2p onetry
  bitcoin-cli getpeerinfo
"
```

PR #407, opened 2021-10-12
lnd: Add TimoutSec

sputn1ck:
Starting my large / old lnd node (6gb channel.db) usually takes around 10 mins.

This PR fixes the service terminating while starting up.


vv01f:
in case somebody wants to use [MicroVM](https://search.nixos.org/flakes?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=microvm) for encapsulation, consider this has a default timeout of 90s and might thus lead to a corrupted block db and eventually in much longer startup time for bitcoind due to revalidation.

PR #416, opened 2021-10-31
secure-node: stop pruning liquidd

### LINKED ISSUE #413: liquid pruned by default in secure-node preset
Is there a reason for this?

I just noticed my liquid node was pruned and wouldn't have guessed that this came from the "secure-node" preset.

https://github.com/fort-nix/nix-bitcoin/blob/c40a7a75cc2d7647324188726dd178d0cb71bf61/modules/presets/secure-node.nix#L40-L45

jonasnick:
There is no security reason why pruning should be enabled and therefore it
surprises users. Turning on pruning in the first place was simply a mistake.

Fixes #413

### DIFF
--- a/modules/presets/secure-node.nix
+++ b/modules/presets/secure-node.nix
@@ -12,6 +12,11 @@ in {
     ./enable-tor.nix
   ];
 
+  options = {
+    # Used by ../versioning.nix
+    nix-bitcoin.secure-node-preset-enabled = {};
+  };
+
   config =  {
     # For backwards compatibility only
     nix-bitcoin.secretsDir = mkDefault "/secrets";
--- a/modules/presets/secure-node.nix
+++ b/modules/presets/secure-node.nix
@@ -44,7 +44,9 @@ in {
     };
 
     services.liquidd = {
-      prune = 1000;
+      # Enable `validatepegin` to verify that a transaction sending BTC into
+      # Liquid exists on Bitcoin. Without it, a malicious liquid federation can
+      # make the node accept a sidechain that is not fully backed.
       validatepegin = true;
       listen = true;
     };
--- a/modules/versioning.nix
+++ b/modules/versioning.nix
@@ -169,6 +169,20 @@ let
             This includes the final database size (0.4) plus some extra storage (0.2).
       '';
     }
+    {
+      version = "0.0.57";
+      condition = config.nix-bitcoin ? secure-node-preset-enabled && config.services.liquidd.enable;
+      message = ''
+        The `secure-node.nix` preset does _not_ set `liquidd.prune = 1000` anymore.
+
+          - If you want to keep the same behavior as before, manually set
+            `services.liquidd.prune = 1000;` in your configuration.nix.
+          - Otherwise, if you want to turn off pruning, you must instruct liquidd
+            to reindex by setting `services.liquidd.extraConfig = "reindex=1";`.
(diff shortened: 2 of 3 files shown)


modules/presets/secure-node.nix
```diff
@@ -39,7 +39,9 @@ in {
     };
 
     services.liquidd = {
-      prune = 1000;
+      # With validatepegin liquidd verifies that a transaction sending BTC into
```
erikarvstedt, line 42: pico-nit, but this is one of the few places where comments are relevant for end-users, so every bit of conciseness helps.
```suggestion
      # Enable `validatepegin` to verify that a transaction sending BTC into
```

modules/versioning.nix
```diff
@@ -169,6 +169,23 @@ let
...
+    {
+      version = "0.0.57";
+      condition = config.services.liquidd.enable && config.services.liquidd.prune == 1000;
+      message = ''
+         With this version, the `secure-node.nix` preset does _not_ set
+         `liquidd.prune=1000` anymore. If you are using the preset (i.e. you
+         import `secure-node.nix` in your configuration.nix), you must take one
+         of the following actions:
+
+           - If you want to keep the same behavior as before, set manually set
```
erikarvstedt, line 181:
```suggestion
           - If you want to keep the same behavior as before, manually set
```

modules/versioning.nix
```diff
@@ -169,6 +169,23 @@ let
...
+             `services.liquidd.prune = 1000;` in your configuration.nix.
+           - Otherwise, if you want to turn off pruning, you must instruct liquidd
+             to reindex by setting `services.liquidd.extraConfig = "-reindex";`.
```
erikarvstedt, line 184:
```suggestion
             to reindex by setting `services.liquidd.extraConfig = "reindex=1";`.
```
`-reindex` only works as a cmdline arg.

modules/versioning.nix
```diff
@@ -169,6 +169,23 @@ let
             This includes the final database size (0.4) plus some extra storage (0.2).
       '';
     }
+    {
+      version = "0.0.57";
+      condition = config.services.liquidd.enable && config.services.liquidd.prune == 1000;
```
erikarvstedt, line 174:
```suggestion
      condition = config.services.liquidd.enable;
```
With the new release, `prune = 1000` is not set anymore, so this condition will be `false` for `secure-node.nix` users.
One could negate the `prune` condition (`config.services.liquidd.prune != 1000`), but I think this is a bit too clever.

jonasnick: Of course... good catch. Should we add a readonly enable option to secure-node.nix to be able to detect this better in the future?

erikarvstedt: Yes, great idea. Can be implemented like this:
```
  # in secure-node.nix
  options = {
    # Used by ../versioning.nix
    nix-bitcoin.secure-node-preset-enabled = {};
  };

  # in versioning.nix
  condition = config.services.liquidd.enable && config.nix-bitcoin ? secure-node-preset-enabled;
```

erikarvstedt:
With the awesome `secure-node-preset-enabled` fix, we can now [greatly simplify the message](https://github.com/erikarvstedt/nix-bitcoin/commits/liquid-prune-ea).

erikarvstedt (approved):
ACK bac8518e7c91d7fb25bd89ac363b6939332aea1a

PR #418, opened 2021-11-02
update nixpkgs

nixbitcoin:
Now that unstable contains a bunch of pkg updates I thought now might also be a good time to update stable.

### DIFF
--- a/modules/bitcoind-rpc-public-whitelist.nix
+++ b/modules/bitcoind-rpc-public-whitelist.nix
@@ -25,7 +25,6 @@
   "gettxoutsetinfo"
   "scantxoutset"
   "verifytxoutproof"
-  "waitfornewblock"
   # Mining
   "getblocktemplate"
   "getmininginfo"
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "flake-utils": {
       "locked": {
-        "lastModified": 1631561581,
-        "narHash": "sha256-3VQMV5zvxaVLvqqUrNz3iJelLw30mIVSfZmAaauM3dA=",
+        "lastModified": 1634851050,
+        "narHash": "sha256-N83GlSGPJJdcqhUxSCS/WwW5pksYf3VP1M13cDRTSVA=",
         "owner": "numtide",
         "repo": "flake-utils",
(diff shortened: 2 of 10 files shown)


erikarvstedt:
[Here are some patches](https://github.com/erikarvstedt/nix-bitcoin/commits/update-nixpkgs-ea) that are partially dependent on this PR. Could you add them here?

nixbitcoin:
Done. Also had to patch spark-wallet because it was failing to build after I cleared out my `/nix/store`.

erikarvstedt (approved):
ACK b3e868d0af9a1b724d8466affbaa4c7d5454084d

modules/bitcoind-rpc-public-whitelist.nix
```diff
@@ -25,7 +25,6 @@
   "gettxoutsetinfo"
   "scantxoutset"
   "verifytxoutproof"
-  "waitfornewblock"
```
jonasnick, line 28: I don't think that the whitelist is minimal for our services right now. If we wanted to do that we could remove quite a bit more. It consists of RPCs that are rather benign and `waitfornewblock` seems to fall in that category. So I'd rather keep it in. If we wanted to improve on this we could have individual whitelists per service.

erikarvstedt: The main reason for removing it is that the old whitelist inspired https://github.com/romanz/electrs/issues/522 and the subsequent fix in electrs.
I'll let you decide.

erikarvstedt: `waitfornewblock` is [not recommnended for public use](https://github.com/bitcoin/bitcoin/pull/10519#issuecomment-323094194). The old whitelist inspired https://github.com/romanz/electrs/issues/522 and the subsequent fix in electrs. I'll let you decide.

jonasnick: There's a difference between public for nix-bitcoin services and public for the rest of the world. But I see that it's generally recommended to not use it. So it's fine by me to remove it (also, we're able to notice lower quality applications trying to use waitfornewblock).

jonasnick (commented):
utACK b3e868d0af9a1b724d8466affbaa4c7d5454084d

PR #420, opened 2021-11-03
missing sys import for error path set_onion_address

radixrat:
- fix missing import utilized here: https://github.com/fort-nix/nix-bitcoin/blob/master/modules/nodeinfo.nix#L71


erikarvstedt (approved):
ACK ae1c90997d15dff25ddf23345bd73121d6cc7673

Thank you!

PR #425, opened 2021-11-24
Misc. improvements

erikarvstedt:
Commit `services: add tor.* options` has no effect on the test derivations.
It unlocks section `Allow bitcoind RPC connections from LAN` in the docs.


modules/joinmarket-ob-watcher.nix
```diff
@@ -100,7 +95,7 @@ in {
         SystemCallFilter = nbLib.defaultHardening.SystemCallFilter ++ [ "mbind" ] ;
         Restart = "on-failure";
         RestartSec = "10s";
-      } // nbLib.allowTor;
+      } // nbLib.allowAllIPAddresses;
```
jonasnick, line 98: > because it doesn't make sense to lock down these Tor-only services.

Why not? Locking down is good, even if the service promises to be Tor-only.

modules/bitcoind.nix
```diff
@@ -407,7 +413,7 @@ in {
         Restart = "on-failure";
         UMask = mkIf cfg.dataDirReadableByGroup "0027";
         ReadWritePaths = cfg.dataDir;
-      } // nbLib.allowedIPAddresses cfg.enforceTor
+      } // nbLib.allowedIPAddresses (cfg.tor.proxy && cfg.tor.enforce)
```
jonasnick, line 416: This seems to contradict the description of tor.enforc which would suggest that setting tor.enforce alone is enough to enforce tor. Also should we set the default value of `cfg.tor.proxy` to cfg.tor.enforce to make sure that outbound bitcoind options (for example) happen if Tor is enforced?

Instead of having 2 options `enforce` and `proxy` we could alternatively have a single option `tor` with the values "enforce", "out-only", "false"?

jonasnick (commented):
The cookbook is nice! Contains a lot of useful stuff. How do we distinguish it from usage.md? It seems like the parts on options, debugging and secrets would belong in usage.md.

modules/onion-addresses.nix
```diff
@@ -58,8 +58,20 @@ in {
         CapabilityBoundingSet = "CAP_CHOWN CAP_FSETID CAP_SETFCAP CAP_DAC_OVERRIDE CAP_DAC_READ_SEARCH CAP_FOWNER CAP_IPC_OWNER";
       };
       script = ''
+        waitForFile() {
+          file=$1
+          for ((i=0; i<50; i++)); do
```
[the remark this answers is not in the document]
erikarvstedt: Fixed.

modules/rtl.nix
```diff
@@ -186,6 +202,8 @@ in {
       "d '${cfg.dataDir}' 0770 ${cfg.user} ${cfg.group} - -"
     ];
 
+    services.rtl.tor.enforce = mkIf (cfg.extraCurrency != null) false;
```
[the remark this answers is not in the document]
erikarvstedt: Fixed.

docs/cookbook.md
```diff
@@ -0,0 +1,193 @@
+# Customizing your configuration
+
+## Get started with Nix
+
+See [nix.dev](https://nix.dev/) for a pragmatic introduction to Nix and NixOS.
```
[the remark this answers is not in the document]
erikarvstedt: Fixed in `fixup! cookbook: improve links to external docs`.

modules/joinmarket-ob-watcher.nix
erikarvstedt: Fixed.

modules/bitcoind.nix
erikarvstedt: > This seems to contradict the description of tor.enforc which would suggest that setting tor.enforce alone is enough to enforce tor.

Yes, this is an important point. Fixed in `fixup! tor: don't disable tor enforcement when tor.proxy is disabled`.

> Also should we set the default value of cfg.tor.proxy to cfg.tor.enforce to make sure that outbound bitcoind options (for example) happen if Tor is enforced?

I think it's simpler and clearer if both options are kept orthogonal.

> we could alternatively have a single option tor with the values "enforce", "out-only", "false"?

This is complicated by the fact that some services like `electrs` or `btcpayserver` don't support Tor proxying.
Also, setting `bitcoind.tor = false` looks a bit as this would also disable the hidden service.

erikarvstedt:
> How do we distinguish it from usage.md? It seems like the parts on options, debugging and secrets would belong in usage.md.

I've now restructured the docs in this way:
- Add section `Managing your configuration` to `Cookbook`, which includes section `Updating nix-bitcoin` from `Usage`.
- The sole topic of `Usage` is now how to enable and interact with node services. Thus rename `Usage` -> `Services` (Long title: `Using services`).
- Rename `Cookbook` to `Configuration` (Long title: `Configuration and Maintenance`).
  Put `Configuration and Maintenance` before `Using services` in the README.

modules/bitcoind.nix
jonasnick: > I think it's simpler and clearer if both options are kept orthogonal.

You're right.

jonasnick (commented):
Fixups look good!

erikarvstedt:
Squashed all fixups.

jonasnick (commented):
ACK b68eedc971ffa8c3b7336f0c91d1c97ef6aaca60

docs/configuration.md
```diff
@@ -0,0 +1,228 @@
...
+# 3. Fix data dir permissions on the nix-bitcoin node
+ssh root@nix-bitcoin-node 'chown -R bitcoin: /var/lib/bitcoind'
+
+# 4. Start bitcoind
+ssh root@nix-bitcoin-node 'systemctl start bitcoind'
+```
+
+You can use the same workflow for other services.\
+The default data dir path is `/var/lib/<service>` for all services.\
+Some services require extra steps:
```
nixbitcoin, line 156: JoinMarket also requires copying the wallet password

docs/configuration.md
```diff
@@ -0,0 +1,228 @@
...
+
+  Copy the postgresql database:
+  ```shell
+  # Export (on the other node)
+  sudo -u postgres pg_dump YOUR_BTCPAYSERVER_DB > export.sql
+  # Restore (on the nix-bitcoin node)
+  sudo -u postgres psql btcpaydb < export.sql
+  ```
+
+# Use bitcoind from another node
```
nixbitcoin, line 171: This is especially useful in combination with WireGuard. I'll add that to `configuration.md` later on.

modules/presets/enable-tor.nix
```diff
@@ -1,27 +1,42 @@
...
+    # Use Tor as a proxy for outgoing connections
+    # and restrict all connections to Tor
+    #
+    bitcoind = defaultEnableTorProxy;
+    clightning = defaultEnableTorProxy;
+    lnd = defaultEnableTorProxy;
+    lightning-loop = defaultEnableTorProxy;
+    liquidd = defaultEnableTorProxy;
     # disable Tor enforcement until btcpayserver can fetch rates over Tor
-    # btcpayserver.enforceTor = true;
```
nixbitcoin, line 19: This should also be a `TODO-EXTERNAL`.

docs/configuration.md
erikarvstedt: Fixed.

modules/presets/enable-tor.nix
erikarvstedt: Great, thanks. Fixed.

docs/configuration.md
```diff
@@ -0,0 +1,235 @@
...
+  sudo -u postgres pg_dump YOUR_BTCPAYSERVER_DB > export.sql
+  # Restore (on the nix-bitcoin node)
+  sudo -u postgres psql btcpaydb < export.sql
+  ```
+
+- joinmarket
+
+  Copy your wallet to `/var/lib/joinmarket/wallets/wallet.jmdat`.\
+  Write your wallet password, without a trailing newline, to
+  `$secretsDir/lnd-wallet-password` (See: [Secrets dir](#secrets-dir)).
```
nixbitcoin, line 176: Typo

docs/configuration.md
```diff
@@ -0,0 +1,235 @@
...
+```
+This way, the systemd service still exists, but is not automatically started.
+
+# Appendix
+
+## Secrets dir
+
+The secrets dir is set by option `nix-bitoin.secretsDir` and has the
+following default values:
+- `/secrets` (if you're using the `secure-node.nix` template)
```
nixbitcoin, line 231: Aren't passwords kept in `/var/src/secrets`?

erikarvstedt: Ah, right, when krops is used. Fixed.

docs/configuration.md
erikarvstedt: Fixed

nixbitcoin (approved):
```
Hash: SHA512

ACK def64a73b8d87a1c6cb261508b53c9477d3b049f

Running on nixbitcoin.org
-----BEGIN PGP SIGNATURE-----

iQIzBAEBCgAdFiEEiDOlIkszv4ojhh+OtgROy6La5dAFAmGk03gACgkQtgROy6La
5dBDFQ//eMjdpDRlCZNh1mow38LEoFAqg6r360eQs2R0eqgbq27tMYoqFq0jVzbH
Fr4I+XHgY6j8ItplkoairQ/vTdtwMaqJw41kiw3qtF7H7cU3W5WTcvNf65aufWsF
CKgmewP47yuOYqbnYXpxAvzd+UFOuq6ZrZXjbXJH0hBXHejqO4Hx72cZFjV0U5Dl
kGeZ8wnwabkf9IafE/ObJPuZqmmlno/3rov17DzYhAE9jtJBEqDhTqao+fjWCPWU
+JqoiwhBZOOH7v53rtkZInpNUKu1v1cvw1yqiLD6Ndoclo5XOUpHQLCUH+SIFLmH
mAOXPFJaBFxDXjesoSFFW5/Nve032JrWBchmTW6wv3HOon8MMcP20H9Q/32w+/UV
IfzkoIZZMfW0Nchnv4kEfaSp/U/bmhx5gOzZ3p04J9qh8EgtRe5vgbbiAmLLtJbM
3ilU9sYefTcfv85XI41G0Mcwkkpc4WxlS8eGU0gmPFP4jpJMkqnxuNjGb4dz6Bii
yzdieFKH6PpzPiwSCoPN8uUyuMids9qa/jwNHEsgHneeWmQ62YTR1M2Z9ufcNs6y
FLWd2/SzG/9xjfh62Ne3gIMwYvhRcD6+ySGVXf00ZYScwwz37QRzrTTa/vtHtQkR
vhgpNX+0qSNT+woU6iEvdG5RIV2HbTt+zUrR9IrDfW4km8mEqcw=
=xtXf
-----END PGP SIGNATURE-----
```

jonasnick (commented):
ACK def64a73b8d87a1c6cb261508b53c9477d3b049f

PR #435, opened 2022-01-04
run-tests: fix `vm` cmd

erikarvstedt:
#### Copy of commit msg
Previously, `qemu` failed with error `restrict=on : Parameter 'restrict' expects 'on' or 'off'` due to the trailing space after `on`.


jonasnick (approved):
ACK 98bc0917d7d659ad75a89b17599fcf7ffeb19487

PR #415, opened 2021-10-29
WIP: Add clightning backup plugin

jonasnick:

PR #436, opened 2022-01-06
clightning: native database replication

nixbitcoin:
Not ready for merge until clightning 0.10.3.

Supersedes https://github.com/fort-nix/nix-bitcoin/pull/415

Coauthor: @erikarvstedt

### DIFF
--- a/docs/services.md
+++ b/docs/services.md
@@ -26,6 +26,104 @@ systemctl cat bitcoind
 systemctl show bitcoind
 ```
 
+# clightning database replication
+
+The clightning database can be replicated to a local path
+or to a remote SSH target.\
+When remote replication is enabled, nix-bitcoin mounts a SSHFS to a local path.\
+Optionally, backups can be encrypted via `gocryptfs`.
+
+Note: You should also backup the static file `hsm_secret` (located at
+`/var/lib/clightning/bitcoin/hsm_secret` by default), either manually
+or via the `services.backups` module.
+
+## Remote target via SSHFS
+
+1. Add this to your `configuration.nix`:
+   ```nix
+   services.clightning.replication = {
+     enable = true;
+     sshfs.destination = "user@hostname:directory";
+     # This is optional
+     encrypt = true;
+   };
+   programs.ssh.knownHosts."hostname".publicKey = "<ssh public key from running `ssh-keyscan` on the host>";
+   ```
+
+   Leave out the `encrypt` line if you want to store data on your destination
+   in plaintext.\
+   Adjust `user`, `hostname` and `directory` as necessary.
+
+2. Deploy
+
+3. To allow SSH access from the nix-bitcoin node to the target node, either
+   use the remote node config below, or copy the contents of `$secretsDir/clightning-replication-ssh.pub`
+   to the `authorized_keys` file of `user` (or use `ssh-copy-id`).
+
+4. You can restrict the nix-bitcoin node's capabilities on the SSHFS target
+   using OpenSSH's builtin features, as detailed
+   [here](https://serverfault.com/questions/354615/allow-sftp-but-disallow-ssh).
+
+   To implement this on NixOS, add the following to the NixOS configuration of
+   the SSHFS target node:
+   ```nix
+   systemd.tmpfiles.rules = [
+     # Because this directory is chrooted by sshd, it must only be writable by user/group root
+     "d /var/backup/nb-replication 0755 root root - -"
+     "d /var/backup/nb-replication/writable 0700 nb-replication - - -"
+   ];
+
+   services.openssh = {
+     extraConfig = ''
+       Match user nb-replication
+         ChrootDirectory /var/backup/nb-replication
+         AllowTcpForwarding no
+         AllowAgentForwarding no
+         ForceCommand internal-sftp
+         PasswordAuthentication no
+         X11Forwarding no
+     '';
+   };
+
+   users.users.nb-replication = {
+     isSystemUser = true;
+     group = "nb-replication";
+     shell = "${pkgs.coreutils}/bin/false";
+     openssh.authorizedKeys.keys = [ "<contents of $secretsDir/clightning-replication-ssh.pub>" ];
+   };
+   users.groups.nb-replication = {};
+   ```
+
+   With this setup, the corresponding `sshfs.destination` on the nix-bitcoin
+   node is `"nb-replication@hostname:writable"`.
+
+## Local directory target
+
+1. Add this to your `configuration.nix`
+   ```nix
+   services.clightning.replication = {
+     enable = true;
+     local.directory = "/var/backup/clightning";
+     encrypt = true;
+   };
+   ```
+
+   Leave out the `encrypt` line if you want to store data in
+   `local.directory` in plaintext.
+
+2. Deploy
+
+clightning will now replicate database files to `local.directory`. This
+can be used to replicate to an external HDD by mounting it at path
+`local.directory`.
+
+## Custom remote destination
+
+Follow the steps in section "Local directory target" above and mount a custom remote
+destination (e.g., a NFS or SMB share) to `local.directory`.\
+You might want to disable `local.setupDirectory` in order to create the mount directory
+yourself with custom permissions.
+
 # Connect to RTL
 Normally you would connect to RTL via SSH tunneling with a command like this
 
@@ -221,7 +319,6 @@ lndconnect-onion --host=mynode.org
 
 5. Edit your deployment tool's configuration and change the node's address to `localhost` and the ssh port to `<random port of your choosing>`.
    If you use krops as described in the [installation tutorial](./install.md), set `target = "localhost:<random port of your choosing>";` in `krops/deploy.nix`.
-
 
 6. After deploying the new configuration, it will connect through the SSH tunnel you established in step iv. This also allows you to do more complex SSH setups that some deployment tools don't support. An example would be authenticating with [Trezor's SSH agent](https://github.com/romanz/trezor-agent), which provides extra security.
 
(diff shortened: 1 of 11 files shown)


modules/clightning-db-replication.nix
```diff
@@ -0,0 +1,223 @@
...
+          "''${gocryptfs[@]}" "$cipherDir" ${plaintextDir}
+        '';
+
+      preStop =
+        optionalString cfg.encrypt ''
+          umount ${plaintextDir} || true
+        '' +
+        optionalString useSshfs ''
+          umount ${sshfsDir}
+        '';
```
jonasnick, line 193: What if the service aborts in a way that the preStop isn't executed? Should we try unmounting at service start?

jonasnick (commented):
Very cool stuff!

Ideally we would also document how to restrict the capabilities of the nix-bitcoin node on an SSHFS target as much as possible. It should only be able to access the directory and not be able to access or execute anything else.

modules/clightning-db-replication.nix
```diff
@@ -0,0 +1,223 @@
...
+
+  config = mkIf cfg.enable {
+    assertions = [
+      { assertion = useSshfs || (localDir != null);
+        message = ''
+          services.clightning.replication: One of `sshfs.destination` or
+          `local.directory` must be set.
+        '';
+      }
+      { assertion = !useSshfs || (localDir == null);
```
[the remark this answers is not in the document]
nixbitcoin: I think this would actually make things more complex, because not only do you need to state and sshfs destination and port, but also directory and ssh options. For local directory it is also relevant if you want to have the service set up the directory automatically or if you want to set it up manually. The two are separate methods, and I think we should keep them as such.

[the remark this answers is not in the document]
erikarvstedt: The correct type would be `oneOf submodule`. Compared to using assertions this is more complex, less readable and unexportable by `optionAttrSetToDocList` which is used when generating option docs.

modules/clightning-db-replication.nix
```diff
@@ -0,0 +1,223 @@
...
+          services.clightning.replication: Only one of `sshfs.destination` and
+          `local.directory` must be set.
+        '';
+      }
+    ];
+
+    systemd.tmpfiles.rules = optional (localDir != null && cfg.local.setupDirectory)
+      "d '${localDir}' 0770 ${user} ${group} - -";
+
+    services.clightning.extraConfig = let
```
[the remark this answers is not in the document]
nixbitcoin: Fixed in ab947857d08b41b96c470724f8f309c368a1fd64

docs/services.md
```diff
@@ -26,6 +26,58 @@ systemctl cat bitcoind
 systemctl show bitcoind
 ```
 
+# Setup clightning database replication
+
```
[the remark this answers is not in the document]
nixbitcoin: Fixed in d91cfb6272518d52d2bb72b9d0a2b9cbc008c05e

docs/services.md
```diff
@@ -26,6 +26,58 @@ systemctl cat bitcoind
...
+1. Append to your `configuration.nix`
+
+    ```
+    services.clightning.replication = {
+      enable = true;
+      sshfs.destination = "user@hostname:directory";
+      encrypt = true;
+    };
+    programs.ssh.knownHosts."hostname".publicKey = "<ssh public key from `ssh-keyscan`>";
```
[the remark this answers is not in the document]
nixbitcoin: Fixed in ad816293f7c14a0104a5f891aff58a9786515a6d

modules/clightning-db-replication.nix
nixbitcoin: You mean (re-)unmounting so that the service will mount the directory fresh every time?

erikarvstedt: > aborts in a way that the preStop isn't executed

We should use `postStop` instead. Barring a systemd bug or a crash of the systemd process (both very improbable), `postStop` is guaranteed to be executed after the service was started.
`preStop` wouldn't be called if `ExecStart` mounts a SSHFs but fails setting up the cryptfs, thus leaking a SSHFs mount after service stop.

nixbitcoin: Implemented in ab9a38b4bd99aca00d1274eda2ea05f1dee41f2e

nixbitcoin:
> Ideally we would also document how to restrict the capabilities of the nix-bitcoin node on an SSHFS target as much as possible. It should only be able to access the directory and not be able to access or execute anything else.

We could to this using `Subsystem sftp internal-sftp` in `OpenSSH` as detailed [here](https://serverfault.com/questions/354615/allow-sftp-but-disallow-ssh).

modules/clightning-db-replication.nix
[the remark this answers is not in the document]
jonasnick: Okay, sad.

modules/clightning-db-replication.nix
```diff
@@ -0,0 +1,221 @@
...
+with lib;
+let
+  options.services.clightning.replication = {
+    enable =  mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Enable live replication of the clightning database.
+        This ensures that funds are safe at all times by immediately backing up
+        each channel update.
```
jonasnick, line 12: Ensuring that funds are safe is hyperbole if only for a reason that you can't roll back. Perhaps "This prevents losing off-chain funds when the primary wallet file becomes inaccessible."

nixbitcoin: Fixed in 8dd8bd252f08948bd163d5b9bd250189781a00a8

docs/services.md
```diff
@@ -26,6 +26,96 @@ systemctl cat bitcoind
...
+          ChrootDirectory /var/backup/%u
+          X11Forwarding no
+          AllowTcpForwarding no
+          AllowAgentForwarding no
+          ForceCommand internal-sftp
+      '';
+    };
+
+    users.groups.sftponly = {};
+    users.users.<user>.extraGroups = [ "sftponly" ];
```
jonasnick, line 88: The stackoverflow post additionally mentions setting the users shell to `/usr/bin/false`.

nixbitcoin:
On 24.01.2022 13:05, Jonas Nick wrote:
>The stackoverflow post additionally mentions setting the users shell to `/usr/bin/false`.

Doesn't work on NixOS, I tried.

--
nixbitcoindev [DD11 F9AD 5308 B3BA]
Use PGP

prusnak:
> Not ready for merge until clightning 0.10.3.

clightning 0.11.0.1 was merged to nixpkgs in https://github.com/NixOS/nixpkgs/pull/170584

modules/clightning-db-replication.nix
```diff
@@ -0,0 +1,221 @@
...
+    systemd.tmpfiles.rules = optional (localDir != null && cfg.local.setupDirectory)
+      "d '${localDir}' 0770 ${user} ${group} - -";
+
+    services.clightning.wallet = let
+      mainDB = "${clightning.dataDir}/${network}/lightningd.sqlite3";
+      replicaDB = "${destDir}/lightningd.sqlite3";
+    in "sqlite3://${mainDB}:${replicaDB}";
+
+    systemd.services.clightning = {
+      bindsTo = mkIf useMounts [ "clightning-replication-mounts.server" ];
```
erikarvstedt, line 139:
```suggestion
      bindsTo = mkIf useMounts [ "clightning-replication-mounts.service" ];
```

nixbitcoin:
Running backup server in my local network over sshfs with no issues. Haven't tried across the internet for fear of latency issues. Ready for testing/merge IMO.

jonasnick (commented):
utACK 2203031ee8576814de79f62b011349a449dc55d3 mod [fixups](https://github.com/jonasnick/nix-bitcoin/commits/clightning-replication-nb-jn)

nixbitcoin:
Fixups merged and rebased

elsirion (approved):
Looking forward to using it :smiley:

docs/services.md
```diff
@@ -26,6 +26,103 @@ systemctl cat bitcoind
...
+    users.groups.sftponly = {};
+    users.users.<user> = {
+      shell = "${pkgs.coreutils}/bin/false";
+      extraGroups = [ "sftponly" ];
+    };
+    ```
+
+    If the `<user>` is set to `nb-replication` and the targets hostname is
+    `hostname`, then the corresponding `sshfs.destination` on the nix-bitcoin
+    node is `"nb-replication@hostname:/var/backup/nb-replication/writeable"`.
```
elsirion, line 96: I'm recklessly running e086fc567baf394eb5d5bf873eec9f614f36c6b2 in production now and one thing I noticed is that due to `ChrootDirectory /var/backup/%u` this path should be `/writable` instead of `/var/backup/nb-replication/writeable`.

nixbitcoin:
Rebased to master

nixbitcoin:
Added a test for sshfs client-server replication

erikarvstedt:
[Fixups](https://github.com/erikarvstedt/nix-bitcoin/commits/clightning-backup-nb-ea).

nixbitcoin:
Thanks for the phenomenal review. Let's get this merged @jonasnick!

erikarvstedt (approved):

erikarvstedt (changes requested):
[Fixup](https://github.com/erikarvstedt/nix-bitcoin/commit/e335543ecb913b87c0bffcb2b34bea0cd59277cb).

erikarvstedt (approved):
ACK 5eb01498d503d3325bdceeb498c947a82e7cf96c

jonasnick (approved):
Wow, our first test network. Thanks @erikarvstedt!

ACK 5eb01498d503d3325bdceeb498c947a82e7cf96c

Issue #439, opened 2022-01-11
joinmarket-yieldgenerator _DEFAULT_INTEREST_RATE error


alumbigi:
After I added a fidelity bond, joinmarket-yieldgenerator throws an error on startup that looks like this:

```
Unhandled error in Deferred:
Traceback (most recent call last):
  File "/nix/store/.../python3.9/site-packages/twisted/internet/base.py", line 1318, in run
    self.mainLoop()
  File "/nix/store/.../python3.9/site-packages/twisted/internet/base.py", line 1328, in mainLoop
    reactorBaseSelf.runUntilCurrent()
  File "/nix/store/.../python3.9/site-packages/twisted/internet/base.py", line 994, in runUntilCurrent
    call.func(*call.args, **call.kw)
  File "/nix/store/.../python3.9/site-packages/twisted/internet/task.py", line 253, in __call__
    d = maybeDeferred(self.f, *self.a, **self.kw)
--- <exception caught here> ---
  File "/nix/store/.../python3.9/site-packages/twisted/internet/defer.py", line 191, in maybeDeferred
    result = f(*args, **kwargs)
  File "/nix/store/.../python3.9/site-packages/jmclient/maker.py", line 48, in try_to_create_my_orders
    self.fidelity_bond = self.get_fidelity_bond_template()
  File "/nix/store/.../python3.9/site-packages/jmclient/yieldgenerator.py", line 154, in get_fidelity_>
    interest_rate = get_interest_rate()
  File "/nix/store/.../python3.9/site-packages/jmclient/configure.py", line 587, in get_interest_rate
    return float(global_singleton.config.get('POLICY', 'interest_rate',
builtins.ValueError: could not convert string to float: '""" + _DEFAULT_INTEREST_RATE + """'
```
to fix it, I edited the file `/var/lib/joinmarket/joinmarket.cfg`
replaced `“”” + _DEFAULT_INTEREST_RATE + “””` with `0.015`
and restarted joinmarket-yieldgenerator

Notes:
 - if I restart joinmarket, my edit get reversed
 - NixOS 21.11, nix-bitcoin 0.0.60 and 0.0.61
 - I am trying to learn how to code and test a change that adds the ability to set something like services.joinmarket.interestrate in configuration.nix. I am not there yet
 - thanks

nixbitcoin:
[...]

PR #440, opened 2022-01-11
joinmarket: fix fidelity bond default values

nixbitcoin:
Fixes https://github.com/fort-nix/nix-bitcoin/issues/439

Does this solve your issue @alumbigi?

### DIFF
--- a/modules/joinmarket.nix
+++ b/modules/joinmarket.nix
@@ -201,8 +201,8 @@ let
     tx_broadcast = self
     minimum_makers = 4
     max_sats_freeze_reuse = -1
-    interest_rate = """ + _DEFAULT_INTEREST_RATE + """
-    bondless_makers_allowance = """ + _DEFAULT_BONDLESS_MAKERS_ALLOWANCE + """
+    interest_rate = 0.015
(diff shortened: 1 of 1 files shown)


alumbigi:
I _think_ it does fix it since I made the same corresponding manually inside the deployment and it works.

However I don't know what the right way to test it is.
Should I:
 - edit my `nix-bitcoin-release.nix`
 - replace the 'builtins.fetchTarball' section with a `builtins.fetchGit` section that points to this commit 
 - run nix-shell / deploy ?
Or am I off track? Thank you

jonasnick (approved):
utACK 40c5f63f04105b8597c022121cca94d7a351fa2a

@alumbigi One way to test things is to overwrite `nix-bitcoin-release.nix` with a path to a local checkout of the nix-bitcoin repository. For example
```
$ cat nix-bitcoin-release.nix
../nix-bitcoin
```

However, it's not really advisable to test PRs with your production node if it has funds or important data on it. A better way would be to recreate the conditions that lead to your bug in one of the examples and then see if it's fixed with the PR.

alumbigi:
@jonasnick thank you

PR #448, opened 2022-02-03
[LND] Remove pruning assertion

sputn1ck:
As of `lnd v0.13.0-beta` lnd started to support pruned bitcoind nodes.

PR: https://github.com/lightningnetwork/lnd/pull/5154

This PR removes the assertion that checks for a pruned bitcoind node.


prusnak (approved):
ACK 793a127372ec50131f17cf3e14adee11f7f87148

PR #450, opened 2022-02-03
Misc. improvements

erikarvstedt:


helper/fetch-release
```diff
@@ -14,15 +14,15 @@ fi
 TMPDIR=$(mktemp -d)
 trap "rm -rf $TMPDIR" EXIT
 
-GPG_HOME=$TMPDIR/gpg-home
+export GPG_HOME=$TMPDIR/gpg-home
```
jonasnick, line 17: isn't this variable called `GNUPGHOME`?

modules/bitcoind.nix
```diff
@@ -386,7 +388,12 @@ in {
           ''
         ) (builtins.attrNames cfg.rpc.users);
       in ''
-        ${optionalString cfg.dataDirReadableByGroup "chmod -R g+rX '${cfg.dataDir}/blocks'"}
+        ${optionalString cfg.dataDirReadableByGroup ''
+          if [[ -e '${cfg.dataDir}/blocks' ]]; then
+            chmod -R g+rX '${cfg.dataDir}/blocks'
```
jonasnick, line 393: If this bitcoind service starts for the first time, the blocks dir does not exist yet, so `cfg.dataDirReadableByGroup` won't have the desired effect?

modules/versioning.nix
```diff
@@ -181,6 +181,27 @@ let
...
+                  config.nix-bitcoin.secretsDir == "/etc/nix-bitcoin-secrets";
+      message = ''
+        The `secure-node.nix` preset does not set the secrets directory
+        to "/secrets" anymore.
+        Instead, the default location "/etc/nix-bitcoin-secrets" is used.
+
+        To upgrade, choose one of the following:
+
+        - Continue using "/secrets":
+          Add `nix-bitcoin.secretsDir = "/secrets"` to your configuration.nix.
```
jonasnick, line 196: nit: `s/secrets"/secrets";`

jonasnick (commented):
Concept ACK

The flake changes look benign but I don't fully understand what is happening there.

helper/fetch-release
erikarvstedt: Aww, not my proudest moment. Thanks for catching this.
I quickly drafted that commit a while a ago and then just cherry-picked it without proper checking.

modules/bitcoind.nix
erikarvstedt: With `dataDirReadableByGroup = true`, the blocks dir is automatically created and populated with group read permissions.

See:
```bash
# dataDirReadableByGroup = false
run-tests.sh -s bitcoind container --run c ls -al /var/lib/bitcoind
# Result
drwx------ 3 bitcoin bitcoin 4096 Feb  5 21:45 blocks

# dataDirReadableByGroup = true
run-tests.sh -s "{
  services.bitcoind.enable = true;
  services.bitcoind.dataDirReadableByGroup = true;
}" container --run c ls -al /var/lib/bitcoind
# Result
drwxr-x--- 3 bitcoin bitcoin 4096 Feb  5 21:46 blocks
```

`chmod -R g+rX '${cfg.dataDir}/blocks'` only exists so that users can enable `dataDirReadableByGroup` on an existing bitcoind instance that was created without `dataDirReadableByGroup`.

erikarvstedt:
> The flake changes look benign but I don't fully understand what is happening there.

I guess you're referring to commit `flake: optimize nixpkgs importing`.
[Here's an example flake](https://gist.github.com/661536b4fb160918f1e8d8245baee80c) that explains it.

helper/fetch-release
jonasnick: I only noticed this because the commit message said that we would follow the same approach with the fetcher scripts, but wasn't able to find GPG_HOME.

modules/bitcoind.nix
jonasnick: Ah sorry, I missed the umask setting.

jonasnick (commented):
ACK d959d5b5581a02b6faef30f876489ca1cf89c564

jonasnick:
Thanks for the example flake. That was helpful.

jonasnick (approved):

PR #452, opened 2022-02-12
makeShell: use final system configuration in generate-secrets

erikarvstedt:
#### Copy of commit msg
This fixes a bug where the version update message for v0.0.65 is erroneously triggered because the krops config is not included when evaluating secrets.


jonasnick (approved):
ACK 1af84e8b7e47b54126d23078e4d381dcffde2883

Issue #409, opened 2021-10-22
Backups while the services are runnnig


seberm:
Hello,
maybe I am just missing something, but do the backups using the `backups` service make sense? Have you ever tried to recover some of the services from these duplicity backups?

My point is that using the `backups` service I can copy all of the important data dirs. But are the services (like clightning, bitcoind, electrs, ...) stopped during the backup?

It seems that the `cp`ing files while the services are running can easily lead to a corrupted backup files that cannot be recovered from.

For example, what is the current recommended way to backup the c-lightning's channel states? The channels states are saved in `/var/lib/clightning/bitcoin/lightningd.sqlite3` database. Without it I am not able to recover the **in-channel** funds. It does not make sense to backup this file while clightning is running:
- https://lightning.readthedocs.io/BACKUP.html#backing-up-while-lightningd-is-running

Another example is running a backup with an `services.backups.with-bulk-data` option enabled. Does it make sense to backup `chainstate/*` or `blocks/index/*` files while `bitcoind` is still running? I am almost sure that the levelDB files in the backup will be corrupted.

What are your thoughts on this? Shouldn't we stop these services during the backup or at least allow the user to parametrize this behaviour?

Please, this is not a criticism, I really appreciate your work!

Thanks,
Ota

PR #454, opened 2022-02-14
lnd: point out SCB functionality more clearly

nixbitcoin:
I think it is not clearly apparent to new users that we support LND Static Channel Backups out of the box. Especially in conjunction with our `backups` module, this provides a fairly basic recovery strategy for LND.

At least a partial answer to https://github.com/fort-nix/nix-bitcoin/issues/409


jonasnick (approved):
ACK b8524720a5d652922173990050e2cf728cceb7d1

PR #462, opened 2022-03-08
Add PeerSwap

sputn1ck:
This PR adds [peerswap](https://github.com/sputn1ck/peerswap). Peerswap allows you to rebalance channels with your direct peers using submarine swaps on bitcoin and on liquid. It is alpha software so should be used with caution.

I added modules for lnd as well as the clightning plugin.

Let me know if the testcase makes sense. I don't know how to run clightning plugin tests, however running
`./run-tests.sh -s "{ services.clightning.enable = true;services.clightning.plugins.peerswap.enable = true;}" container --run c "lightning-cli peerswap-listswaps"` went fine

### DIFF
--- a/pkgs/peerswap/default.nix
+++ b/pkgs/peerswap/default.nix
@@ -1,0 +1,26 @@
+{ pkgs, lib, fetchFromGitHub, buildGoModule, goSubPackages}:
+
+buildGoModule rec {
+  pname = "peerswap-lnd";
+  version = "0.2.0-beta";
+
+  src = fetchFromGitHub {
+      repo = "peerswap";
+      owner = "sputn1ck";
+      rev = "d37d4be0be7899e48694068f03c1b02fde1734a0";
+      sha256 = "sha256:1axmxrk3svrzpg18sp2c21x7bqakhk3hszz1smpy6pc1x3s7hfsh";
+  };
+
+  subPackages = goSubPackages;
+  vendorSha256 = "sha256:13jgk2r8ac2vxrs10xjhilp0x1qvjjgs6knkhf67bmsbx3n9abnl";
+  proxyVendor = true;
+
+  meta = with lib; {
+    description = "P2P lightning channel rebalancing using lightning<->on-chain atomic swaps";
+    homepage = "peerswap.dev";
+    maintainers = with maintainers; [ sputn1ck ];
+    license = licenses.mit;
+    platforms = platforms.linux;
+
+  };
+}
--- a/modules/clightning-plugins/default.nix
+++ b/modules/clightning-plugins/default.nix
@@ -17,6 +17,7 @@ in {
     ./prometheus.nix
     ./summary.nix
     ./zmq.nix
+    ./peerswap-cln.nix
   ];
 
   inherit options;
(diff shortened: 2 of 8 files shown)


erikarvstedt:
[Fixups](https://github.com/erikarvstedt/nix-bitcoin/commits/peerswap-ea/).

- `peerswap-liquid-rpcwallet` should probably be renamed to `peerswap-liquid-wallet`.
- `allowlisted_peers` should be renamed to `allowed_peers`.
- Currently, running `peerswap-lnd` and the clightning plugin in parallel fails with:
  ```
  -4:Wallet file verification failed. Failed to create database path '/var/lib/liquidd/liquidv1/wallets/peerswap'. Database already exists.
  ```
  Should we assign different wallet names to each implementation?
- Why does `peerswap-lnd` require `lnd` group access?
- When running the clightning plugin with liquid support it exits right after starting. You can check this with:
  ```
  run-tests.sh -s '{
    services.liquidd.enable = true;
    services.clightning.plugins.peerswap.enable = true;
  }' container --run c journalctl -u clightning
  ```
  Output:
  ```
  lightningd[490]: INFO    plugin-peerswap: Liquid swaps enabled
  lightningd[490]: INFO    plugin-peerswap: Killing plugin: exited during normal operation
  ```
  
  You can also run a VM and check the logs (`journalctl -u clightning`):
  ```
  run-tests.sh -s '{
    services.liquidd.enable = true;
    services.clightning.plugins.peerswap.enable = true;
  }' vm
  ```

- Does it make sense to enable liquid peerswaps by default when liquid is enabled?
  Could this be too intrusive, given that a wallet is created? Or do you expect that most liquid users want to enable liquid peerswap support?

jonasnick:
concept ACK for adding peerswap

> Fixups

Randomly noticed that peerswap isn't mentioned in the README.

> Does it make sense to enable liquid peerswaps by default when liquid is enabled?

A downside with peerswap is that it has full access to all liquidd wallets, which is
something that we should mention somewhere and allow the user to explicitly opt-in to (similarly to how we note this for joinmarket in the example config).

jonasnick:
By the way, would be good to get rid of the "DO NOT USE ON MAINNET YET" warning in the peerswap README before we merge this.

nixbitcoin:
concept ACK for adding peerswap

> Should we assign different wallet names to each implementation?

Yes

> Does it make sense to enable liquid peerswaps by default when liquid is enabled?
Could this be too intrusive, given that a wallet is created? Or do you expect that most liquid users want to enable liquid peerswap support?

Too intrusive in my opinion, especially because this is alpha software and it requires full access to all liquidd wallets, as @jonasnick mentioned above.

pkgs/peerswap/default.nix
```diff
@@ -0,0 +1,26 @@
+{ pkgs, lib, fetchFromGitHub, buildGoModule, goSubPackages}:
+
+buildGoModule rec {
+  pname = "peerswap-lnd";
+  version = "0.2.0-beta";
+
+  src = fetchFromGitHub {
+      repo = "peerswap";
+      owner = "sputn1ck";
```
northben, line 9: should this not fetch the official repo at https://github.com/ElementsProject/peerswap/ ?

northben (commented):
I'm keen to use PeerSwap with clightning.

It looks like the PR currently pulls a private repo from OP however.

jonasnick:
Agree @northben that it would be nice to have in nix-bitcoin. There hasn't been any activity in this PR for a long time. Closing.

PR #464, opened 2022-03-10
secrets: fix secrets setup warning

erikarvstedt:
#### Commit msg
With newer nixpkgs versions, a type error was shown instead of the warning message when `secretsSetupMethod` was unset.

Reproduce with:
```bash
nix eval --impure --expr '
(import <nixpkgs/nixos> {
  configuration = {
    imports = [
      <nix-bitcoin/modules/modules.nix>
    ];
  };
}).system
'
```

### DIFF
--- a/modules/secrets/secrets.nix
+++ b/modules/secrets/secrets.nix
@@ -66,7 +66,7 @@ let
     };
 
     secretsSetupMethod = mkOption {
-      type = types.str;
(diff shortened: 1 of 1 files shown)


jonasnick (approved):
ACK 91203d8d06344fb84abe19a11aeefd414407bc55

Issue #186, opened 2020-05-28
Add default monitoring configuration


reardencode:
I don't think that nix-bitcoin should go so far as to install a complete monitoring and visualization stack, but some basic service status alerting could be really useful.

One example of how this can be achieved is:

https://gitlab.com/simple-nixos-mailserver/nixos-mailserver/-/blob/master/default.nix#L576

mmilata:
Would be nice to provide prometheus exporters, personally I'd run prometheus+grafana on separate machine:
* https://github.com/jvstein/bitcoin-prometheus-exporter
* https://github.com/lightningd/plugins/tree/master/prometheus (#71 related)
* lnd has built-in exporter which has to be [enabled at compile time](https://github.com/lightninglabs/lndmon/blob/master/INSTALL.md#lnd)

jonasnick:
As far as I know monit can only send alerts via email. That's easy to do from a mail server, but not for us really.

I have a prometheus + grafana setup with lnd. You can change the package used by the lnd module with the `package` option in order to use an lnd with the compilation flag enabled. However, the default metrics it exports are not useful. You really want https://github.com/lightninglabs/lndmon too. I'm relatively happy with that setup (though lndmon should export more things), prometheus queries and alerting through grafana is quite powerful.

Perhaps we should 1) make it easy to enable prometheus exporters and 2) add an option that enables a prometheus/grafana setup that "just works".

mmilata:
Opened nixpkgs PR for the bitcoind exporter: https://github.com/NixOS/nixpkgs/pull/89267 It's probably to more convenient to have it in nixpkgs because there's already a bunch of infrastructure for exporters that I'm not sure can be used from the outside. It's not hard to make it into standalone module though, let me know if there's interest.

nixbitcoin:
ACK standalone module and "just works" setup

mmilata:
lndmon package + module: NixOS/nixpkgs#89449. Let me know if you'd like to become comaintainer.

nixbitcoin:
IMO we should limit the scope and just make it a matrix or mail notification whenever a systemd service goes down.

joaothallis:
> IMO we should limit the scope and just make it a matrix or mail notification whenever a systemd service goes down.

[...]

PR #472, opened 2022-04-05
monit-nb: add module

nixbitcoin:
Closes https://github.com/fort-nix/nix-bitcoin/issues/186

nix-bitcoin monitoring via monit

Messages are sent to the local mail spool and can then be retrieved via IMAP over the dovecot onion service. Thunderbird users can connect like so https://www.autistici.org/docs/anon/tor

Unsolved problem: On some deployment mechanisms the system switch happens before secrets are deployed. Therefore, on these systems, `users.users.monitmail.passwordFile` fails because the file doesn't exist yet.


nixbitcoin:
Updated: I changed monit to only alert on service status changes. This works under the assumption that everything will be running on the first deploy. Also I added the service to `configuration.nix` and `README.md`.

Running this in production now and fetching mail via `mpop` over tor with neomutt.

PR #476, opened 2022-04-30
defaultHardening: allow `clone3` system call

erikarvstedt:
#### Copy of commit msg
clone3 is the latest version of the clone system call, which is already allowed.
clone3 is required by nbxplorer 2.3.20.


erikarvstedt:
It turns out that the new nxbplorer release requires more changes. I'll bundle these in a single PR.

PR #481, opened 2022-05-07
services: set systemd list options as list values

erikarvstedt:
#### Copy of commit msg
This makes our list definitions mergeable with custom list values set by users.
Previously, a module error ("value is a string while a list was expected") was thrown instead.

This commit was partly auto-generated with this script:

```ruby
#!/usr/bin/env ruby
Dir["**/*.nix"].each do |file|
  src = File.read(file)
  fixed = src.gsub(/ReadWritePaths *= *(.*?);/) do
    "ReadWritePaths = [ #{$1} ];"
  end
  File.write(file, fixed) if fixed != src
end
```


jonasnick (approved):
ACK e6bb281a8893671a1898ea01dc382fa98e3e9e70

PR #483, opened 2022-05-11
pkgs-unstable: inherit system from stable pkgs

### LINKED ISSUE #482: infinite recursion encountered
Hi, I'm trying to import `nix-bitcoin` into an existing project. I've tried using `examples/importable-configuration.nix` as a starting point and got a nixos system built. However, the bitcoin service fails to start with `exec format error`. Debugging a bit revealed that the `bitcoin` binary was a Mach-O format binary.

I'm developing on a MacOS machine, but the image is being built on a remote builder. All the other services are built and run fine, but `nix-bitcoin` does not compile for linux, and ends up getting compiled for MacOS everytime.

Any assistance or pointers would be highly appreciated.

erikarvstedt:
Previously, `builtins.defaultSystem` was implicitly used.
This fixes NixOS system builds for systems other than `defaultSystem`. 
(Flake-based systems are unaffected.)

Fixes #482

### DIFF
--- a/pkgs/default.nix
+++ b/pkgs/default.nix
@@ -3,7 +3,11 @@ let
 in
 # Set default values for use without flakes
 { pkgs ? import <nixpkgs> { config = {}; overlays = []; }
-, pkgsUnstable ? import nixpkgsPinned.nixpkgs-unstable { config = {}; overlays = []; }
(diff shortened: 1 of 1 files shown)


jonasnick (approved):
ACK 472bcf1565a7d8558b8dacfd4ee78a19112c8a60

PR #486, opened 2022-05-17
Minor improvements

erikarvstedt:


modules/presets/bitcoind-remote.nix
```diff
@@ -22,4 +22,7 @@ in {
       RemainAfterExit = true;
     };
   };
+
+  # Disable banlist loading
+  systemd.services.bitcoind-import-banlist = lib.mkForce {};
```
jonasnick, line 27: Why disable the banlist specifically in this preset?

erikarvstedt: I've removed this.
My original reasoning was that client banlist loading is redundant in a scenario where multiple nix-bitcoin clients connect to a nix-bitcoin server.
But it's still relevant in case of a non nix-bitcoin server.

jonasnick (approved):
ACK ef93a9a8bd8b14d3f50e4b4fd6564a50234b0e4a

PR #487, opened 2022-05-20
Add feeadjust core lightning plugin

elsirion:
The `feeadjuster` plugin sets channel fees based on their how balanced they are. This PR implements a subset of its config options with some opinionated defaults and allows running the fee adjustment once a day via a systemd timer.

I don't really know what I'm doing in nix, so any feedback and ways to do things better are much appreciated.

### DIFF
--- a/modules/clightning-plugins/feeadjuster.nix
+++ b/modules/clightning-plugins/feeadjuster.nix
@@ -21,6 +21,11 @@ let maybeFlag = enable: flag: if enable then flag + "\n" else ""; in
       default = "default";
       description = "Adjustment method to calculate channel fee (soft=less difference, hard=high difference)";
     };
+    adjustDaily = mkOption {
+      type = types.bool;
+      default = true;
+      description = "Automatically update fees daily";
+    };
   };
 
   config = mkIf cfg.enable {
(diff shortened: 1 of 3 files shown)


modules/clightning-plugins/feeadjuster.nix
```diff
@@ -0,0 +1,61 @@
+{ config, lib, ... }:
+
+with lib;
+let cfg = config.services.clightning.plugins.feeadjuster; in
+let maybeFlag = enable: flag: if enable then flag + "\n" else ""; in
```
elsirion, line 5: There has to be a builtin for this, I just couldn't find it …

prusnak: Yes, there is - it's called `optionalString` see https://github.com/NixOS/nixpkgs/blob/772e39b23727a5e69bf021fe1bde079b965580e8/lib/strings.nix#L199

erikarvstedt:
[Fixup](https://github.com/erikarvstedt/nix-bitcoin/commits/cl-feeadjuster-ea).
You can squash this and `Add timer to run feeadjust daily` into the first commit.

modules/clightning-plugins/feeadjuster.nix
elsirion: Thx!

erikarvstedt:
Thanks. Can you also squash `Add timer to run feeadjust daily` into the first commit?

modules/clightning-plugins/feeadjuster.nix
```diff
@@ -0,0 +1,60 @@
...
+      description = "Automatically update fees daily";
+    };
+  };
+
+  config = mkIf cfg.enable {
+    services.clightning.extraConfig = ''
+      plugin=${config.nix-bitcoin.pkgs.clightning-plugins.feeadjuster.path}
+      feeadjuster-adjustment-method="${cfg.method}"
+    '' +
+    (optionalString (!cfg.fuzz) "feeadjuster-deactivate-fuzz") +
```
erikarvstedt, line 35: This is missing a trailing newline, so all options are defined in a single line without separation. This is all worked out in my fixup.

elsirion:
Thx @erikarvstedt! Sorry for the confusion, I had just started applying @prusnak's fix when you added your comment and didn't see it before force pushing. I just used your version now and squashed it. Unfortunately git didn't preserve you co-authorship while squashing, I can look into how to add it.

modules/clightning-plugins/feeadjuster.nix
elsirion: I used your version

erikarvstedt:
Thanks, no worries about the authorship.
I've just noticed that we should also test the feedjuster service; this is implemented [here](https://github.com/erikarvstedt/nix-bitcoin/commits/cl-feeadjuster-ea-2).

erikarvstedt:
Thanks! I've [rebased this to master](https://github.com/erikarvstedt/nix-bitcoin/commits/cl-feeadjuster-ea-3) so we can use the new `clightning.package` option.

erikarvstedt (approved):
ACK f9315db52c03263a0bf9be025d768864cd7beb3a

PR #488, opened 2022-05-22
Update `clightning-plugins`

erikarvstedt:
Commit `clightning: add package option` is also useful for #487.


jonasnick (approved):
ACK ae94665ad18ca3b49f485cce38bb4f2023de3f2e

PR #489, opened 2022-05-27
Update nixpkgs & joinmarket: 0.9.5 -> 0.9.6

nixbitcoin:


prusnak:
ISTM that joinmarket 0.9.6 is compatible with twisted 22.4.0 - see https://github.com/JoinMarket-Org/joinmarket-clientserver/pull/1264

We can just fetchpatch https://github.com/JoinMarket-Org/joinmarket-clientserver/commit/15ac4de721c2fc0cb393f8ee6a534967fab8ad33.patch and we won't need to introduce custom twisted package

jonasnick:
Can you reorder the commits so they pass the tests individually? The update-nixpkgs commit doesn't pass if the joinmarket update isn't applied yet.

jonasnick:
@nixbitcoin With the previous version of this PR, the joinmarket update could've come before the nixpkgs update. Ok, fine with me as is then.

erikarvstedt (approved):
ACK cb295d1a166ecea96f5ff336c7d9cac9087f1953

PR #494, opened 2022-06-03
Fix cln rpc path for btcpayserver on regtest

afilini:


jonasnick (approved):
utACK 961192313c5369593fd53693f62e6ff23dd361db

PR #496, opened 2022-06-04
Add lnbits-legend

afilini:
This PR adds "lnbits legend" (https://github.com/lnbits/lnbits-legend) with support for both clightning and lnd.

Unfortunately it requires quite a few python packages that are not shipped by nixos, so I had to add them here as well.

This is my first relatively large contribution to a project in nix, so I apologize if there are any obvious mistakes.

### DIFF
--- a/modules/lnbits.nix
+++ b/modules/lnbits.nix
@@ -1,0 +1,259 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+let
+  options.services = {
+    lnbits = {
+      enable = mkEnableOption "LNbits, free and open-source lightning-network wallet/accounts system. ";
+      listen = {
+        address = mkOption {
+          type = types.str;
+          default = "127.0.0.1";
+          description = "Address to listen on.";
+        };
+        port = mkOption {
+          type = types.port;
+          default = 5000;
+          description = "Port to listen on.";
+        };
+        forceHttps = mkOption {
+          type = types.bool;
+          default = false;
+          description = "Force HTTPS.";
+        };
+      };
+      publicAddress = {
+        host = mkOption {
+          type = types.str;
+          default = config.services.lnbits.listen.address;
+          description = "Public facing hostname or ip, used when building callback URLs.";
+        };
+        port = mkOption {
+          type = types.port;
+          default = config.services.lnbits.listen.port;
+          description = "Public facing port, used when building callback URLs.";
+        };
+      };
+      dataDir = mkOption {
+        type = types.path;
+        default = "/var/lib/lnbits";
+        description = "The data directory for lnbits.";
+      };
+      lightningBackend = mkOption {
+        type = types.nullOr (types.enum [ "clightning" "lnd" ]);
+        default = null;
+        description = "The lightning node implementation to use.";
+      };
+      database = {
+        backend = mkOption {
+          type = types.enum [ "sqlite" "postgres" ];
+          default = "sqlite";
+          description = "The database implementation used.";
+        };
+
+        postgresConfig = {
+          databaseName = mkOption {
+            type = types.str;
+            default = "lnbits";
+            description = "Database name.";
+          };
+        };
+      };
+      allowedUsers = mkOption {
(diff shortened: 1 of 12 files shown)


prusnak:
Couple of questions:
- why not lnbits-infinity? it's much easier to package golang stuff
- what use-case do you want to bring to nix-bitcoin? if it's just L3 custodial LN wallets, maybe lndhub.go is better choice? (but it only supports lnd for the time being)

afilini:
> why not lnbits-infinity? it's much easier to package golang stuff

On their website they say "(Legend is the commonly used and tested version, Infinity is more experimental)", so I figured "legend" is still the main implementation, at least for the time being.

> what use-case do you want to bring to nix-bitcoin?

To be honest I don't know many of the alternatives and their tradeoffs compared to lnbits, the reason why I decided to package this specifically is that I personally need a backend for a Bleskomat and since I had to package lnbits for myself I thought I might as well share it in case somebody else finds it useful.

From my perspective I also feel like lnbits is still the most popular backend in used today, or at least this is what I see mentioned the most on twitter. There are a few interesting plugins that I'd like to try out and being in Python it also seems very easy to extend.

So in summary I'm not exactly trying to add a specific functionality to nix-bitcoin, I'm just adding a piece of software that I personally need and will use in the future and I guess other people might want to try as well.

afilini:
> Would you be okay with initially adding lndbits as a Flake?

I've never used flakes, so this may take a while, but I can give it a shot!

erikarvstedt:
@afilini, I'll happily do the flakes conversion and report back soon.

arcbtc:
@afilini is also 100% correct. Infinity is on a back burner for most of the community until we get legend out of beta, then we will all pile in on infinity. Both have slightly different approaches, which I am sure will serve different uses, so although its a bit weird I imagine both continuing to exist

MatthewCroughan:
It would be very helpful if @afilini could help maintain the Nix code I'm adding to lnbits-legend upstream, and if @jonasnick could add nodes to the README that link out to projects that have added Nix properly to their own repositories, rather than having it all packaged here.

As part of what I have done in the link Ben shared, I have also converted lnbits-legend to poetry, and have had to modify a lot of the Python code to stop making assumptions about relative paths, and give absolute paths instead. There's a lot of Python refactor work that needs to be done in lnbits-legend to make it work correctly anyway. This is something we can do gradually over time.

I'm working on submitting a PR to lnbits-legend which notes all of the Python refactor work that needs to be done, along with some decent Nix code that works around all of the problems and introduces `services.lnbits.enable` along with a VM test. The VM test is something you can run right now:

If you do `nix build github:matthewcroughan/lnbits-legend/matthewcroughan/nixify#checks.x86_64-linux.vmTest` you'll run a test that:
1. Builds lnbits
2. Runs two virtual machines
3. Proves that a client VM can talk to a server VM running LNBits correctly, and that `curl` returns the correct HTML title, rather than uvicorn crashing (which it did a lot in my testing, until I managed to make it work! :D)

MatthewCroughan:
The PR I've made to lnbits-legend is now visible here https://github.com/lnbits/lnbits-legend/pull/721

arcbtc:
Matts PR was merged, so you can use the flake or:
https://github.com/lnbits/lnbits-legend/blob/main/docs/guide/installation.md#option-4-nix

HaosGames:
I am currently trying to fix the flake for the upstream lnbits but am stuck at figuring out which other paths to symlink to mutable locations. https://github.com/lnbits/lnbits/pull/1855
Maybe someone else is willing to help out

prusnak:
I think we can close this PR now that lnbits has functional flake in the upstream repo: https://github.com/lnbits/lnbits/blob/dev/flake.nix

MatthewCroughan:
@prusnak I am not sure it makes any sense to have nix-bitcoin contain a reference to lnbits. That kind of defeats the purpose and point of flakes. Flakes that vendor other's flakes don't make sense.

PR #498, opened 2022-06-27
clightning/lnd: allow connections to remote bitcoind in systemd service

seberm:
clightning and lnd services should be allowed to connect to a remote bitcoind daemon

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -214,6 +214,21 @@ services.bitcoind = {
 };
 ```
 
+If a `secure-node.nix` or `tor-enable.nix` preset is imported in your
+configuration or a `tor.enforce` option is explicitly enabled, you also need to
+allow remote connections for **every** service which needs to connect to the
+remote bitcoind:
+
+```
+systemd.services.<service>.serviceConfig = {
+  IPAddressAllow = [ ${services.bitcoind.rpc.address} ];
+};
+```
+
+> Please note that configuration above applies only if the remote bitcoind **is
+> not** accessed via Tor.
+
+
 Now save the password of the RPC user to the following files on your nix-bitcoin node:
 ```shell
 $secretsDir/bitcoin-rpcpassword-privileged


jonasnick (commented):
Hi @seberm, thanks for your contribution! Glad you figured out what the problem was.

I don't think this is the right solution. The `tor.enforce` [description](https://search.nixos.org/flakes?channel=22.05&from=0&size=50&sort=relevance&type=options&query=nix-bitcoin+clightning.tor.enforce) says

> Whether to enforce Tor on a service by only allowing connections from and to localhost and link-local addresses.

This PR would circumvent this enforcement. So if you want to connect to an external bitcoind you should either connect through Tor or turn off `tor.enforce`. I think what we should do at this point is to document this clearly in the ["Use bitcoind from another node"](https://github.com/fort-nix/nix-bitcoin/blob/master/docs/configuration.md#use-bitcoind-from-another-node) section. Note that `tor.enforce` is only enabled because you presumably also imported the `secure-node.nix` prefix which is not imported in this documentation section.

Just updating the docs is certainly not ideal because in your situation you probably want `tor.enforce` for everything but the IP address of your external bitcoind. We should also think about how we can allow such customization without redefining the semantics of `tor.enforce`.

seberm:
> Hi @seberm, thanks for your contribution! Glad you figured out what the problem was.
> 
> I don't think this is the right solution. The `tor.enforce` [description](https://search.nixos.org/flakes?channel=22.05&from=0&size=50&sort=relevance&type=options&query=nix-bitcoin+clightning.tor.enforce) says
> 
> > Whether to enforce Tor on a service by only allowing connections from and to localhost and link-local addresses.
> 
> This PR would circumvent this enforcement. So if you want to connect to an external bitcoind you should either connect through Tor or turn off `tor.enforce`. I think what we should do at this point is to document this clearly in the ["Use bitcoind from another node"](https://github.com/fort-nix/nix-bitcoin/blob/master/docs/configuration.md#use-bitcoind-from-another-node) section. Note that `tor.enforce` is only enabled because you presumably also imported the `secure-node.nix` prefix which is not imported in this documentation section.

I agree. We should definitely document the current behaviour at least.

> Just updating the docs is certainly not ideal because in your situation you probably want `tor.enforce` for everything but the IP address of your external bitcoind. We should also think about how we can allow such customization without redefining the semantics of `tor.enforce`.

Yeah, that's exactly what I want.

Maybe we could just somehow detect, if [`bitcoind-remote.nix`](https://github.com/fort-nix/nix-bitcoin/blob/master/modules/presets/bitcoind-remote.nix) preset is imported and then add its IP address into `nbLib.allowedIPAddresses`? But I understand that when `tor.enforce` is active, this probably shouldn't be possible at all. Or do you have a better idea?

> So if you want to connect to an external bitcoind you should either connect through Tor or ...

Is this even possible right know? If I set the `services.bitcoind.address = "<remote-onion-address>.onion"` and `services.bitcoind.rpc.address = "<remote-onion-address>.onion"`, the tor daemon fails to start:

```
$ tor -f /nix/store/23hvkknx8i56g4p6bji4c8l6f4a8hzwq-torrc --verify-config
Jun 29 15:05:15.416 [notice] Tor 0.4.7.8 running on Linux with Libevent 2.1.12-stable, OpenSSL 1.1.1o, Zlib 1.2.12, Liblzma 5.2.5, Libzstd 1.5.2 and Glibc 2.34 as libc.
Jun 29 15:05:15.416 [notice] Tor can't help you if you use it wrong! Learn how to be safe at https://support.torproject.org/faq/staying-anonymous/
Jun 29 15:05:15.416 [notice] Read configuration file "/nix/store/23hvkknx8i56g4p6bji4c8l6f4a8hzwq-torrc".
Jun 29 15:05:15.425 [warn] Unparseable address in hidden service port configuration.
```

The problem is in a torrc configuration:
```
HiddenServiceDir /var/lib/tor/onion/bitcoind
HiddenServicePort 8333 <remote-onion-address>.onion:8334
```

I think we shouldn't configure a bitcoind hidden service when a `bitcoind-remote.nix` preset is active or at least when `services.bitcoind.address` is an onion address. What do you think?

Thank you!

docs/configuration.md
```diff
@@ -214,6 +214,19 @@ services.bitcoind = {
...
+> configuration or a `tor.enforce` option is explicitly enabled, you also need
+> to allow remote connections for **every** service which needs to connect to
+> the remote bitcoind:
+> 
+> ```
+> systemd.services.<service>.serviceConfig = {
+>   IPAddressAllow = ["10.10.0.2/32"];
+>   IPAddressAllow = [ ${services.bitcoind.rpc.address} ];
+> };
+> ```
```
jonasnick, line 227: Is there a particular reason that this is a blockquote? I don't think we do this anywhere else.

docs/configuration.md
```diff
@@ -214,6 +214,19 @@ services.bitcoind = {
...
 ```
 
+> Note: If a `secure-node.nix` or `tor-enable.nix` preset is imported in your
+> configuration or a `tor.enforce` option is explicitly enabled, you also need
+> to allow remote connections for **every** service which needs to connect to
+> the remote bitcoind:
+> 
+> ```
+> systemd.services.<service>.serviceConfig = {
+>   IPAddressAllow = ["10.10.0.2/32"];
```
jonasnick, line 224: What's this address? An example for a hardcoded ip address?

docs/configuration.md
```diff
@@ -214,6 +214,19 @@ services.bitcoind = {
 };
 ```
 
+> Note: If a `secure-node.nix` or `tor-enable.nix` preset is imported in your
+> configuration or a `tor.enforce` option is explicitly enabled, you also need
```
jonasnick, line 218: Perhaps also mention that this only applies if the remote bitcoind isn't accessed via Tor?

jonasnick (commented):
Thanks for the updated PR.

> But I understand that when tor.enforce is active, this probably shouldn't be possible at all.

Yes, this should be the behavior of tor.enforce. There are many ways to achieve what you want. Perhaps easiest (but not very clean) approach would be to add a tor.enforceOverride option.

> I think we shouldn't configure a bitcoind hidden service when a bitcoind-remote.nix preset is active or at least when services.bitcoind.address is an onion address. What do you think?

Interesting, I would have thought this just works. OTOH I don't know how useful it is to access a remote bitcoind via Tor.

jonasnick (approved):
ACK 8cac8285347bc8063ae357675ee4bf8e2d632f27

PR #502, opened 2022-07-04
krops: don't reimport pkgs

erikarvstedt:
#### Copy of commit msg
Previously, the krops pkg was accessed by evaluating nixpkgs with the krops overlay.
Now directly call krops with an existing pkgs instance.


jonasnick (approved):
ACK 68ef662654c9703083cc0fb8e3a1f3c970ad4026

PR #503, opened 2022-07-04
nodeinfo: make extensible

erikarvstedt:
#### Copy of commit msg
Module authors can now add extra services via option `nix-bitcoin.nodeinfo.services`.
Mark as internal because we're not yet providing a user-friendly, stable API.

Test this with:
```
read -d '' config <<'EOF' || :
{ config, ... }: {
  options = {
    services.myservice.enable = mkOption { default = true; };
  };

  config = {

    nix-bitcoin.nodeinfo = {
      enable = true;
      services = {
        myservice = _: _: ''
          infos["myservice"] = "example output"
        '';
      };
    };
  };
}
EOF
run-tests.sh -s "$config" container --run c nodeinfo
```

### DIFF
--- a/modules/nodeinfo.nix
+++ b/modules/nodeinfo.nix
@@ -5,40 +5,37 @@ let
   options = {
     nix-bitcoin.nodeinfo = {
       enable = mkEnableOption "nodeinfo";
+
       program = mkOption {
         readOnly = true;
         default = script;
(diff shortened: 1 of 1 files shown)


jonasnick (approved):
That's pretty nice!

ACK f17fcebe117fc69b2c811fc0ab0a64a591f2e20b

PR #506, opened 2022-07-04
bitcoind: set `onlynet=tor` on `tor.enforce`

erikarvstedt:
#### Copy of commit msg
This prevents bitcoind from attempting clearnet connections which result in errors due to Tor enforcement.


erikarvstedt:
There was a misunderstanding on my part. `tor.enforce` and `onlynet=tor` are orthogonal, so this setting doesn't make sense.

PR #509, opened 2022-07-05
bitcoind, liquidd: increase `TimeoutStartSec`

erikarvstedt:
### Copy of commit msg
I've just seen `liquidd` hit the timeout on nixbitcoin.org while `Loading block index`.
This was probably due to HDD contention while starting services during
boot.


jonasnick (commented):
ACK 99f1dc0bff9a9a098a30811f887aa4ec549d3e93

jonasnick (approved):

PR #510, opened 2022-07-06
lnd, lightning-loop: add certificate options `extraIPs` and `extraDomains`

erikarvstedt:
#### Copy of commit msg
This is useful for non-local access to the lnd REST server.

@sputn1ck, does this fit your use case?

### DIFF
--- a/modules/lightning-loop.nix
+++ b/modules/lightning-loop.nix
@@ -40,6 +40,26 @@ let
       default = if cfg.tor.proxy then config.nix-bitcoin.torClientAddressWithPort else null;
       description = "host:port of SOCKS5 proxy for connnecting to the loop server.";
     };
+    certificate = {
+      extraIPs = mkOption {
+        type = with types; listOf str;
+        default = [];
+        example = [ "60.100.0.1" ];
+        description = ''
+          Extra `subjectAltName` IPs added to the certificate.
+          This works the same as loop option `tlsextraip`.
+        '';
+      };
+      extraDomains = mkOption {
+        type = with types; listOf str;
+        default = [];
+        example = [ "example.com" ];
+        description = ''
+          Extra `subjectAltName` domain names added to the certificate.
+          This works the same as loop option `tlsextradomain`.
+        '';
+      };
+    };
     extraConfig = mkOption {
       type = types.lines;
       default = "";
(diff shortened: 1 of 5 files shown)


sputn1ck (dismissed):
Cool PR, exactly what I needed. Now I just need pool and loop to be able to listen to multiple rpc hosts and my remote LIT setup is perfect (currently using ssh tunneling to forward the grpc  connections).

modules/lnd.nix
```diff
@@ -276,9 +302,14 @@ in {
...
-    nix-bitcoin.generateSecretsCmds.lnd = ''
+    nix-bitcoin.generateSecretsCmds.lnd = let
+      extraAltNames = builtins.concatStringsSep "," (
+        (map (ip: "IP:${ip}") cfg.certificate.extraIPs) ++
+        (map (domain: "DNS:${domain}") cfg.certificate.extraDomains)
+      );
+    in ''
       makePasswordSecret lnd-wallet-password
-      makeCert lnd '${optionalString (cfg.rpcAddress != "localhost") "IP:${cfg.rpcAddress}"}'
+      makeCert lnd '${extraAltNames}"}'
```
sputn1ck, line 312: I think `localhost` should always be added, as lncli/loop/pool is broken if not.

erikarvstedt:
> be able to listen to multiple rpc hosts 

You can use `extraConfig`:
```nix
services.lightning-loop.extraConfig = ''
   rpclisten=myhost:myport
'';
```

erikarvstedt:
I've now added the same options to `lightning-loop`, which uses the same certificate setup mechanism as `lnd`.

sputn1ck:
> > be able to listen to multiple rpc hosts
> 
> You can use `extraConfig`:
> 
> ```nix
> services.lightning-loop.extraConfig = ''
>    rpclisten=myhost:myport
> '';
> ```

I'm pretty sure loop and pool currently can only listen on one rpc host, so it would just take the last one you define in `services.lightning-loop.extraConfig`

erikarvstedt:
Ah, I thought it worked like in `lnd`.
Until this is fixed upstream, you could use `socat` or `nginx` for TCP forwarding.

modules/lightning-loop.nix
```diff
@@ -40,13 +40,37 @@ let
...
       type = types.lines;
       default = "";
       example = ''
         debuglevel=trace
       '';
-      description = "Extra lines appended to the configuration file.";
+      description = ''
+        Extra lines appended to the configuration file.
+        See here for all available options:
+        https://github.com/lightninglabs/loop/blob/11ab596080e9d36f1df43edbeba0702b25aa7457/loopd/config.go#L119
```
nixbitcoin, line 72: This should point to master

jonasnick (approved):
ACK c853f380d43279813a8c4ae2a955f9cbf0bb7231

PR #515, opened 2022-07-15
bitcoind-remote: fix quotation marks in preStart script

seberm:
I think there is a small typo in quotation marks.

### DIFF
--- a/modules/presets/bitcoind-remote.nix
+++ b/modules/presets/bitcoind-remote.nix
@@ -13,7 +13,7 @@ in {
 
   systemd.services.bitcoind = {
     preStart = lib.mkAfter ''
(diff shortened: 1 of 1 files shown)


seberm:
I think we could also run a shell lint (e.g. [shellcheck](https://github.com/koalaman/shellcheck) across all/most of the scripts in this repository and fix the issues.

It would be also great if we could run shellcheck tests automatically in Cirrus CI for all the generated scripts (e.g. unit's `ExecStart=`, `ExecStartPre=`, `ExecStartPost`, ... options). What do you think?

erikarvstedt (approved):
This patch is non-functional and fixes no error, but I agree this quoting style adds consistenty with the service scripts.

ACK 03b75a025459685fc6ede4e651be0183138e6bb0

erikarvstedt:
> What do you think?

Concept ACK. Would you like to implement this?

seberm:
> Would you like to implement this?

I'll try to look into it, thanks! :)

PR #530, opened 2022-08-04
onion-services: don't propagate restart of `onion-addresses` to depending services

erikarvstedt:
#### Copy of commit msg
Currently, public services that publish their onion addresses have a `requires` dependeny on service `onion-addresses`, so they are restarted when `onion-addresses` restarts.

This has the downside that most of nix-bitcoin's services are restarted when a new onion service is added and the system config has the following common settings:
- nix-bitcoin.onionServices.bitcoind.public = true
- nix-bitcoin.operator.enable = true

Sequence of events:
1. onion service is added
2. `onion-addresses` restarts, because the new onion service is made available to `operator`
3. bitcoind restarts
4. all depending services restart

Fix this by using a `wants` dependency.

### DIFF
--- a/modules/bitcoind.nix
+++ b/modules/bitcoind.nix
@@ -376,6 +376,9 @@ in {
       # are not restarted when the secrets target restarts.
       # The secrets target always restarts when deploying with one of the methods
       # in ./deployment.
+      #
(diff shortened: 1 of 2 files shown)


jonasnick (approved):
utACK a3490dfc0f7fbb8d06cf0dd9a2e2beaca6c804d1

PR #533, opened 2022-08-12
clightning: add trustedcoin plugin

neverupdate:
Adds the [trustedcoin](https://github.com/fiatjaf/trustedcoin) plugin to allow using clightning without bitcoind, and an option to disable the bcli plugin in clightning, so trustedcoin can be used instead.

I didn't remove the clightning dependency on the bitcoind service though, but that could also be done.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,7 @@ NixOS modules ([src](modules/modules.nix))
     * [prometheus](https://github.com/lightningd/plugins/tree/master/prometheus): lightning node exporter for the prometheus timeseries server
     * [rebalance](https://github.com/lightningd/plugins/tree/master/rebalance): keeps your channels balanced
     * [summary](https://github.com/lightningd/plugins/tree/master/summary): print a nice summary of the node status
+    * [trustedcoin](https://github.com/fiatjaf/trustedcoin): replaces bitcoind with trusted public explorers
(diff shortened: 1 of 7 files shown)


MatthewCroughan:
I actually did this exact same thing a week ago and never up-streamed it. Our results are basically identical! Bravo.

Can you split the commit messages up like so?:

`trustedcoin: add pkg`
`trustedcoin: add module`
`readme: reference trustedcoin source`
`clightning-plugins: add trustedcoin`
`clightning: add useBcli option to module`

Since nix-bitcoin is a library like nixpkgs, it makes sense to make the commit messages granular like this,

neverupdate:
Should be good now!

MatthewCroughan:
I suppose the only thing lacking, is a test. Because this plugin disables the bcli and doesn't require the bitcoin daemon in how it's recommended to be used, it is quite a tricky one to test that mode, without modifying the testing infrastructure.

erikarvstedt:
Does it ever make sense to use `trustedcoin` with only `bcli` disabled?
With this setting, `trustedcoin` uses bitcoind-rpc as its main data source and falls back to public explorers when bitcoind-rpc fails.
But bitcoind-rpc should never fail with nix-bitcoin. Or might this be relevant during the initial block downloading in bitcoind?

Wouldn't it be better to add an option `clightning.useBitcoind` to disable bitcoind alltogether when using clightning with trustedcoin?

PR #539, opened 2022-08-22
Minor improvements

erikarvstedt:


helper/run-in-nix-env
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
```
jonasnick, line 2: How about we add a header that documents what this script does?

jonasnick (commented):
Needs rebase, but looks good to me otherwise.

helper/run-in-nix-env
erikarvstedt: +1. Done.

erikarvstedt: `run-in-nix-env` was needlessly convoluted, I've cleaned it up.

jonasnick (approved):
ACK 67a573dae86ae43d30395864351b81ec03c8bef8

Issue #540, opened 2022-08-22
Add CLN Watchtower


HaosGames:
I would suggest adding a Watchtower for CLN so that funds in channels can be better secured. 
I think https://github.com/talaia-labs/rust-teos might be a good one

seberm:
It seems to me the rust-teos is still under heavy development. Is it considered stable? The first "stable" version of the project ([v0.1.1](https://github.com/talaia-labs/rust-teos/releases/tag/v0.1.1)) was released on 22 july 2022.

I can try to prepare patches which add the rust-teos into nix-bitcoin, then we can maybe add it into nix-bitcoin as experimental? What is your opinion @erikarvstedt @jonasnick ?

PR #543, opened 2022-08-30
Add basic support for teos (The Eye of Satoshi)

seberm:
This MR adds basic support for [rust-teos](https://github.com/talaia-labs/rust-teos) (package & module).

Fixes: #540


pkgs/teos/default.nix
```diff
@@ -0,0 +1,66 @@
...
+    cargoSha256 = "sha256-+V7Irf//Lj7VGFY/rdUrSC0hrblnIIYsWgcG4CKLy1w=";
+    buildAndTestSubdir = "teos";
+
+    inherit version src cargoPatches buildInputs nativeBuildInputs LIBCLANG_PATH;
+
+    cargoTestFlags = [
+      "--workspace"
+
+      # Problems with "Too many open files" error
+      "--exclude=teos"
```
meryacine, line 55: If I understand this correctly, this means `teos` isn't built at all!
Surely that's not the thing you are looking for, or is it?

Regarding the too many open files issue, I fixed it before by changing the open files soft and hard limits in `/etc/security/limits.conf`:
```
* soft nofile 262144
* hard nofile 524288
```
That was in an ubuntu 18 system.

erikarvstedt: Many Linux distros (including NixOS) have a soft limit of 1024 open files (`NOFILE`).
If you really need to exceed this (which I would doubt in cause of `teos`), your binary should use `setrlimit` to raise the limit to an appropriate value within in the bounds of the hard limit (which is 524288 for most distros).

meryacine: IIRC, this issue happened to me during compilation and not in runtime.

erikarvstedt: Still, users shouldn't have to raise their `NOFILE` soft limit when building teos. Can you open an issue in the teos repo where we can continue discussing this?

seberm: I have created the issue:
- https://github.com/talaia-labs/rust-teos/issues/119

seberm:
Hello @erikarvstedt , would you mind to take a look and give me some initial review?

seberm:
I think it makes sense to add teos package directly into nixpkgs and a teos module maintain here in nix-bitcoin:
- https://github.com/NixOS/nixpkgs/pull/195782

3cola (commented):
I may be wrong but services.api.onionPort needs to become services.onionPort to be used by the nix-bitcoin onion-services module.
If not, you will expose port 8814 to the tor network and not port 9814.
Also, you may need to specify an externalPort for tor. Commonly used by teos is the port 2121, like this:
```
nix-bitcoin.onionServices.teos.externalPort = 2121;
```

3cola (commented):
btc_network should be "mainnet" and not anymore "bitcoin"

seberm:
I think you are right. Good catch!

I've made the recommended changes. Could you please help me with the testing? I have currently my test environment somehow broken.

Thanks!

modules/teos.nix
```diff
@@ -0,0 +1,178 @@
...
+    port = mkOption {
+      type = types.port;
+      default = 8814;
+      description = mdDoc "Port to listen for RPC connections.";
+    };
+    externalPort = mkOption {
+      type = types.port;
+      default = 2121;
+      description = mdDoc "Override an external port to listen for onion API connections.";
+    };
```
3cola, line 21: this should not be here but you should setup an externalPort in modules/onion-services.nix, at the end of the file something like: 
```
nix-bitcoin.onionServices.teos = {
  enable = true;
  externalPort = 2121;
};
```

3cola: if you don't want to touch this module, you can set it up like in my comment for the example...

examples/configuration.nix
```diff
@@ -216,6 +216,12 @@
   # services.charge-lnd.policies = ''
   # '';
 
+  ### TEoS - The Eye of Satoshi
+  # Set this to enable teos, a Lightning watchtower compliant with BOLT13,
+  # written in Rust.
+  #
+  # services.teos.enable = true;
+
```
3cola, line 224: This could be an example configuration in case you don't want to touch the modules/onion-services.nix but still want use tor to expose the teosd endpoint.
```
    # teos
    nix-bitcoin.onionServices.teos = {
      enable = true;
      externalPort = 2121;
    };
    services.teos.enable = true;
```

3cola (commented):
externalPort should be in nix-bitcoin.onionServices.teos and not in services.teos.

3cola:
I am just testing the teosd setup for now, but not the watchtower plugin for c-lightning.

seberm:
Hello @3cola ,
I am continuing the work on integration of TEoS into nix-bitcoin (https://github.com/seberm/nix-bitcoin/tree/feature/add-rust-teos-v4).

> Commonly used by teos is the port 2121, like this: ....

It seems the TEoS [changed](https://github.com/talaia-labs/rust-teos/pull/145) the default port for onion from `2121` to match the clearnet one (`9814`). Do you think we really still want to use the `2121`?

I think we should follow the upstream and keep this port set to `9814` for both - clearnet and onion.

seberm:
I've moved the `api` configuration section to the top level so it is now correctly loaded by `onionServices`. The `port` and `address` options from top-level were moved under the `rpc` namespace.

I've also added some documentation about the plugin usage when connecting to a tower behind the tor.

I still need to review the tests. I would like to add some tests for watchtower plugin itself (e.g. register it with a local teos tower) and maybe add some basic tests for `teos-cli` (just make sure it will connect to the tower via local RPC).

I would be glad for a review. Thanks!

seberm:
Hello @erikarvstedt @prusnak,
would you mind helping with the review? Especially the integration into the nix-bitcoin codebase?

The tests are passing and I also deployed the teos service to my nix-bitcoin node and I did some tower registration requests using httpie over onion. From my point of view it seems the tower (teos), its client (teos-cli) and the cln plugin are working correctly.

It would be great if we could move this forward to getting this PR merged. Or would it be better to implement it as a flake?

Also, do you think that it makes sense to add watchtower's `dataDir` into the regular [bulk data backups](https://github.com/fort-nix/nix-bitcoin/blob/master/modules/backups.nix#L16)?

Thanks

3cola:
> Hello @3cola , I am continuing the work on integration of TEoS into nix-bitcoin (https://github.com/seberm/nix-bitcoin/tree/feature/add-rust-teos-v4).
> 
> > Commonly used by teos is the port 2121, like this: ....
> 
> It seems the TEoS [changed](https://github.com/talaia-labs/rust-teos/pull/145) the default port for onion from `2121` to match the clearnet one (`9814`). Do you think we really still want to use the `2121`?
> 
> I think we should follow the upstream and keep this port set to `9814` for both - clearnet and onion.

I don't have an opinion on the choice of port number. If upstream is now using only 9814, we should follow.

modules/clightning-plugins/teos-watchtower.nix
```diff
@@ -0,0 +1,51 @@
...
+      description = mdDoc "The package providing TEoS watchtower plugin binaries.";
+    };
+    port = mkOption {
+      type = types.port;
+      default = config.services.teos.port;
+      description = mdDoc "Tower API port.";
+    };
+    dataDir = mkOption {
+      type = types.path;
+      default = "${config.services.clightning.dataDir}/.watchtower";
```
mmilata, line 22: Is there a reason to make this a "hidden" directory?

mmilata (commented):
drive-by review, did not spot any major problem :+1:

modules/clightning-plugins/teos-watchtower.nix
seberm: Not really actually.

I've chosen this name based on a default watchtower plugin configuration `DEFAULT_TOWERS_DATA_DIR=~/.watchtower` which should in our case match with `${config.services.clightning.dataDir}/.watchtower`:
- https://github.com/talaia-labs/rust-teos/blob/a4accedfeecf46edb9c22738ac9787b80dc8af02/watchtower-plugin/src/constants.rs#L3
- https://github.com/talaia-labs/rust-teos/blob/a4accedfeecf46edb9c22738ac9787b80dc8af02/watchtower-plugin/README.md?plain=1#L101

But we always specify the directory explicitly using the `TOWERS_DATA_DIR`, so it's probably better to not hide the directory.

seberm:
Any chance of getting this merged? I'll be glad for the review/comments and hints for improvement. Thanks! @erikarvstedt @jonasnick

jonasnick:
Thanks @seberm for your effort to keep this PR up to date!

I talked to Sergi, the main developer of teos, and he says that he isn't actively developing it but is planning to continue maintenance mode. He hasn't managed to make a new release so far and suggests to use master instead of the release.

So I don't think there is anything fundamental that's blocking this module as long as it is useful even in its current state. Needs review before it can get merged.

seberm:
Hello @jonasnick,

> He hasn't managed to make a new release so far and suggests to use master instead of the release.

The `teos` package is now [part of nixkpgs upstream](https://github.com/NixOS/nixpkgs/pull/195782). I am not sure if we can use the master branch directly there.

However, if Sergi recommends using the master branch, I can prepare a package overlay at least for nix-bitcoin.

**For anyone interested in getting this merged**, please test it and feel free to provide a review. I'll be glad to address any issues.

Thanks!

HaosGames (commented):
Looks good to me. 
I started testing this now with CLN as a backend and the teos daemon on a separate host with a pruned bitcoind. 
Works great so far with minimal configuration.

HaosGames:
Still running great on my side. I would love to see this in a release.

sr-gi:
I'm planning to cut a release sooner rather than later, which will include all the changes pointed out in the PR description AFAICT. Is there anything else that'd needed?

JosephGoulden (approved):
I'm running the watchtower and plugin modules from this PR and its working well.

It would be nice to have a service that automatically registers with the tower and renews the subscription when needed. But we can add that later.

Also I opened a [PR](https://github.com/talaia-labs/rust-teos/pull/266) to add a flake to TEOS.

HaosGames (approved):

seberm:
@nixbitcoin @erikarvstedt @jonasnick

Is it worthwhile to continue working on and rebasing this PR? Is there a possibility that nix-bitcoin will incorporate these changes?
Thanks!

PR #545, opened 2022-09-02
test/lib/create-git-repo: fix variable quoting

erikarvstedt:
Remove quotes around var that may expand to nothing.

### DIFF
--- a/test/lib/create-git-repo.sh
+++ b/test/lib/create-git-repo.sh
@@ -2,14 +2,14 @@
 (
   # shellcheck disable=SC2154,SC2164
   cd "$scriptDir/.."
-  amend=--amend
+  amend=(--amend)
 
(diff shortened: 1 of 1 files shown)


test/lib/create-git-repo.sh
```diff
@@ -10,6 +10,7 @@
   fi
   git add .
   if ! git diff --quiet --cached; then
-    git commit -a "$amend" -m -
+    # shellcheck disable=SC2086
+    git commit -a $amend -m -
```
jonasnick, line 14: I'm fine with merging this fix for now but https://www.shellcheck.net/wiki/SC2086 recommends using an array. Then we wouldn't have to add an obscure "shellcheck disable".

jonasnick (approved):
ACK 49e30f7172b154f6fc9e6ad062d1f624e7120d58

PR #546, opened 2022-09-03
Update nixpkgs

jonasnick:
Draft because the feeadjuster clightning plugin crashes, most likely because it's incompatible with the `pyln-client` library that ships with clightning 0.12.0. @elsirion do you have an idea how to fix this?


jonasnick:
Actually, the issue does not have anything to do with feeadjuster.  The problem is that the pyln-client library fails to parse the config, see
https://github.com/ElementsProject/lightning/issues/5546
https://github.com/ElementsProject/lightning/issues/5563

jonasnick:
I added a patch based on https://github.com/ElementsProject/lightning/pull/5574 that fixes pyln-client. Let's wait until it is ACKd.

pkgs/python-packages/pyln-client/default.nix
```diff
@@ -1,4 +1,4 @@
-{ buildPythonPackage, poetry-core, pytestCheckHook, clightning, pyln-bolt7, pyln-proto }:
+{ buildPythonPackage, fetchpatch, poetry-core, pytestCheckHook, clightning, pyln-bolt7, pyln-proto }:
```
erikarvstedt, line 1: `fetchpatch` is unused.

jonasnick: fixed

PR #547, opened 2022-09-05
tests: disable `nixosSearch`

erikarvstedt:
The nixos-search binary currently fails with:
```
Error: Getting flake info caused an error: Failed to analyze flake ../..

Caused by:
    0: [22]: data did not match any variant of untagged enum FlakeEntry at line 1 column 11194
    1: data did not match any variant of untagged enum FlakeEntry at line 1 column 11194
```
This happens locally and in [CI](https://github.com/erikarvstedt/nix-bitcoin/runs/8193778511).

The [NixOS option search](https://search.nixos.org/flakes?channel=unstable&from=0&size=30&sort=relevance&type=options&query=mail) also has stopped showing results (for all queries).

The error doesn't happen when run in a firejail container with `--net=none`. I currently don't have time for further debugging.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -35,7 +35,6 @@ task:
     - name: flake
       build_script:
         - nix flake check
-        - ./test/nixos-search/ci-test.sh
 
     - name: shellcheck
       build_script:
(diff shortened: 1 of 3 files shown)


jonasnick (approved):
ACK 7985e075e54bf083b9aa679556a4e027fb45f301

PR #551, opened 2022-09-23
tests: Reenable `flake-info`

erikarvstedt:


erikarvstedt:
The git checkout in Cirrus is currently broken.

erikarvstedt:
@jonasnick, the bwrap env exposed the host's `/etc/nix/nix.conf` to the `flake-info` process.
In your case, `nix.conf` probably contained the default value of `flake-registry`, which is `https://channels.nixos.org/flake-registry.json`. This file was then fetched by the underlying call to `nix`, which caused the connection failure.

Does the latest force-push fix this for you?

By the way, it makes sense to disable the global `flake-registry`, like so:
```
nix.settings.flake-registry = "/dev/null";
```
Otherwise, the remote registry is updated every hour (default value of nix setting `tarball-ttl`) when using the `nix` CLI, which adds a considerable, annoying delay. (IMO this is a design error, I've been planning to further investigate this and report it upstream.)
After disabling the global registry, you can still setup your own registry via files `/etc/nix/registry.json` or `~/.config/nix/registry.json`. [Here's](https://gist.github.com/5ca09ffe1aee646e70af17175001dced) a reasonable default.

jonasnick (approved):
Thanks! The latest force-push fixed this for me.

ACK 277510c7ee4acf0f2bc69eb3cd867d18d5aca41d

PR #552, opened 2022-09-23
test/shellcheck-services: simplify accessing service definitions

erikarvstedt:
#### Copy of commit msg
This also improves performance by removing the extra module evaluation.

#### Discussion
In my first implementation I totally missed that all required option info is available via module arg `options`.


erikarvstedt:
Included in #559.

PR #559, opened 2022-10-22
Define tests via flake

### LINKED ISSUE #554: Make test framework accessible via Flakes
This allows other flakes (e.g. flakes that define extra nix-bitcoin modules) to use the test framework.
When this is implemented, nix-bitcoin is fully extensible via flakes.
State: I'm working on this.

erikarvstedt:
#### Copy of main commit msg
Advantages:
- Pure test evaluations
- The test framework can now be used by flakes that extend nix-bitcoin
- Most features of `run-tests.sh` are now accessible via `nix build`/`nix run`.
  We keep `run-tests.sh` for advanced features like `scenarioOverridesFile` and adhoc scenarios.

Other changes:
- `run-tests.sh` now builds aggregate VM tests like `basic` or `buildable` by creating all VMs in a single evaluation.
  This speeds up the tests and eases debugging by separating the eval and build steps.
- Use the new `nix` CLI which has improved build output logging by prefixing output lines with the origin drv name.

#### Discussion
I've created [example-extension-flake](https://github.com/erikarvstedt/example-extension-flake) as a demo flake that adds extra nix-bitcoin modules and pkgs, including tests.

Here are some manual cmds that cover all testing features:
```bash
# Run these cmds in the nix-bitcoin src root

test/run-tests.sh basic
test/run-tests.sh buildable
test/run-tests.sh all

test/run-tests.sh -s default
test/run-tests.sh -s clightning # adhoc scenario
test/run-tests.sh -s clightningReplication # regular NixOS test
test/run-tests.sh -s default eval
test/run-tests.sh -s default vm
test/run-tests.sh -s default run
test/run-tests.sh -s default debug
test/run-tests.sh -s default container
test/run-tests.sh -s default container --run c nodeinfo

test/run-tests.sh -s "{ nix-bitcoin.nodeinfo.enable = true; }"
test/run-tests.sh -s "{ nix-bitcoin.nodeinfo.enable = true; }" container --run c nodeinfo

## Running tests via the nix CLI
# Equivalent to `test/run-tests.sh -s default`
nix build --no-link --print-out-paths .#tests.default

# Equivalent to `test/run-tests.sh -s default run`
nix run .#tests.default.run

# Equivalent to `test/run-tests.sh -s default debug`
nix run .#tests.default.run -- --debug

# Equivalent to `test/run-tests.sh -s default vm`
nix run .#tests.default.vm

# Equivalent to `test/run-tests.sh -s default container`
nix run .#tests.default.container
# Equivalent to `test/run-tests.sh -s default container --run c nodeinfo`
nix run .#tests.default.container -- --run c nodeinfo

# Use the following cmds on NixOS with `system.stateVersion` <22.05
nix run .#tests.default.containerLegacy
nix run .#tests.default.containerLegacy -- --run c nodeinfo
```
Fixes #554


erikarvstedt:
Rebased.

ekimber:
The example extension flake is really helpful, however, I believe what I have been trying to do isn't possible as an extension. I would like to add a flake I made for btc-rpc-proxy and use that with a pruned bitcoind in core lightning. Or, perhaps I would like to use a different bitcoin implementation entirely, such as btcd. From this point of view it would be better to have entirely separate flakes for core lightning, lnd, etc. but I guess this just isn't what the purpose that nix-bitcoin is providing. It's an opinionated stack. However, I would still like to reuse a lot of the configuration from nix-bitcoin!

erikarvstedt:
What's the advantage of btc-rpc-proxy over the native bitcoind RPC user accounts?

You can replace the bitcoind service with a custom implementation like so:
- Add `disabledModules = [ (nix-bitcoin-flake + "modules/bitcoind.nix") ];` to your config
- Add a custom bitcoind module that defines all interface options that are used by other services (like `services.bitcoind.address`).

ekimber:
Thanks, I'll try that.

The proxy does peer block fetching the same way lnd does it, so you can use a pruned node with core lightning

erikarvstedt:
So you're running bitcoind and `btc-rpc-proxy` on the same node. Then it's best to just replace `clightning.nix` with your own patched version.

erikarvstedt:
So you're running bitcoind and `btc-rpc-proxy` on the same node. Then you can simply override bitcoind-related options:
```nix
services.clightning.extraConfig = ''
  bitcoin-datadir=...
  bitcoin-rpcconnect=...
  bitcoin-rpcport=...
  bitcoin-rpcuser=...
'';
```

jonasnick (commented):
Very nice! Works well for me except

```
❯ ./run-tests.sh -s clightningReplication
error: The option `services.clightningReplication' does not exist. Definition values:
       - In `/nix/store/ml0shjsrv72lsa3l43aci55wxpc18bpk-source/nixos/lib/build-vms.nix':
           {
             enable = true;
           }
(use '--show-trace' to show detailed location information)
```

which presumably happens because `getTest` creates a test for a scenario, but doesn't pick up the existing clightningReplication test.

erikarvstedt:
Thanks, nice catches!
I've updated the test cmds in the PR description.

jonasnick (approved):
ACK edbaeb98134a2059e8bb063ecd6fd16c0aa54f5d

PR #569, opened 2022-11-08
treewide: set shebang for bash scripts

erikarvstedt:
#### Copy of commit msg
These scripts previously failed when called with syscalls like `execve` (used by, e.g., Python's `subprocess.run`) that use no default interpreter for scripts without a shebang.


dpc (approved):

jonasnick (approved):
ACK 0447c5bacbe4c7d431681b480997651e2f43d0d1

PR #571, opened 2022-11-11
lnd: support `INADDR_ANY` addresses for `bitcoind.zmqpubraw*`

erikarvstedt:
Use the following for manual testing. lnd currently doesn't support IPv6 zmq sockets (but bitcoind does), so only test IPv4 here:
```bash
run-tests.sh -s '{
  services.lnd.enable = true;
  services.bitcoind = {
    zmqpubrawblock = "tcp://0.0.0.0:12345";
    zmqpubrawtx =  "tcp://0.0.0.0:12346";
  };
}' container --run c bash -c 'cat /var/lib/lnd/lnd.conf && systemctl status lnd'
```

### DIFF
--- a/modules/lnd.nix
+++ b/modules/lnd.nix
@@ -175,13 +175,15 @@ let
 
     bitcoind.rpchost=${bitcoindRpcAddress}:${toString bitcoind.rpc.port}
     bitcoind.rpcuser=${bitcoind.rpc.users.${rpcUser}.name}
-    bitcoind.zmqpubrawblock=${bitcoind.zmqpubrawblock}
-    bitcoind.zmqpubrawtx=${bitcoind.zmqpubrawtx}
+    bitcoind.zmqpubrawblock=${zmqHandleSpecialAddress bitcoind.zmqpubrawblock}
+    bitcoind.zmqpubrawtx=${zmqHandleSpecialAddress bitcoind.zmqpubrawtx}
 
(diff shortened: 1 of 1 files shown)


modules/lnd.nix
```diff
@@ -175,14 +175,16 @@ let
...
-    bitcoind.zmqpubrawtx=${bitcoind.zmqpubrawtx}
+    bitcoind.zmqpubrawblock=${zmqHandleSpecialAddress bitcoind.zmqpubrawblock}
+    bitcoind.zmqpubrawtx=${zmqHandleSpecialAddress bitcoind.zmqpubrawtx}
 
     wallet-unlock-password-file=${secretsDir}/lnd-wallet-password
 
     ${cfg.extraConfig}
   '';
 
+  zmqHandleSpecialAddress = builtins.replaceStrings [ "0.0.0.0" "[::]" ] [ "127.0.0.1" "[::1]" ];
```
jonasnick, line 186: can't we just use `nbLib.address` ?

jonasnick:
Concept ACK

modules/lnd.nix
erikarvstedt: `nbLib.address` only handles IP addresses, while `zmqpub*` has format `tcp://<address>:28333`

jonasnick (approved):
ACK c5493717b71ea23ddaca7249c9d77e38ea20056a

PR #574, opened 2022-12-12
Add dev helper and docs

### LINKED ISSUE #557: Add developer docs
Document dev tooling and best practices for developing and debugging nix-bitcoin.

erikarvstedt:
Fixes #557.

### DIFF
--- a/dev/README.md
+++ b/dev/README.md
@@ -1,0 +1,104 @@
+This directory contains docs and helper scripts for developing and debugging:
+
+- [`dev.sh`](./dev.sh): misc dev helpers
+- [`dev-features.sh`](./dev-features.sh): helpers for developing specific
+  nix-bitcoin features, like services
+- [`topics`](./topics) features specific topics
+- [`dev-scenarios.nix`](./dev-scenarios.nix): extra test scenarios used in the above scripts
+
+See also: [test/README.md](../test/README.md)
+
+## Run a dev shell
+
+There are two ways to run a dev shell:
+
+### 1. Run command `nix develop`
+
+This starts a shell with [`test/run-tests.sh`](../test/run-tests.sh) and
+the scripts in dir [`helper`](../helper) added to `PATH`.
+
+### 2. Setup and start the `direnv` dev env
+
+This is an opinionated, [direnv](https://direnv.net/)-based dev env, optimized for developer experience.
+
+[`dev-env/create.sh`](./dev-env/create.sh) creates a git repo with the following contents:
+- Dir `src` which contains the nix-bitcoin repo
+- Dir `bin` for helper scripts
+- File `scenarios.nix` for custom test scenarios
+- File `.envrc` that defines a [direnv](https://direnv.net/) environment,
+  mainly for adding nix-bitcoin and helper scripts to `PATH`
+
+#### Installation
+
+1. [Install direnv](https://direnv.net/docs/installation.html).\
+   If you use NixOS (and Bash as the default shell), just add the following to your system config:
+   ```nix
+     environment.systemPackages = [ pkgs.direnv ];
+     programs.bash.interactiveShellInit = ''
+       eval "$(direnv hook bash)"
+     '';
+   ```
+
+2. Create the dev env:
+   ```bash
+   # Set up a dev environment in dir ~/dev/nix-bitcoin.
+   # The dir is created automatically.
+   ./dev-env/create.sh ~/dev/nix-bitcoin
+
+   cd ~/dev/nix-bitcoin
+
+   # Enable direnv
+   direnv allow
+   ```
+
+3. Optional: Editor integration
+   - Add envrc support to your editor
+   - Setup your editor so you can easily execute lines or paragraphs from a shell script
+     file in a shell.\
+     This simplifies using dev helper scripts like [`./dev.sh`](./dev.sh).
+
+#### Explore the dev env
+```bash
+# The direnv is automatically activated when visiting any subdir of ~/dev/nix-bitcoin
+cd ~/dev/nix-bitcoin
+
+ls -al . bin lib
+
+# The direnv config file
+cat .envrc
+
+# You can use this file to define extra scenarios
+cat scenarios.nix
+
+# Binary `dev-run-tests` runs nix-bitcoin's `run-tests.sh` with extra scenarios from ./scenarios.nix
+# Example:
+# Run command `nodeinfo` in `myscenario` (defined in ./scenarios.nix) via a container
+dev-run-tests -s myscenario container --run c nodeinfo
+
+# Equivalent (shorthand)
+te -s myscenario container --run c nodeinfo
+
+# Run the tests for `myscenario` in a VM
+te -s myscenario
+
+# Start an interactive shell inside a VM
+te -s myscenario vm
+```
+
+See also: [test/README.md](../test/README.md)
+
+## Adding a new service
+
+It's easiest to use an existing service as a template:
+- [electrs.nix](../modules/electrs.nix): a basic service
+- [clightning.nix](../modules/clightning.nix): simple, but covers a few more features.\
+  (A `cli` binary and a runtime-composed config to include secrets.)
+- [rtl.nix](../modules/rtl.nix): includes a custom package, defined in [pkgs/rtl](../pkgs/rtl).\
+  Most other services use packages that are already included in nixpkgs.
+
+## Switching to a new NixOS release
+
+- [flake.nix](../flake.nix): update `nixpkgs.url`
+- [cirrus.yml](../.cirrus.yml): update toplevel container -> image attribute
+- [examples/configuration.nix](../examples/configuration.nix): update `system.stateVersion`
+- Treewide: check if any `TODO-EXTERNAL` comments can be resolved
--- a/dev/dev-env/create.sh
+++ b/dev/dev-env/create.sh
@@ -1,0 +1,65 @@
+#!/usr/bin/env bash
+# shellcheck disable=SC2016
+set -euo pipefail
+
+destDir=${1:-nix-bitcoin}
+
+scriptDir=$(cd "${BASH_SOURCE[0]%/*}" && pwd)
+
+mkdir -p "$destDir/"{bin,lib}
+cd "$destDir"
+
+if [[ ! -e src ]]; then
+    echo "Cloning fort-nix/nix-bitcoin"
+    git clone https://github.com/fort-nix/nix-bitcoin src
+fi
+
+echo 'export root=$PWD
+export src=$root/src
+PATH_add bin
+PATH_add src/helper' > .envrc
+
+if [[ ! -e scenarios.nix ]]; then
+    cp "$scriptDir/template-scenarios.nix" scenarios.nix
+fi
+
+install -m 755 <(
+    echo '#!/usr/bin/env bash'
+    echo 'exec run-tests.sh --extra-scenarios "$root/scenarios.nix" "$@"'
+) bin/dev-run-tests
+
+install -m 755 <(
+    echo '#!/usr/bin/env bash'
+    echo 'exec $root/src/test/run-tests.sh --out-link-prefix /tmp/nix-bitcoin/test "$@"'
+) bin/run-tests.sh
+
+ln -sfn dev-run-tests bin/te
+
+## nix-bitcoin-firejail
+
+echo '# Add your shell config files here that should be accessible in the sandbox
+whitelist ${HOME}/.bashrc
+read-only ${HOME}/.bashrc' > lib/nix-bitcoin-firejail.conf
+
+install -m 755 <(
+    echo '#!/usr/bin/env bash'
+    echo '# A sandbox for running shells/binaries in an isolated environment:'
+    echo '# - The sandbox user is the calling user, with all capabilities dropped'
+    echo '#   and with no way to gain new privileges (e.g. via `sudo`).'
+    echo '# - $HOME is bind-mounted to a dir that only contains shell config files and files required by direnv.'
+    echo '#'
+    echo '# You can modify the firejail env by editing `lib/nix-bitcoin-firejail.conf` in your dev env dir.'
+    echo 'exec firejail --profile="$root/lib/nix-bitcoin-firejail.conf" --profile="$root/src/dev/dev-env/nix-bitcoin-firejail.conf" "$@"'
+) bin/nix-bitcoin-firejail
+
+echo "1" > lib/dev-env-version
+
+## git
+
+echo '/src' > .gitignore
+
+if [[ ! -e .git ]]; then
+    git init
+    git add .
+    git commit -a -m init
+fi
--- a/dev/dev-env/dev-shell.nix
+++ b/dev/dev-env/dev-shell.nix
@@ -1,0 +1,16 @@
+pkgs:
+
+pkgs.mkShell {
+  shellHook = ''
+    # A known rev from the master branch to test whether `nix develop`
+    # is called inside the nix-bitcoin repo
+    rev=5cafafd02777919c10e559b5686237fdefe920c2
+    if git cat-file -e $rev &>/dev/null; then
+      root=$(git rev-parse --show-toplevel)
+      export PATH=$root/test:$root/helper:$PATH
+    else
+      echo 'Error: `nix develop` must be called inside the nix-bitcoin repo.'
+      exit 1
+    fi
+  '';
+}
--- a/dev/dev-env/template-scenarios.nix
+++ b/dev/dev-env/template-scenarios.nix
@@ -1,0 +1,20 @@
+{ pkgs, lib, scenarios, nix-bitcoin }:
+with lib;
+rec {
+  # For more examples, see `scenarios` and `exampleScenarios` in ./src/test/tests.nix
+
+  template = { config, pkgs, lib, ... }: {
+    imports = [
+      (nix-bitcoin + "/modules/presets/secure-node.nix")
+      scenarios.netnsBase
+      scenarios.regtestBase
+    ];
+    test.container.enableWAN = true;
+    test.container.exposeLocalhost = true;
+  };
+
+  myscenario = { config, pkgs, lib, ... }: {
+    services.clightning.enable = true;
+    nix-bitcoin.nodeinfo.enable = true;
+  };
+}
--- a/dev/dev-features.sh
+++ b/dev/dev-features.sh
@@ -1,0 +1,290 @@
+# shellcheck disable=SC2086,SC2154
+
+#―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
+# Run tests
+# See ../test/README.md for a tutorial
+# and ../test/run-tests.sh for a complete documentation
+
+# Start a shell in a container
+run-tests.sh -s electrs container
+
+# Run a command in a container.
+# The container is deleted afterwards.
+run-tests.sh -s electrs container --run c journalctl -u electrs
+
+# Run a bash command
+run-tests.sh -s bitcoind container --run c bash -c "sleep 1; journalctl -u bitcoind"
+
+run-tests.sh -s '{
+  imports = [ scenarios.regtestBase ];
+  services.electrs.enable = true;
+}' container --run c journalctl -u electrs
+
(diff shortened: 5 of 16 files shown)


dev/create-dev-env.sh
```diff
@@ -0,0 +1,102 @@
...
+set -euo pipefail
+
+destdir=${1:-nix-bitcoin}
+
+mkdir -p "$destdir/bin"
+cd "$destdir"
+
+if [[ ! -e src ]]; then
+    echo "Cloning fort-nix/nix-bitcoin"
+    git clone https://github.com/fort-nix/nix-bitcoin src
```
dpc, line 12: After cloning, don't you want to `cd` into it?

It seems to me that all this cloning (or not) etc. might just confuse people. Generally I would expect people who run it to already have `gtt clone`d the repo manually. And then having that as pre-condition can be used to double check if the CWD is correct etc.

erikarvstedt: > After cloning, don't you want to cd into it?

No, the setup script simply clones the nix-bitcoin repo to dir `src`, without later changing its contents.

> Generally I would expect people who run it to already have `git cloned` the repo manually

Yes, but if the script moved/renamed the user's existing files, it would be even more confusing.
Users can simply replace the cloned repo with their local clone.

The dev env layout is very opinionated; it's what I prefer for creating auxiliary scripts/notes/resources for a public repo.
But I'm very interested in suggestions on how to improve it.

dev/create-dev-env.sh
```diff
@@ -0,0 +1,102 @@
...
+  };
+
+  myscenario = { config, pkgs, lib, ... }: {
+    services.clightning.enable = true;
+    nix-bitcoin.nodeinfo.enable = true;
+  };
+}' >> scenarios.nix
+fi
+
+install -m 755 <(
```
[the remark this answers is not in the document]
erikarvstedt: Reasons against linking:
Users should be able to easily modify these files.
E.g., the firejail config must be edited when a shell config is used that includes files other than `~/.bashrc`.
Same goes for `bin/run-tests.sh`, where users might want to edit the `--out-link-prefix` (e.g., I use a persistent (i.e. not in `/tmp`) global gcroots dir for all sorts of nix builds).

But yeah, we could move some of these inline defs to separate files.

dev/README.md
```diff
@@ -0,0 +1,84 @@
...
+- A `.envrc` file that defines a [direnv](https://direnv.net/) environment,
+  mainly for adding nix-bitcoin and helper scripts to `PATH`
+
+### Installation
+
+1. [Install direnv](https://direnv.net/docs/installation.html).\
+   If you use NixOS (and Bash as the default shell), just add the following to your system config:
+   ```nix
+     environment.systemPackages = [ pkgs.direnv ];
+     programs.bash.interactiveShellInit = ''
```
dpc, line 23: What's this for? In fedimint we just add [interactive bash to the dev shell to fix some issues with non-interactive bash](https://github.com/fedimint/fedimint/blob/4dca27643e673eba166d10ebe4cdbaa4f3fff888/flake.nix#L816)

dev/create-dev-env.sh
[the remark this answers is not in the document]
erikarvstedt: I've extracted some of the inline defs.
Also, the firejail config is now split up into a personal config in the dev-env dir and a static config in the nix-bitcoin repo.
Thanks for the pointer.

dev/create-dev-env.sh
dpc: > The dev env layout is very opinionated; it's what I prefer for creating auxiliary scripts/notes/resources for a public repo.
But I'm very interested in suggestions on how to improve it.

Just thinking aloud, I'm not even all that familiar with nix-bitcoin, or generally developing strictly Nix-focused project (as opposed to a Rust project, that just uses Nix for building and tooling).

Generally if the project includes it's own tests and dev-env integration, I would expect `git clone <project-repo> && cd <project> && nix develop (or direnv)` to just set up everything for me (add binaries to the $PATH, create some repo-scoped configs and so on).

dev/README.md
erikarvstedt: This installs direnv on a NixOS system.
direnv automatically sets up env vars when browsing to a project dir in a shell or when opening a project file in an editor (when editor support is installed).
The nix-bitcoin dev env exclusively uses direnv and no nix-shell.

dev/create-dev-env.sh
erikarvstedt: > Just thinking aloud

Highly appreciated, thanks!

> create some repo-scoped configs

Yeah, like our `scenarios.nix` file. How exactly would you implement this? Where would these files be created? (Note that they must be placed in a separate repo).

erikarvstedt: Note that you can develop nix-bitcoin just fine by simply running `test/run-tests.sh` and editing `tests.nix`.
I'll mention this in the docs.

dpc: > Yeah, like our scenarios.nix file. How exactly would you implement this? Where would these files be created? (Note that they must be placed in a separate repo).

Oh, I don't really understand how these works, but generally for files like these it's "if they don't exist, copy/ln from a template of some kind and have them in `.gitignore`". [Like here](https://github.com/fedimint/fedimint/blob/4dca27643e673eba166d10ebe4cdbaa4f3fff888/flake.nix#L832)

The "must be placed in a separate repo" seems quite confusing to me, but I wouldn't spend too much time explaining it to me. :D

dev/README.md
dpc: Oh, I see. I got confused.

BTW. `direnv` is nice and all and I do use it personally, but in projects using Nix, I generally ask people "to call `nix develop` and *optionally* setup direnv". Any advantage of hard direnv requirement?

erikarvstedt: Yes:
- For our env, we just [set a few env vars](https://github.com/erikarvstedt/nix-bitcoin/blob/a7c31404ae4bfdb61a519ff9a73fcd0c800818f8/dev/dev-env/create.sh#L15) that simply point to already existing filesystem locations. No derivations need to be built, so we don't need any complex nix machinery.
- `nix develop` doesn't have access to the PWD, so for setting up repo-relative paths it would only work when called from the repo root. 
  This is bad UX. I simply want to pop up a shell anywhere in the repo and have my environment ready.
Actually, this can be worked around by using `shellHook`. In the nixbitcoin.org repo, which has an extensive `nix develop` shell, we [use this trick](https://github.com/fort-nix/nixbitcoin.org/blob/c2eb4fad9408a15520916dffb4d33ad2ed6843a2/lib/dev-env.nix#L21) to set env var `root` to point to the repo root, regardless of where `nix develop` is called.
Still, the problem that `nix develop` has to be called to enter the dev env remains.

erikarvstedt: I've edited the previous post.

erikarvstedt: I'll add `nix develop` support which just puts `helper` and `test/run-tests.sh` in PATH.

dpc: > Users should be able to simply pop up a shell anywhere in the repo and have their environment ready.

But with direnv one still needs to `direnv allow` which is kind of the same as `nix develop`. And having to setup `direnv` is an extra step.

> For our env, we just [set a few env vars](https://github.com/erikarvstedt/nix-bitcoin/blob/a7c31404ae4bfdb61a519ff9a73fcd0c800818f8/dev/dev-env/create.sh#L15) that simply point to already existing filesystem locations.

`shellHook` can do that all the same, no?

dev/create-dev-env.sh
erikarvstedt: > must be placed in a separate repo

Sorry, that was misleading. What I meant is that files like `scenarios.nix` should be put under version control, but, obviously, in an extra repo. This extra repo could simply be located inside the nix-bitcoin repo, but this has some downsides, as mentioned above.

dev/dev-env/create.sh
```diff
@@ -0,0 +1,56 @@
...
+    git clone https://github.com/fort-nix/nix-bitcoin src
+fi
+
+echo 'export root=$PWD
+export src=$root/src
+PATH_add bin
+PATH_add src/helper' > .envrc
+
+if [[ ! -e scenarios.nix ]]; then
+    cp "${BASH_SOURCE[0]%/*}/template-scenarios.nix" scenarios.nix
```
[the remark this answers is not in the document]
erikarvstedt: I can't repro this. What's the sequence of cmds to trigger this error?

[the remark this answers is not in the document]
jonasnick:
```
❯ git checkout upstream-pull/574/head && cd dev/dev-env && ./create.sh ../../nix-bitcoin-dev
HEAD is now at 3d74fb52 fixup! add reference to `tests.nix` in `template-scenarios.nix`
Cloning fort-nix/nix-bitcoin
Cloning into 'src'...
remote: Enumerating objects: 8666, done.
remote: Counting objects: 100% (338/338), done.
remote: Compressing objects: 100% (169/169), done.
remote: Total 8666 (delta 199), reused 286 (delta 167), pack-reused 8328
Receiving objects: 100% (8666/8666), 3.18 MiB | 2.73 MiB/s, done.
Resolving deltas: 100% (5790/5790), done.
cp: cannot stat './template-scenarios.nix': No such file or directory
```

same if I don't provide an argument to create.sh.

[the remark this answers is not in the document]
erikarvstedt: Thanks, fixed. The script used `BASH_SOURCE`, which can be a relative path, after changing the working dir.

dev/dev-env/dev-shell.nix
```diff
@@ -0,0 +1,15 @@
+pkgs:
+
+pkgs.mkShell {
+  shellHook = ''
+    # A known rev from the master branch
```
jonasnick, line 5: Maybe "A known rev from the master branch to test whether `nix develop` is called inside the nix-bitcoin repo". I was confused for a moment why we hardcode a rev.

dev/dev.sh
```diff
@@ -0,0 +1,126 @@
+#―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
+# Access nix-bitcoin flake packages
+
+nb() {
+    nix build --no-link --print-out-paths --print-build-logs "$@"
+}
+
+# A package defined by nix-bitcoin
+nb .#joinmarket
```
jonasnick, line 9: Should we run (parts of) the new shell scripts in ci?

dev/dev-env/create.sh
```diff
@@ -0,0 +1,58 @@
...
+    echo 'exec $root/src/test/run-tests.sh --out-link-prefix /tmp/nix-bitcoin/test "$@"'
+) bin/run-tests.sh
+
+ln -sfn dev-run-tests bin/te
+
+## nix-bitcoin-firejail
+
+echo '# Add your shell config files here that should be accessible in the sandbox
+whitelist ${HOME}/.bashrc
+read-only ${HOME}/.bashrc' > nix-bitcoin-firejail.conf
```
jonasnick, line 42: Could you document somewhere what constraints this firejail puts on a process and what kind of binaries can or cannot run in the jail?

dev/dev-env/dev-shell.nix
```diff
@@ -0,0 +1,15 @@
+pkgs:
+
+pkgs.mkShell {
```
jonasnick, line 3: it would be more logical for me to put this file in `/dev` because it has nothing to do with the "dev-env" that's created with `create.sh`.

dev/dev-env/create.sh
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env bash
+# shellcheck disable=SC2016
+set -euo pipefail
+
+destDir=${1:-nix-bitcoin}
```
jonasnick, line 5: If we want to change anything about the dev-env (add, delete or update files), how does a dev with an existing dev-env upgrade it to the latest version?

dev/dev-features.sh
```diff
@@ -0,0 +1,275 @@
...
+c journalctl -u joinmarket
+
+# Manually wait for string 'started http server, visit http://127.0.0.1:62601/'
+# This can take a while
+c journalctl -f -u joinmarket-ob-watcher
+
+# Check webinterface
+c curl localhost:62601
+nix run --inputs-from . nixpkgs#lynx -- --dump $ip:62601
+c curl -s localhost:62601 | g "orders found"
```
jonasnick, line 271: Is `g` an alias for `grep`?

dev/dev-features.sh
```diff
@@ -0,0 +1,275 @@
...
+# Analyze container boot performance
+c systemd-analyze critical-chain
+
+# Listening TCP sockets
+c netstat -nltp
+# Listening sockets
+c netstat -nlp
+
+# The container root filesystem
+ls -al /var/lib/containers/nb-test
```
jonasnick, line 42: Doesn't exist on my machine: `ls: cannot access '/var/lib/containers/nb-test': No such file or directory`

dev/dev-features.sh
```diff
@@ -0,0 +1,275 @@
...
+
+run-tests.sh -s jm-ob-watcher container
+
+c systemctl status joinmarket-ob-watcher
+c journalctl -u joinmarket-ob-watcher
+c journalctl -u joinmarket
+
+# Manually wait for string 'started http server, visit http://127.0.0.1:62601/'
+# This can take a while
+c journalctl -f -u joinmarket-ob-watcher
```
jonasnick, line 266: This never appears for me, instead
```
Dec 17 22:36:55 nb-test jm-ob-watcher[488]: 2022-12-17 22:36:55,369 [ERROR]  We failed to connect and handshake with ANY directories; onion messaging is not functioning.
```

dev/dev-env/template-scenarios.nix
```diff
@@ -0,0 +1,20 @@
+{ pkgs, lib, scenarios, nix-bitcoin }:
+with lib;
+rec {
+  # For more examples, see `scenarios` and `exampleScenarios` in ./src/test/tests.nix
```
jonasnick, line 4: Also see `dev-scenarios`?

In general, I was quite confused about the various locations of the scenarios. We have
- example scenarios, which are somewhat documenting what's happening in tests.nix
- we have `template-scenarios.nix` (aka scenarios.nix), which is a file supposed to be changed by a dev user
- we have `dev-scenarios.nix`, which are our collection of scenarios for users (although the line to `dev-features.sh` is quite blurry)

It seems to all make sense, but is still quite confusing. I wonder if there was a way to somehow simplify this.

dev/dev-features.sh
```diff
@@ -0,0 +1,275 @@
...
+post users '{"email": "a@a.a", "password": "aaaaaa", "isAdministrator": true}'
+
+# Login with:
+# user: a@a.a
+# password: aaaaaa
+runuser -u "$(logname)" -- xdg-open http://$ip:23000
+
+# create store
+post stores '{"name": "a", "defaultPaymentMethod": "BTC_LightningNetwork"}'
+post stores '{"name": "a", "defaultPaymentMethod": "BTC"}'
```
jonasnick, line 216: I'm a bit concerned that all the API calls in this file (and manual builds) are gonna break frequently and silently (since it's not in CI). Are the files in the `dev/` directory more like "notes" in the sense that they are supposed to give hints, but may not work correctly all the time?

jonasnick (commented):
Lots of cool stuff. Thanks! Left a few comments.

dev/dev.sh
erikarvstedt: In the past, I've failed to directly run extra-container on the Cirrus CI node. I'll investigate this further and maybe later add tests for the dev scripts.

dev/dev-env/create.sh
erikarvstedt: Done.

dev/dev-env/dev-shell.nix
erikarvstedt: Done.

dev/dev-env/create.sh
erikarvstedt: The user can re-run the `create.sh` script. No harm is done when user modifications are overwritten, because the dev env is under version control.
To allow future automatic updates, I've added file `lib/dev-env-version`.

dev/dev-features.sh
erikarvstedt: Yes. Fixed.

dev/dev-features.sh
erikarvstedt: Fixed.

dev/dev-features.sh
erikarvstedt: Tor is under heavy load right now, so this took 11 minutes for me.
But I've seen this finishing in under a minute in the past.
I've amended the comments to explain this.

dev/dev-env/template-scenarios.nix
erikarvstedt: A user checking the definition of `scenarios` (tests.nix) will see that it includes `dev-scenarios.nix`.

- `template-scenarios.nix` must be a separate file because it's part of the user's dev env.
- We need `dev-scenarios.nix` because defining (and editing) complex scenarios inline as a bash string in `dev-features.sh` is bad UX.
   We could move `dev-scenarios.nix` into `tests.nix`, but this adds unneeded complexity to `tests.nix`. I don't see how we could improve the file layout.

dev/dev-features.sh
erikarvstedt: > they are supposed to give hints, but may not work correctly all the time

Yes. But I'm often using these snippets while developing, so I'll keep them up to date.
As for the RTL manual build and the btcp API calls, these have been working, unaltered, for a long time.

erikarvstedt:
Rebased to master.

erikarvstedt:
Rebased to master.

dev/topics/rtl.sh
```diff
@@ -0,0 +1,43 @@
...
+#―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
+# Build RTL manually
+# [[~/s/RTL/dockerfiles/Dockerfile][dockerfile]]
+# [[~/s/RTL/package.json][package.json]]
+
+git clone https://github.com/Ride-The-Lightning/RTL ~/s/RTL
+
+nix build -o /tmp/nix-bitcoin-dev/nodejs --inputs-from . nixpkgs#nodejs-16_x
+# Start a shell in a sandbox
+(cd ~/s/RTL && nix-bitcoin-firejail --whitelist="$(pwd)")
```
jonasnick, line 34:
```
$ (cd ~/tmp/RTL && nix-bitcoin-firejail --whitelist="$(pwd)")
direnv: unloading
nix-bitcoin-firejail: command not found
```

dev/dev-env/template-scenarios.nix
jonasnick: maybe mention dev-scenarios.nix and its purpose in the dev readme?

dev/dev-env/create.sh
jonasnick: That the dev env is under version control is a good point. Also, good idea to add the `dev-env-version`.

dev/dev.sh
```diff
@@ -0,0 +1,128 @@
...
+nix eval --inputs-from . nixpkgs#electrs.version
+nix eval --inputs-from . nixpkgs-unstable#electrs.version
+
+#―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――――
+# Inspect test systems
+
+# Build a test system
+nix build -o /tmp/system --print-build-logs "$(nix eval --raw .#tests.default --apply '
+  test: test.nodes.machine.config.system.build.toplevel
+')"
```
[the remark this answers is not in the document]
erikarvstedt: Fixed in `fixup! fix accessing test machine configs`

dev/topics/rtl.sh
erikarvstedt: Hmm, it works for me.
Hopefully fixed in `fixup! topics/rtl: fix sandbox starting`.

jonasnick: Works for me now too. Thanks.

jonasnick:
LGTM now. Feel free to squash.

jonasnick (approved):
ACK b4d7e1aa8f4ed6db1da1486c95016f266731be03

PR #575, opened 2022-12-12
Improve `lndconnect`, support WireGuard

### LINKED ISSUE #555: Add clearnet support for `lndconnect`
Allow using lndconnect via clearnet (including wireguard/tailscale).

erikarvstedt:
Fixes #555


erikarvstedt:
Rebased to https://github.com/fort-nix/nix-bitcoin/pull/586.

erikarvstedt:
Rebased to master.

modules/lndconnect.nix
```diff
@@ -0,0 +1,202 @@
...
+  # TODO-EXTERNAL:
+  # lndconnect requires a --configfile argument, although it's unused
+  # https://github.com/LN-Zap/lndconnect/issues/25
+  pkgs.hiPrio (pkgs.writeScriptBin name ''
+    ${shebang}
+    url=$(
+      ${getExe config.nix-bitcoin.pkgs.lndconnect} --url \
+        ${optionalString enableOnion "--host=$(cat ${config.nix-bitcoin.onionAddresses.dataDir}/${onionService})"} \
+        --port=${toString port} \
+        ${if enableOnion || certPath == null then "--nocert" else "--tlscertpath='${certPath}'"} \
```
jonasnick, line 101: It seems like `certPath` is always non-null. Can we remove this "--nocert" branch?

dev/dev-scenarios.nix
```diff
@@ -45,4 +45,34 @@ with lib;
...
+
+    services.clightning.extraConfig = "disable-dns";
+
+    services.lnd = {
+      enable = true;
+      lndconnect = {
+        enable = true;
+        onion = true;
+      };
+    };
```
jonasnick, line 68: Shouldn't I expect to see an lnd-rest entry in nodeinfo?
```
dev-run-tests -s wireguard-lndconnect-online container --run c nodeinfo
Building container
/nix/store/ncgfyk9r7hqxnxvs8k2hzsc716v16b3l-container

Installing containers:
nb-test

Starting containers:
nb-test

Running command.
{
  "bitcoind": {
    "local_address": "127.0.0.1:8333"
  },
  "clightning": {
    "local_address": "127.0.0.1:9735",
    "nodeid": "030cb0ec1c10551a0527187af5bbf83afdee5eefde404aca85c18364e3a902609d"
  },
  "clightning-rest": {
    "local_address": "127.0.0.1:3001",
    "onion_address": "n3ahtp4ariczami6tllmpoofulmf43efxtfm6d4tewqf3h7qeekyrjid.onion:3001"
  },
  "lnd": {
    "local_address": "localhost:9736",
    "onion_address": "zrdmpcmxfwsrc2ggjxqdqizp72ivxvno3jca33hyxadxmq7gykvcrxyd.onion:9736",
    "nodeid": "03e01e8ed3810d5a50958392de464986994e4f29a08d2e7f80ef134f020bd81a08"
  }
}
```

docs/services.md
```diff
@@ -142,60 +142,153 @@ You can find the `<onion-address>` with command `nodeinfo`.
...
+
+   # For clightning
+   services.clightning-rest = {
+     enable = true;
+     lndconnect.enable = true;
+   };
+   ```
+3. Deploy your configuration.
+
+4. Add the following port forwarding rule to your node's external firewall:
```
jonasnick, line 239: This is only necessary if the node is behind NAT or a firewall which is regularly not the case in nix-bitcoin deployments. Therefore I was unsure at first if "node's external firewall" meant externally facing firewall on the nix-bitcoin node (vs. the internal netns firewall). Maybe we can say something like:

```
If your node is behind a firewall or NAT, add the following port forwarding rule to the external device:
```

jonasnick (commented):
Concept ACK

modules/lndconnect.nix
erikarvstedt: Fn `mkLndconnect` implements generic cert handling because it is exposed as option `nix-bitcoin.mkLndconnect".
If we remove this option we can remove generic cert handling. What do you prefer?

dev/dev-scenarios.nix
erikarvstedt: Fixed in `nodeinfo/lnd: add `rest_address``.

jonasnick: Shouldn't this also output the onion (if configured)?

modules/lndconnect.nix
jonasnick: Ok then let's keep it as is.

dev/dev-scenarios.nix
erikarvstedt: Fixed in `nodeinfo/lnd: add onion_rest_address`.

jonasnick (commented):
ACK cea69b73d279a3ab4506fc972045a4398d66ac97

jonasnick (approved):

PR #577, opened 2022-12-18
Upgrade to NixOS 22.11

erikarvstedt:
This is currently deployed on https://nixbitcoin.org, via https://github.com/fort-nix/nixbitcoin.org/pull/51.

PR #579, opened 2022-12-22
Add `agenix` support

erikarvstedt:
[agenix](https://github.com/ryantm/agenix) is a popular secrets deployment scheme where [age](age-encryption.org)-encrypted secrets are committed to the system repo and deployed as part as of the system derivation.

The workflow for nix-bitcoin users:
- Include the nix-bitcoin `age` module
- Configure the `age` module
- `nix run` a package in a nix-bitcoin node flake to auto-generate encrypted secrets
- Commit the secrets and deploy

This PR is based on https://github.com/fort-nix/nix-bitcoin/pull/577.
First new commit: "secrets: expose `generateSecretsScript`"

Current status: draft. This is hot off the press and hasn't seen a thorough review.


seberm:
Hello @erikarvstedt,
do you still plan to add the agenix support?

jpentland:
So the goal here is to generate all secrets during the configuration process, and then commit them encrypted, rather than at runtime as it is currently?

erikarvstedt:
Yes, the goal is that secrets are part of the node config repo, rather than being randomly generated on the node system (via `nix-bitcoin.generateSecrets = true`).

This approach is already available when deploying via krops. Agenix however works with all deployment methods and with flakes.

PR #553, opened 2022-10-10
pkgs: add lnd 0.15.2

erikarvstedt:
#### Copy of commit msg
Includes an emergency hotfix:
https://github.com/lightningnetwork/lnd/releases/tag/v0.15.2-beta

PR #583, opened 2023-01-11
cln: add alternative bitcoin backend config option

elsirion:
Getting a CLN node up and running in pruned mode is annoying and seems impossible when using https://prunednode.today/. It appears the easiest way to bootstrap a CLN node in such an environment is using an alternative chain data plugin like [`trustedcoin`](https://github.com/nbd-wtf/trustedcoin) till CLN has processed all relevant blocks and then switching over to the pruned `bitcoind` again.

With this change users will be able to activate such a plugin with the `alternativeBitcoinBackend` option. If someone ever gets around to properly packaging `trustedcoin` we could make it a normal plugin (that also removes the `bitcoin-datadir=${config.services.bitcoind.dataDir}` line when active). The current approach is a bit more flexible and you could e.g. write and use your own version of `trustedcoin` with it.
```nix
    clightning = {
      enable = true;
      alternativeBitcoinBackend = "${trustedcoin}/usr/bin/trustedcoin";
    };
```

Properly packaging `trustedcoin` seemed too much work since I don't know go that well, I'm going with the following hack for now:

```nix
  autopatchelf = import (fetchTarball {
    url = "https://github.com/svanderburg/nix-patchtools/archive/6cc6fa4e0d8e1f24be155f6c60af34c8756c9828.tar.gz";
    sha256 = "sha256:0yff6fj0jv1d6kmrq4div1z8xxvgiqfndhfcqr9snw1h1mkr4yva";
  });
  trustedcoin = pkgs.stdenv.mkDerivation {
    name = "trustedcoin";
    src = pkgs.fetchurl {
      url = "https://github.com/nbd-wtf/trustedcoin/releases/download/v0.6.1/trustedcoin-v0.6.1-linux-amd64.tar.gz";
      sha256 = "sha256-J0oWvlsQPYiZCcwOv2G8c1O5A/ghM7JZG2VCLdBXBP8=";
    };
    buildInputs = [ autopatchelf ];
    libs = pkgs.lib.makeLibraryPath [ pkgs.glibc ];
    unpackPhase = ''
      tar -xzf "$src"
    '';
    installPhase = ''
      mkdir -p $out/usr/bin
      mv trustedcoin $out/usr/bin
      autopatchelf $out/usr/bin
    '';
    dontStrip = true;
  };
```


erikarvstedt:
Stalled PR https://github.com/fort-nix/nix-bitcoin/pull/533 adds a trustedcoin package and plugin, and clightning modules changes that roughly equal this PR.
You've just answered my question raised in https://github.com/fort-nix/nix-bitcoin/pull/533#issuecomment-1229428074, so the proposed cln modules changes (which keep a bitcoind dependency) are seem to be the right way to go.

Do you want to pull in the pkg/plugin from #553?

elsirion:
I just didn't check the open PRs, this was more of a quickfix for myself that I thought to upstream in case people find it useful. #533 seems like the better design, probably just needs a rebase?

PR #586, opened 2023-01-20
Misc. improvements

erikarvstedt:
Some of these are required for the upcoming [mempool flake](https://github.com/erikarvstedt/nix-bitcoin-mempool).

### DIFF
--- a/helper/start-bash-session.sh
+++ b/helper/start-bash-session.sh
@@ -1,0 +1,12 @@
+# Start an interactive bash session in the current bash environment.
+# This is helpful for debugging bash scripts like pkg update scripts,
+# by adding `source <path-to>/start-bash-session.sh` at the location to
+# be inspected.
+
+
+# BASH_ENVIRONMENT contains definitions of read-only variables like 'BASHOPTS' that
+# cause warnings on evaluation. Suppress these warnings while sourcing.
+#
+# shellcheck disable=SC2016
+BASH_ENVIRONMENT=<(declare -p; declare -pf) \
+  bash --rcfile <(echo 'source $BASH_ENVIRONMENT 2>/dev/null')
--- a/modules/netns-isolation.nix
+++ b/modules/netns-isolation.nix
@@ -215,9 +215,11 @@ in {
 
         };
       };
-    in foldl (services: n:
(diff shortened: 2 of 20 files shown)


helper/start-bash-session.sh
```diff
@@ -0,0 +1,8 @@
+# Start an interactive bash session in the current bash environment.
```
jonasnick, line 1: Looks like this is currently unused. What is it for?

modules/netns-isolation.nix
```diff
@@ -297,6 +299,7 @@ in {
         id = 31;
         connections = [ "bitcoind" ];
       };
+      # id = 32 reserved for mempool
```
jonasnick, line 302: How about "mempool" -> "mempool extension flake" to clarify that mempool isn't part of this code base?

Wouldn't it be better to explicitly reserve a different range for extension flakes, e.g., anything > 1000?

helper/start-bash-session.sh
erikarvstedt: It's used in the mempool flake:
https://github.com/erikarvstedt/nix-bitcoin-mempool/blob/a47cc2d5a1c08b32eee79fa079bbadbc7bf5cf57/pkgs/mempool/generate.sh#L6

modules/netns-isolation.nix
erikarvstedt: Mempool will be added as a nix-bitcoin module when the backend bug is resolved.

New extension flakes will be assigned `id`s >= 100. (`id`s map to the last byte of a IPv4 address.)

erikarvstedt: Done.

New extension flakes will be assigned `id`s >= 100. (`id`s map to the last byte of a IPv4 address.)

helper/start-bash-session.sh
jonasnick: Could you add a comment explaining how this is related to nix-bitcoin if it's unused in nix-bitcoin ("useful for flakes" or something)?

erikarvstedt: Done.

jonasnick (approved):
ACK addfa8ec6bab320de0b808fd76809635947377e8

PR #587, opened 2023-01-21
Fulcrum: Fix available memory detection

erikarvstedt:


jonasnick (approved):
ACK 86dc7e2669004d4abeeff4fae0817d7a06d40e00

For completeness, [this is the code in fulcrum that attempts to access meminfo](https://github.com/cculianu/Fulcrum/blob/52a3cc1b428c379b1738f339e3a0276063eb1572/src/Util.cpp#L370) and defaults to 2GB. Thanks @erikarvstedt for `enter_service`. Very useful for debugging this.

PR #591, opened 2023-02-03
Readme: Add mempool flake

erikarvstedt:


DanConwayDev:
As I understand it, this is the first tool to be integrated into nix-bitcoin using the flakes model.

I'd imagine regular users of NixOS would know exactly how to apply this and other similar flakes without any issues based on the nix-bitcoin-mempool readme. Users of nix-bitcoin that don't use NixOS elsewhere (like me) may struggle.
 
Currently references to using flakes in the documentation assume the user is adding nix-bitcoin to a wider configuration rather than running a machine dedicated to nix-bitcoin. 

I'm guess that everything after [2. Nix Installation of install.md](https://github.com/fort-nix/nix-bitcoin/blob/master/docs/install.md#2-nix-installation) would be different and krops would no longer be used for deployment.

erikarvstedt:
I've improved section `In a non-flakes config` in the [README](https://github.com/fort-nix/nix-bitcoin-mempool). Does this help?

DanConwayDev:
Yes. That tweak make it much more intuitive to include the non-flake config. Thanks. It working on my node.

I suppose there isn't much of a value proposition for users like me moving to flakes and consequently there isn't a need for detailed instructions how.

README.md
```diff
@@ -88,18 +88,22 @@ NixOS modules ([src](modules/modules.nix))
...
 * Helper
   * [netns-isolation](modules/netns-isolation.nix): isolates applications on the network-level via network namespaces
   * [nodeinfo](modules/nodeinfo.nix): script which prints info about the node's services
   * [backups](modules/backups.nix): duplicity backups of all your node's important files
-  * [operator](modules/operator.nix): adds non-root user `operator` who has access to client tools (e.g. `bitcoin-cli`, `lightning-cli`)
+  * [operator](modules/operator.nix): configures a non-root user who has access to client tools (e.g. `bitcoin-cli`, `lightning-cli`)
+
+#### Extension modules
+
+* [Mempool](https://github.com/fort-nix/nix-bitcoin-mempool): Bitcoin visualizer, explorer and API service
```
jonasnick, line 106: We could define extension module here. Maybe something like "Extension modules are maintained separately from the main nix-bitcoin repository and have their own review and release process".

jonasnick (commented):
I tested the mempool lightly. Works like a charm!

README.md
erikarvstedt: Great idea. Fixed.

jonasnick (commented):
ACK bf8de6744ed4a30984354fb8c31e7e45910e8609

jonasnick (approved):

PR #597, opened 2023-03-03
Integrate trustedcoin clightning plugin

seberm:
This PR tries to add a trustedcoin support into the nix-bitcoin.

The work is based on PRs:
- https://github.com/fort-nix/nix-bitcoin/pull/533 
- https://github.com/fort-nix/nix-bitcoin/pull/583

The code is rebased on current master.

I have removed the `clightning.useBcli` option and replaced it with `clightning.useBitcoind`. When setting this option to `true`, the bitcoind unit is forcerly disabled and `clightning.plugins.trustedcoin.enabled` is set to `true` as suggested in https://github.com/fort-nix/nix-bitcoin/pull/533#issuecomment-1229428074.

I also tried to add some basic tests.

@erikarvstedt , could you please take a look and give me some suggestions? Please keep in mind this is only a draft but I would like to move this feature forward.

Thanks!

CC: @neverupdate , @MatthewCroughan , @elsirion 
Closes #533

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ NixOS modules ([src](modules/modules.nix))
     * [prometheus](https://github.com/lightningd/plugins/tree/master/prometheus): lightning node exporter for the prometheus timeseries server
     * [rebalance](https://github.com/lightningd/plugins/tree/master/rebalance): keeps your channels balanced
     * [summary](https://github.com/lightningd/plugins/tree/master/summary): print a nice summary of the node status
-    * [trustedcoin](https://github.com/nbd-wtf/trustedcoin): replaces bitcoind with trusted public explorers
+    * [trustedcoin](https://github.com/nbd-wtf/trustedcoin) [[experimental](docs/services.md#trustedcoin-hints)]: replaces bitcoind with trusted public explorers
     * [zmq](https://github.com/lightningd/plugins/tree/master/zmq): publishes notifications via ZeroMQ to configured endpoints
   * [clightning-rest](https://github.com/Ride-The-Lightning/c-lightning-REST): REST server for clightning
   * [lnd](https://github.com/lightningnetwork/lnd) with support for announcing an onion service and [static channel backups](https://github.com/lightningnetwork/lnd/blob/master/docs/recovery.md)
--- a/README.md
+++ b/README.md
@@ -79,6 +79,7 @@ NixOS modules ([src](modules/modules.nix))
     * [prometheus](https://github.com/lightningd/plugins/tree/master/prometheus): lightning node exporter for the prometheus timeseries server
     * [rebalance](https://github.com/lightningd/plugins/tree/master/rebalance): keeps your channels balanced
     * [summary](https://github.com/lightningd/plugins/tree/master/summary): print a nice summary of the node status
+    * [trustedcoin](https://github.com/nbd-wtf/trustedcoin): replaces bitcoind with trusted public explorers
     * [zmq](https://github.com/lightningd/plugins/tree/master/zmq): publishes notifications via ZeroMQ to configured endpoints
   * [clightning-rest](https://github.com/Ride-The-Lightning/c-lightning-REST): REST server for clightning
   * [lnd](https://github.com/lightningnetwork/lnd) with support for announcing an onion service and [static channel backups](https://github.com/lightningnetwork/lnd/blob/master/docs/recovery.md)
--- a/modules/clightning-plugins/trustedcoin.nix
+++ b/modules/clightning-plugins/trustedcoin.nix
@@ -1,0 +1,21 @@
+{ config, lib, pkgs, ... }:
+
+with lib;
+let cfg = config.services.clightning.plugins.trustedcoin; in
+{
+  options.services.clightning.plugins.trustedcoin = {
+    enable = mkEnableOption "Trustedcoin (clightning plugin)";
+    package = mkOption {
+      type = types.package;
+      default = config.nix-bitcoin.pkgs.trustedcoin;
+      defaultText = "config.nix-bitcoin.pkgs.trustedcoin";
+      description = "The package providing trustedcoin binaries.";
+    };
+  };
+
+  config = mkIf cfg.enable {
+    services.clightning.extraConfig = ''
+      plugin=${cfg.package}/bin/trustedcoin
+    '';
+  };
+}
--- a/modules/clightning-plugins/trustedcoin.nix
+++ b/modules/clightning-plugins/trustedcoin.nix
@@ -9,7 +9,7 @@ let cfg = config.services.clightning.plugins.trustedcoin; in
       type = types.package;
       default = config.nix-bitcoin.pkgs.trustedcoin;
       defaultText = "config.nix-bitcoin.pkgs.trustedcoin";
-      description = ''
+      description = mdDoc ''
         The package providing trustedcoin binaries. Trustedcoin will try to
         use a bitcoind as a trusted source for getting block data. If this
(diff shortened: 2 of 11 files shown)


modules/clightning.nix
```diff
@@ -146,11 +149,12 @@ in {
 
   config = mkIf cfg.enable {
     services.bitcoind = {
-      enable = true;
+      enable = mkForce cfg.useBitcoind;
```
elsirion, line 152: Trustedcoin is useful to run CLN with a pruned `bitcoind` (it only falls back to explorers if the local `bitcoind` can't answer a request), so always deactivating `bitcoind` to run `trustedcoin` is suboptimal.

elsirion (changes requested):

modules/clightning.nix
jonasnick: My understanding is that trustedcoin indeed loads [from bitcoind if available](https://github.com/nbd-wtf/trustedcoin#using-bitcoind). In that case, I think it makes sense to leave `enable = true;` as is.

Thanks @seberm btw for another attempt at this PR!

modules/clightning-plugins/trustedcoin.nix
```diff
@@ -0,0 +1,26 @@
...
+let cfg = config.services.clightning.plugins.trustedcoin; in
+{
+  options.services.clightning.plugins.trustedcoin = {
+    enable = mkEnableOption "Trustedcoin (clightning plugin)";
+    package = mkOption {
+      type = types.package;
+      default = config.nix-bitcoin.pkgs.trustedcoin;
+      defaultText = "config.nix-bitcoin.pkgs.trustedcoin";
+      description = mdDoc ''
+        The package providing trustedcoin binaries. The trustedcoin will try to
```
elsirion, line 13: "The trustedcoin" sounds a bit odd

test/tests.py
```diff
@@ -433,6 +433,18 @@ def get_block_height(ip, port):
...
+@test("trustedcoin")
+def _():
+    machine.wait_for_unit("bitcoind")
+    machine.wait_for_unit("clightning")
+
+    # Let's check the trustedcoin plugin was correctly initialized
+    machine.wait_until_succeeds(log_has_string("clightning", "plugin-trustedcoin[^^]\[0m\s+initialized plugin"))
+    machine.wait_until_succeeds(log_has_string("clightning", "plugin-trustedcoin[^^]\[0m\s+bitcoind RPC working"))
+    machine.wait_until_succeeds(log_has_string("clightning", "plugin-trustedcoin[^^]\[0m\s+tip: 0"))
+    machine.wait_until_succeeds(log_has_string("clightning", "plugin-trustedcoin[^^]\[0m\s+estimatefees error: none of the esploras returned usable responses"))
```
elsirion, line 445: Couldn't this test break if esplora gets fixed and suddenly returns usable responses?

elsirion (approved):
Looks good to me

test/tests.py
elsirion: Ah, I don't understand the testing framework enough apparently :see_no_evil: just assumed testing for something being broken was odd.

modules/clightning-plugins/trustedcoin.nix
elsirion: I'd just write "`trustedcoin` will try to …", but also not a native English speaker, so take with a grain of salt :)

test/tests.py
jonasnick: > so the trustedcoin should never get a response from the esploras.

Yes but the trustedcoin plugin could change it's output in that case. But I think the test is ok.

pkgs/trustedcoin/default.nix
```diff
@@ -0,0 +1,23 @@
...
+  };
+
+  vendorSha256 = "sha256-xvkK9rMQlXTnNyOMd79qxVSvhgPobcBk9cq4/YWbupY=";
+
+  subPackages = [ "." ];
+
+  meta = with lib; {
+    description = "Light bitcoin node implementation";
+    homepage = "https://github.com/fiatjaf/trustedcoin";
+    maintainers = with maintainers; [ nixbitcoin ];
```
jonasnick, line 20: Are you sure @nixbitcoin wants to be the maintainer? If not, add yourself or maybe use `fort-nix`, the organization name.

pkgs/trustedcoin/get-sha256.sh
```diff
@@ -0,0 +1,19 @@
...
+set -euo pipefail
+
+TMPDIR="$(mktemp -d -p /tmp)"
+trap 'rm -rf $TMPDIR' EXIT
+cd "$TMPDIR"
+
+echo "Fetching latest release"
+repo=fiatjaf/trustedcoin
+latest=$(curl -s --show-error "https://api.github.com/repos/${repo}/releases/latest" | jq -r .tag_name)
+echo "Latest release is $latest"
```
jonasnick, line 12: This gives me `Latest release is null`

README.md
```diff
@@ -79,6 +79,7 @@ NixOS modules ([src](modules/modules.nix))
     * [prometheus](https://github.com/lightningd/plugins/tree/master/prometheus): lightning node exporter for the prometheus timeseries server
     * [rebalance](https://github.com/lightningd/plugins/tree/master/rebalance): keeps your channels balanced
     * [summary](https://github.com/lightningd/plugins/tree/master/summary): print a nice summary of the node status
+    * [trustedcoin](https://github.com/fiatjaf/trustedcoin): replaces bitcoind with trusted public explorers
```
jonasnick, line 82: I think the page was moved to https://github.com/nbd-wtf/trustedcoin

test/tests.nix
```diff
@@ -315,6 +315,16 @@ let
       services.lnd.enable = true;
       services.bitcoind.prune = 1000;
     };
+
+    # Test the special clightning setup where trustedcoin plugin is used
+    trustedcoin = {
+      tests.cligtning = mkForce false;
```
jonasnick, line 321: Typo `cligtning`. This shouldn't have had an effect.

modules/clightning-plugins/trustedcoin.nix
```diff
@@ -0,0 +1,26 @@
...
+    enable = mkEnableOption "Trustedcoin (clightning plugin)";
+    package = mkOption {
+      type = types.package;
+      default = config.nix-bitcoin.pkgs.trustedcoin;
+      defaultText = "config.nix-bitcoin.pkgs.trustedcoin";
+      description = mdDoc ''
+        The package providing trustedcoin binaries. Trustedcoin will try to
+        use a bitcoind as a main trusted source for getting on-chain block
+        data. If this fails, it will use a trustedcoin providers instead.
+      '';
```
jonasnick, line 16: Nit: I would either remove everything but the first sentence (since the description of the package options seems a bit like the wrong location for this explanation) or also add the line "This isn't what you should be doing, but sometimes you may need it." from the trustedcoin README.

jonasnick (commented):
Thanks again for your work on this. Left a few comments. Did you test whether this actually works and you can use clightning with an unsynced bitcoind?

jonasnick:
@seberm Thanks for investigating this. I don't know why clightning's RPC does not get ready in time when trustedcoin is enabled. Maybe while syncing the plugin is in some initialization phase which blocks clightnings RPC startup? It seems like trustedcoin is doing some work despite the timeouts. Does clightning stop restarting on timeout at some after a while or does this go on forever?

jonasnick (approved):
ACK a3c654768ca917a3e53a28116c30ba89d16f2838

PR #609, opened 2023-05-07
lnd: fix non-static patch URL

erikarvstedt:
Currently, updating https://github.com/lightningnetwork/lnd/pull/7672 would break our patched `lnd` because the patch URL is not content-addressed.

After this is merged, we should make another release.

Sorry for the mishap.


jonasnick (approved):
ACK 1d69c9c824a1c7e3dfda9e98b431023b70e2f3e3

PR #613, opened 2023-06-09
clightning-plugins/prometheus: fix patch not being applied

erikarvstedt:
Check this with
```bash
diff -u --color \
  $(nix build --no-link --print-out-paths github:fort-nix/nix-bitcoin?ref=v0.0.92#clightning-plugins/prometheus) \
  $(nix build --no-link --print-out-paths .#clightning-plugins/prometheus)
```


jonasnick (commented):
ACK 1b741c7cc11e61456d8deb9cf150978fec76f2cd

jonasnick (approved):

PR #621, opened 2023-06-25
Trustedcoin fixes

erikarvstedt:


erikarvstedt:
cc @seberm

erikarvstedt:
Found a few more issues, will look into this later.

I've spun out `remove from CI, disable temporarily` into https://github.com/fort-nix/nix-bitcoin/pull/623, so we can quickly address the CI failure.

docs/services.md
```diff
@@ -579,26 +579,18 @@ services.clightning = {
...
-
-If you face these issues and you still need to use trustedcoin, use can disable
-clightning's tor hardening by setting this option in your `configuration.nix`
-file:
-
-```
-services.clightning.tor.enforce = false;
+To work around this and connect via clearnet instead, set this option:
+```nix
+services.clightning.plugins.trustedcoin.enableTor = false;
```
jonasnick, line 595: This does not look like an improvement because it suggests that Tor is only disabled for trustedcoin and not for all of clightning.

jonasnick (commented):
Looks good so far modulo the comment about disabling Tor.

docs/services.md
erikarvstedt: No, the point of `trustedcoin.enableTor` is to disable Tor only for trustedcoin, but not for clightning.

`trustedcoin.enableTor` sets `clightning.tor.enforce == false`.
This is required to enable trustedcoin clearnet connections. (trustedcoin runs in the clightning service namespace.)
But the value of `clightning.tor.proxy`, which enables Tor for clightning, is unaffected.

jonasnick: Right, this just sets option just sets `clightning.tor.enforce`.

erikarvstedt:
cc @seberm

seberm:
Looks good to me. I haven't known about the "dev-env", it's really useful.

Thanks for your improvements!

jonasnick (approved):
Thanks @erikarvstedt @seberm

ACK 4e8369705df76b1c3ecd945446a9593516bac6ab

PR #625, opened 2023-07-09
Add versioned release branches

erikarvstedt:
#### Copy of main commit msg
`push-release` now pushes to an extra branch named `nixos-<VERSION>`, alongside branch `release`.
This allows users to track a specific NixOS release, so that their config doesn't break when nix-bitcoin switches to a new NixOS release.


jonasnick (approved):
ACK 6510f269f6a4655e69d4d59e012462dca066a65a

jonasnick:
And created the [branch](https://github.com/fort-nix/nix-bitcoin/tree/nixos-22.11).

PR #626, opened 2023-07-09
secure-node: add `sudo` shell alias for `doas`

erikarvstedt:
#### Copy of commit msg

A convenience helper which allows running most `sudo` cmds while `doas` is enabled.

This is safe because all args supported by both `sudo` and `doas` that lead to command execution (like `-u <user>`) have identical semantics.


jonasnick (approved):
ACK 5eaf104efef7630039c1ed4acdc067eabe9b61ac

PR #631, opened 2023-07-29
rtl: fix lnd, lightning-loop connection errors

erikarvstedt:
#### Copy of commit msg

lnd and lightning-loop resolve `localhost` to an IPv4 address when creating RPC sockets.

Since NixOS 23.05, RTL (nodejs) resolves `localhost` to an IPv6 address when connecting to lnd and lightning-loop, which leads to connection errors.

To fix these and other potential errors, replace all instances of `localhost` with `127.0.0.1`.

### DIFF
--- a/dev/dev-features.sh
+++ b/dev/dev-features.sh
@@ -275,9 +275,9 @@ c journalctl -u joinmarket-ob-watcher
 c journalctl -f -u joinmarket-ob-watcher
 
 # Check webinterface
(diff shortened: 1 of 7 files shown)


jonasnick (approved):
ACK 14ca8b461b3ea098067ea4c3c52922263daa6459

naturallaw777:
Thanks for correcting this! Much appreciated!

PR #636, opened 2023-08-18
docs/install: don't enter examples shell while creating node config

erikarvstedt:
[Rendered doc](https://github.com/erikarvstedt/nix-bitcoin/blob/fix-install-docs/docs/install.md#3-setup-deployment-directory).

#### Copy of commit msg
1. This simplifies the setup

2. Fixes a bug where a user skipped step 4.6 (run `nix-shell`) which he deemed superfluous because a Nix shell was already active.
   This caused an error during deployment.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -210,31 +210,21 @@ You can also build Nix from source by following the instructions at https://nixo
 1. Clone this project
 
     ```
-    cd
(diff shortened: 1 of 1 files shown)


jonasnick (commented):
ACK 329fbac4045d252160f4b068d1956933f816419d

jonasnick (approved):

PR #638, opened 2023-08-20
Minor clightning fixes

erikarvstedt:
Use this to test commit `clightning: don't cleanup socket on startup`:
```bash
read -rd '' script <<'EOF' || :
  systemctl stop clightning
  echo
  echo 'clightning has stopped'
  systemctl status clightning | head -3
  echo
  echo 'the socket still exists after clightning has stopped'
  ls /var/lib/clightning/bitcoin/lightning-rpc
  echo
  systemctl start clightning
  echo 'clightning starts, rpc calls work'
  sleep 0.5
  lightning-cli getinfo
EOF
run-tests.sh -s clightning container --run c bash -c "$script"
```

### DIFF
--- a/modules/clightning.nix
+++ b/modules/clightning.nix
@@ -165,7 +165,7 @@ in {
     ];
 
     systemd.services.clightning = {
-      path  = [ nbPkgs.bitcoind ];
+      path  = [ bitcoind.package ];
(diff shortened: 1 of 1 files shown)


jonasnick (approved):
ACK e253b898585e406c18bde7a5d6ec9cc6985351d5

PR #640, opened 2023-08-31
pkgs/bitcoind: use `pkgs` instead of `pkgsUnstable`

erikarvstedt:
#### Copy of commit msg
All deps required by `bitcoind` are available in `pkgs`.


jonasnick (approved):
ACK ff30af6ed3d7a8f16eb6ff7dbe2e25e635e80156

Issue #611, opened 2023-06-02
Update to bitcoind 25.0


erikarvstedt:
Blocked by https://github.com/bitcoin/bitcoin/issues/27722.
Due to this bug, bitcoind hangs indefinitely during shutdown when a RPC request was aborted previously.
In nix-bitcoin, bitcoind is eventually killed by systemd (default: after 30 min).

If you want to run v25.0 anyway, add the following to your config:
```nix
services.bitcoind.package = pkgs.bitcoind;
```

fanquake:
25.1 is now available, and should contain the fix neccesary to unblock this.

jonasnick:
Merged and released. Thanks @fanquake @prusnak!

PR #646, opened 2023-09-17
upgrade bitcoin package from nixpkgs

RCasatta:


prusnak:
local copy of bitcoin package was introduced by https://github.com/fort-nix/nix-bitcoin/pull/610/commits/3650d4befe6cad275b259c78f5373f8e851d070f - the reason was "Bitcoin Core 25.0 may sometimes hang when shutting down."

has the issue been resolved? I don't think so since https://github.com/bitcoin/bitcoin/issues/27722  is still open

erikarvstedt:
See also our pinned issue [`Update to bitcoind 25.0`](https://github.com/fort-nix/nix-bitcoin/issues/611) which shows how to switch to 25.0.

RCasatta:
Oh I didn't know
Sorry

PR #647, opened 2023-09-18
Shell completion

RCasatta:
Would be cool to have shell completion for the nix installed bitcoin binaries! (also in nixpkgs bitcoin, but I started here)

This is just a draft for multiple reasons:

* on top of #646 which was not right
* doesn't work (the nix build is successful but the resulting image hasn't shell completion for bitcoin-cli)
* it is missing shell completion for other bins
* it makes sense to have completion also withGui
* I am not sure targeting directly the build dir for finding completionion files is right

At the moment I am not going to work on this, however, I wanted to leave it here in case me of the future want to retry or someone else want to start from here

### DIFF
--- a/pkgs/bitcoin/default.nix
+++ b/pkgs/bitcoin/default.nix
@@ -22,6 +22,7 @@
 , nixosTests
 , withGui
 , withWallet ? true
+, installShellFiles
 }:
 
 let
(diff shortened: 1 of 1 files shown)


prusnak:
I think this belongs to the upstream package - can you please send a PR to https://github.com/NixOS/nixpkgs repo?

prusnak:
I will gladly review any PR to nixpkgs when it is ready.

erikarvstedt:
I've created an [upstream PR](https://github.com/NixOS/nixpkgs/pull/256091).

PR #651, opened 2023-10-03
treewide: ensure services are started after secrets setup

erikarvstedt:
#### Copy of commit msg
Now all services that access secrets only run after the secrets setup has finished.

Previously, we assumed that the systemd `after` dependency is transitive, i.e. that adding an `after = [ "bitcoind.service" ]` to a service implicitly pulled in the `after` dependency to `nix-bitcoin-secrets.target` (which is defined for `bitcoind`).
This is not the case. Services could start before secrets setup had finished, leading to service failure.

### DIFF
--- a/modules/backups.nix
+++ b/modules/backups.nix
@@ -106,7 +106,7 @@ in {
 
       systemd.services.duplicity = {
         wants = postgresqlBackupServices;
-        after = postgresqlBackupServices;
+        after = postgresqlBackupServices ++ [ "nix-bitcoin-secrets.target" ];
       };
 
       services.postgresqlBackup = {
--- a/modules/btcpayserver.nix
+++ b/modules/btcpayserver.nix
@@ -174,7 +174,7 @@ in {
     in rec {
       wantedBy = [ "multi-user.target" ];
       requires = [ "bitcoind.service" "postgresql.service" ] ++ optional cfg.btcpayserver.lbtc "liquidd.service";
-      after = requires;
+      after = requires ++ [ "nix-bitcoin-secrets.target" ];
       preStart = ''
         install -m 600 ${configFile} '${cfg.nbxplorer.dataDir}/settings.config'
         {
--- a/modules/clightning.nix
+++ b/modules/clightning.nix
@@ -168,7 +168,7 @@ in {
       path  = [ bitcoind.package ];
       wantedBy = [ "multi-user.target" ];
       requires = [ "bitcoind.service" ];
-      after = [ "bitcoind.service" ];
+      after = [ "bitcoind.service" "nix-bitcoin-secrets.target" ];
       preStart = ''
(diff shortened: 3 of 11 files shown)


erikarvstedt:
### How to reproduce the bug that's fixed by this PR

Add the following to `dev-scenarios.nix`:

```nix
{ lib, scenarios }:
with lib;
rec {
  switch = { pkgs, ... }: {
    services.bitcoind = {
      enable = true;
      package = pkgs.bitcoind;
    };

    # A dummy bitcoind client service
    systemd.services.test = {
      requires = [ "bitcoind.service" ];
      after = [ "bitcoind.service" ];
      script = "sleep infinity";
    };
  };

  switch2 = { lib, ... }: {
    imports = [ switch ];
    systemd.services.test.script = lib.mkBefore ''
      echo hello
    '';
    nix-bitcoin.generateSecretsCmds.test = ''
      sleep 1
    '';
  };

  switch3 = { lib, ... }: {
    imports = [ switch2 ];
    systemd.services.test.after = [ "nix-bitcoin-secrets.target" ];

    # Force generate-secrets.service to restart
    nix-bitcoin.generateSecretsCmds.test = ''
      echo hello
    '';
  };
}
```

Run the following in a dev shell:
```bash
dev-run-tests -s switch container
# Then the following in the container shell:
dev-run-tests -s switch2 container --run c journalctl --output=short-precise
dev-run-tests -s switch3 container --run c journalctl --output=short-precise
```

When switching to config `switch2`, `test.service` and `setup-secrets.service` are both restarted.
Because of the missing `after = [ "nix-bitcoin-secrets.target" ];` dependency in `test.service`, the services are started at the same time:

```
Oct 03 13:48:42.059186 nb-test systemd[1]: Starting setup-secrets.service...
Oct 03 13:48:42.060215 nb-test systemd[1]: Started test.service.
...
Oct 03 13:48:42.071544 nb-test test-start[594]: hello
...
Oct 03 13:48:43.101273 nb-test systemd[1]: Reached target nix-bitcoin-secrets.target.
```

This is fixed in config `switch3`, where `test.service` starts after `setup-secrets.service`.

jonasnick (commented):
Hm, so in a scenario in which we have an order A < B < C for units A, B and C, and A and C are restarted then systemd does not take into account that A < C.

We could introduce a target that activates after nix-bitcoin-secrets and bitcoind and that services could depend on, but I don't think it makes things clearer.

concept ACK

Shouldn't the yieldgenerator's `after` option also include `nix-bitcoin-secrets.target`?

erikarvstedt:
> Shouldn't the yieldgenerator's after option also include nix-bitcoin-secrets.target?

Yes, thanks. Fixed.

jonasnick (approved):
ACK 90ce68cb16cac9455eb5812447052e830528ba59

PR #653, opened 2023-10-08
Revert "clightning: don't cleanup socket on startup"

erikarvstedt:
#### Copy of commit msg
Also, add a detailed comment.

Without this commit, clightning client services may fail to start because of clightning RPC connection failures. This can happen unnoticed due to the service auto restart mechanism.


jonasnick (approved):
ACK fc1466e743a09bafe3fa05e7916f34646b8b243e

Issue #556, opened 2022-10-16
Use Flakes in recommended deployment method, improve deployment performance


erikarvstedt:
Use flakes for improved extensibility and pure builds.
Switch to a faster deployment scheme.

This requires Nix PR `Source tree abstraction` to be merged:
nix-bitcoin config flakes potentially include plaintext secrets files. PR `Source tree abstraction` allows evaluating the flake without copying the source (including secrets) to the store.

prusnak:
> This requires Nix PR `Source tree abstraction` to be merged:

Here is the mentioned Nix PR: https://github.com/NixOS/nix/pull/6530

PR #655, opened 2023-10-19
examples/flake.nix: minor improvements

erikarvstedt:
Add some improvements in response to a user question from Matrix.

### DIFF
--- a/examples/flakes/flake.nix
+++ b/examples/flakes/flake.nix
@@ -1,3 +1,10 @@
+# This is a system configuration template that uses nix-bitcoin.
+#
+# You can adapt this to an existing system flake by copying the parts
+# relevant to nix-bitcoin.
+#
+# Make sure to check and edit all lines marked by 'FIXME:'
+
 {
   description = "A basic nix-bitcoin node";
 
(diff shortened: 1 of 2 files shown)


seberm:
Hello,
maybe a silly question but is there some difference when I want to deploy the flake system configuration using krops? How can I do that?

Thanks

erikarvstedt:
Krops is not recommended. Use a modern deployment tool like colmena or morph. Details: https://github.com/fort-nix/nix-bitcoin/issues/556.

examples/flakes/flake.nix
```diff
@@ -8,13 +8,14 @@
 {
   description = "A basic nix-bitcoin node";
 
-  inputs.nix-bitcoin.url = "github:fort-nix/nix-bitcoin/release";
-  # You can also use a version branch to track a specific NixOS release
-  # inputs.nix-bitcoin.url = "github:fort-nix/nix-bitcoin/nixos-23.05";
+  inputs.nix-bitcoin.url = "github:fort-nix/nix-bitcoin/nixos-23.05";
```
jonasnick, line 11: Doesn't this mean that `nix flake update` will not actually update your system anymore when we update to a new NixOS release? That seems undesirable - at least as a default option. Is the user supposed to notice that there are not updates coming from the nix-bitcoin inputs anymore?

erikarvstedt: Agreed. I've reverted this change.

jonasnick (approved):
ACK 742fd8fdd0a66d705def05e2e85522ed1561a813

PR #658, opened 2023-11-01
presets/wireguard: make compatible with `secure-node` preset

erikarvstedt:
#### Copy of commit msg
Disable Tor enforcement set by the secure-node preset.


jonasnick (approved):
utACK e784e0ceb8db51177bf1da38478d6d4c026eb245

PR #662, opened 2023-12-01
Update to NixOS 23.11

erikarvstedt:
Todo:
- [x] Fix Python-based pkgs like clightning plugins and joinmarket
- [x] Manual testing


erikarvstedt:
@nixbitcoin, could you do some joinmarket testing?

pkgs/python-packages/jmclient/default.nix
```diff
@@ -1,22 +1,63 @@
...
-      --replace "'pyjwt==2.4.0'" "'pyjwt==2.6.0'"
+      --replace "'argon2_cffi==21.3.0'" "'argon2_cffi==23.1.0'"
+    substituteInPlace setup.py \
+      --replace "'pyjwt==2.4.0'" "'pyjwt==2.8.0'"
+    substituteInPlace setup.py \
+      --replace "'werkzeug==2.2.3'" "'werkzeug==2.3.7'"
   '';
 
+  # The unit tests can't be run in a Nix build environment
+  doCheck = false;
```
jonasnick, line 55: If I set this to true the build keeps succeeding. Do you mean that the tests that do run are unnecessary?

pkgs/build-support/fetch-node-modules.nix
```diff
@@ -25,6 +25,9 @@ stdenvNoCC.mkDerivation ({
 
   phases = "unpackPhase patchPhase buildPhase installPhase";
 
+  # npm doesn't support var `SSL_CERT_FILE`.
+  NODE_EXTRA_CA_CERTS = "${cacert}/etc/ssl/certs/ca-bundle.crt";
```
jonasnick, line 29: How did you find out about this CA issue? https://github.com/NixOS/nixpkgs/pull/257513 suggests that npm has hardcoded CAs.

pkgs/python-packages/default.nix
```diff
@@ -5,7 +5,7 @@ let
   # be supported through the NixOS 23.05 release cycle.`
   # TODO-EXTERNAL: consider removing when
   # https://github.com/Simplexum/python-bitcointx/issues/76 and
-  # https://github.com/JoinMarket-Org/joinmarket-clientserver#1451 are resolved.
+  # https://github.com/JoinMarket-Org/joinmarket-clientserver/issues/1451 are resolved.
```
jonasnick, line 8: How about
```
consider removing when a fix to
https://github.com/JoinMarket-Org/joinmarket-clientserver/issues/1613
has been released.
```

jonasnick (commented):
We can also update `install.md` (maybe mention in `dev/README.md` "Switching to a new NixOS release" as well):   
```  
wget  https://releases.nixos.org/nixos/23.11/nixos-23.11.1494.b4372c4924d9/nixos-minimal-23.11.1494.b4372c4924d9-x86_64-linux.iso
sha256sum nixos-minimal-23.11.1494.b4372c4924d9-x86_64-linux.iso
# output: f48cf810432c1f04b291c947b36f824823dfef8ebfa0e1906602a516450189d8
```

pkgs/python-packages/jmclient/default.nix
erikarvstedt: With `doCheck = true`, the logs show `Ran 0 tests in 0.000s`. Running the tests requires adding `pytestCheckHook`, which then shows that the tests hang. (Probably while setting up a bitcoind regtest instance. I haven't looked into this closely.)

pkgs/build-support/fetch-node-modules.nix
erikarvstedt: Without `NODE_EXTRA_CA_CERTS`, the `npm ci` command in `fetchNodeModules` hangs.
You can repro this by setting the hash in `backend = fetchNodeModules` to an empty string and then running `nix build --no-link -L .#mempool-backend`.

erikarvstedt:
- I've updated this PR to the latest 23.11 channel update.
- I've added `python-bitcointx: 1.1.3 -> 1.1.4` which removes the openssl 1.1 dependency and is compatible with the latest joinmarket release.

Running live on nixbitcoin.org.

pkgs/python-packages/default.nix
erikarvstedt: Resolved by the `python-bitcointx` update (see post below).

jonasnick: This comment block doesn't seem to be updated?

jonasnick (commented):
mu-nit: if you touch the commits again, can you add `clightning: 23.08.1 -> 23.11` to the NixOS-update-commit description?

pkgs/build-support/fetch-node-modules.nix
jonasnick: Hm ok, good that we discovered this!

erikarvstedt:
> mu-nit: if you touch the commits again, can you add clightning: 23.08.1 -> 23.11 to the NixOS-update-commit description?

Yes, I always add pkg updates to the description.
This PR doesn't contain a clightning update: `nix eval .#modulesPkgs.clightning.version` = `23.08.1`.

erikarvstedt:
> This comment block doesn't seem to be updated? 

Thanks, fixed.

erikarvstedt:
I've fixed up a few commit descriptions. No src changes.

jonasnick (approved):
> This PR doesn't contain a clightning update: nix eval .#modulesPkgs.clightning.version = 23.08.1.

Ah, you only updated the 23.11 channel.

ACK 85bbdb857a40fbe6fa34e9c095112b494e599501

Thanks @erikarvstedt and also @kristapsk for updating joinmarket!

PR #675, opened 2024-01-15
test: use `nixpkgs-unstable` flake input for nix-bitcoin pkgs

erikarvstedt:
#### Copy of commit msg
By using the `default` flake module for tests, `pkgsUnstable` in `pkgs/default.nix` gets passed the `nixpkgs-unstable` flake input instead of falling back to importing `nixpkgsPinned.nixpkgs-unstable`.

For some use cases this prevents importing `nixpkgs-unstable` twice (once via the `nixpkgs-unstable` flake input, once via `import nixpkgsPinned.nixpkgs-unstable`).

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -26,7 +26,7 @@
         # "armv7l-linux"
       ];
 
-      test = import ./test/tests.nix nixpkgs.lib;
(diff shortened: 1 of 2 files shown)


jonasnick (approved):
utACK afbd8a3cad5d5341d201e32ec35ebbf0265b830e

PR #676, opened 2024-01-15
examples: add persistent container example

erikarvstedt:
This is also helpful for users who want to run an extra testnet/regtest node on a nix-bitcoin system.


jonasnick (commented):
How do we update flake.lock? Should we do it in `helper/update-flake.sh`?

Should we test the container in `run-tests.sh`?

erikarvstedt:
> How do we update flake.lock? Should we do it in helper/update-flake.sh?

I accidentally commited `flake.lock`. The container flake doesn't need to ship with a lockfile, its only input is `nix-bitcoin` itself.

> Should we test the container in run-tests.sh?

Yes, I've added a test. Because it's a container, it only takes a few seconds to run.

jonasnick (approved):
ACK 2c503a9dcf728910efc2aaabfab0db957064ae4a

PR #699, opened 2024-05-14
JoinMarket: Add support for jmwalletd

seberm:


seberm:
Hello @erikarvstedt @jonasnick ,
I am finallizing the patches for jmwalletd support, which is a prerequisite for the Jam (see PR https://github.com/fort-nix/nix-bitcoin/pull/700).

This service should probably only be accessible locally. I don't see any reason to expose it via an onion address.

Additionally, please take a look at the address option, which I have marked as `readOnly` and `internal`. The reason for this is that I haven't found a way to set the listening address for `jmwalletd`. The service always listens on `0.0.0.0`.

Also, do you think it makes sense to add some documentation about this service to the README?

Could you please review this and assist with the service hardening?

Thank you!

naturallaw777:
Just wanted to checking in. What are the next steps or further organization to get this merged?

naturallaw777:
Thanks @seberm for putting this together!

PR #708, opened 2024-06-16
add `liana` support

plebhash:
[Liana](https://github.com/wizardsardine/liana) is a simple Bitcoin wallet with built-in loss protection and inheritance.

Current status: draft.

Roadmap:
- [x] `pkgs/liana/default.nix` builds `lianad` + `liana-cli`
- [ ] `pkgs/liana/default.nix` builds `liana-gui`
- [ ] `modules/liana.nix`

### DIFF
--- a/pkgs/liana/default.nix
+++ b/pkgs/liana/default.nix
@@ -1,0 +1,21 @@
+{ stdenv, lib, fetchFromGitHub, rustPlatform }:
+
+rustPlatform.buildRustPackage rec {
+  pname = "liana";
+  version = "5.0";
+
+  src = fetchFromGitHub {
+    owner = "wizardsardine";
+    repo = pname;
+    rev = version;
+    hash = "sha256-RkZ2HSN7IjwN3tD0UhpMeQeqkb+Y79kSWnjJZ5KPbQk=";
+  };
+
+  cargoHash = "sha256-v3tMz93mNsTy0k27IzgYk9bL2VfqtXImMlnvwgswp6U=";
+
+  meta = {
+    description = "The missing safety net for your coins";
+    homepage = "https://wizardsardine.com/liana/";
+    license = lib.licenses.bsd3;
+  };
+}


modules/liana.nix
```diff
@@ -0,0 +1,87 @@
...
+    };
+    data_dir = mkOption {
+      type = types.path;
+      default = "/var/lib/lianad";
+      description = mdDoc "Path to the folder where we should store the application data.";
+    };
+    main_descriptor = mkOption {
+      type = types.str;
+      default = "wsh(or_d(pk([0dd8c6f0/48'/1'/0'/2']tpubDFMbZ7U5k5hEfsttnZTKMmwrGMHnqUGxhShsvBjHimXBpmAp5KmxpyGsLx2toCaQgYq5TipBLhTUtA2pRSB9b14m5KwSohTDoCHkk1EnqtZ/<0;1>/*),and_v(v:pkh([d4ab66f1/48'/1'/0'/2']tpubDEXYN145WM4rVKtcWpySBYiVQ229pmrnyAGJT14BBh2QJr7ABJswchDicZfFaauLyXhDad1nCoCZQEwAW87JPotP93ykC9WJvoASnBjYBxW/<0;1>/*),older(65535))))#7nvn6ssc";
+      description = mdDoc "The wallet descriptor.";
```
storopoli, line 20: Isn't it better to move these hardcoded defaults to `example` instead of `default`.
I can see easily an user shooting himself in the foot.

plebhash: ideally this is a `systemd` parameter

the problem is that `lianad`'s CLI is still very limited, so everything needs to be written indo a `config.toml`

so the challenge here is:
- declare all options from [`lianad_config_example.toml`](https://github.com/wizardsardine/liana/blob/master/contrib/lianad_config_example.toml) so they're available as `systemd` configurations (which seems to be the most common pattern)
- 

I started doing a [heredoc inside the `preStart`](
https://github.com/plebhash/nix-bitcoin/blob/e69002dc223bd34f538ff6594e72663fc0109ecb/modules/liana.nix#L61) but still unsure whether that is the best approach.

modules/liana.nix
```diff
@@ -0,0 +1,102 @@
...
+network = "signet"
+poll_interval_secs = 30
+
+# these should come from options.services.bitcoind
+[bitcoind_config]
+addr = "127.0.0.1:38332"
+auth = "username:password"
+
+EOF
+        '';
```
storopoli, line 79: This could be a `builtins.fromTOML` follwed by a `toString`, e.g.:

```nix
let
  tomlFile =./example.toml;
  tomlData = builtins.readFile tomlFile;
  tomlObj = builtins.fromTOML tomlData;
  tomlString = toString tomlObj;
in
  tomlString
```

Then you can interpolate the hardcoded nastiness with `options.services.lianad.*`

modules/liana.nix
storopoli: Yes but still in a footgun for the user.
Check my comment below on how to interpolate TOML file contents into a string.

storopoli (changes requested):

modules/liana.nix
storopoli: Refs:

- https://nix.dev/manual/nix/2.22/language/builtins.html#builtins-fromTOML
- https://nix.dev/manual/nix/2.22/language/builtins.html#builtins-readFile
- https://nix.dev/manual/nix/2.22/language/builtins.html#builtins-toString

plebhash: will this allow me to dynamically generate the TOML file and write it into the output's datadir?

it seems like a convenient tool for reading configs that already exist on disk, but our use case is the opposite direction:

we need to generate a TOML file with values coming from `options.services.lianad.*`

0xB10C:
Would upstreaming this into nixpkgs make sense? There are a bunch of Bitcoin related packages and services there already.

dunxen:
> Would upstreaming this into nixpkgs make sense? There are a bunch of Bitcoin related packages and services in nixpkgs already.

I already maintain the `liana` package in nixpkgs.

plebhash:
> > Would upstreaming this into nixpkgs make sense? There are a bunch of Bitcoin related packages and services in nixpkgs already.
> 
> I already maintain the `liana` package in nixpkgs. https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/li/liana/package.nix

that's awesome @dunxen thanks for sharing!

I actually went for different tradeoffs on the `liana/default.nix` currently living on this PR

`nix-bitcoin` is usually headless, so `liana-cli` + `lianad` are a 1st class citizen on this approach
are you open to PRs expanding `nixpkgs/tree/master/pkgs/by-name/liana/package.nix` to include those two binaries + config file?

dunxen:
> are you open to PRs expanding `nixpkgs/tree/master/pkgs/by-name/liana/package.nix` to include those two binaries + config file? that way, `nix-bitcoin` could simply consume `nixpkgs`

Sure, thanks! It would be one package definition but with attributes to choose what to build. So I'd imagine that if you install `lianad` then it would result in the `lianad` and `liana-cli` binaries, else if you do `liana` then it is the GUI (only?).

Kind of how bitcoin does it in nixpkgs:

https://github.com/NixOS/nixpkgs/blob/a71e967ef3694799d0c418c98332f7ff4cc5f6af/pkgs/top-level/all-packages.nix#L35801-L35810

jonasnick:
Concept ACK on adding a liana module to nix-bitcoin. Thanks!

We should also mention the new service in README.md and in examples/configuration.nix and add a test checking that it runs.

plebhash:
thanks @dunxen 

> So we'd need to define `liana` and `lianad` in `all-packages.nix` again since we'd need to change those attributes.

I can't find `liana` anywhere inside [`all-packages.nix`](https://github.com/NixOS/nixpkgs/blob/master/pkgs/top-level/all-packages.nix)? Did you write this starting from the assumption that it was there, or are you suggesting that we add it there for the first time?

dunxen:
> I can't find `liana` anywhere inside [`all-packages.nix`](https://github.com/NixOS/nixpkgs/blob/master/pkgs/top-level/all-packages.nix)? Did you write this starting from the assumption that it was there, or are you suggesting that we add it there for the first time?

Packages in `by-name` are automatically included at the top-level. Liana used to be there but when there was a refactor and the package definition was moved to `by-name`, it was removed from `all-packages`. It needs to go back there if you have attributes to set like in this scenario. I've left some comments on your nixpkgs PR. Thanks!

Apologies to all for the slightly off topic 🙏

modules/liana.nix
wskeele: You can generate a static toml file from a set using ``pkgs.formats.toml``. E.g:

```nix
mySet = { foo = "bar""; };
myTomlFile = (pkgs.formats.toml {}).generate mySet;
```

But this of course cannot contain secrets such as bitcoind credentials that are not going into the nix store. So then the `preStart` rule could execute a command that reads the generated TOML, and appends the additional secrets, writing it to the dataDir for liana.

The lnd module actually already does this:

```nix
      preStart = ''
        install -m600 ${configFile} '${cfg.dataDir}/lnd.conf'
        {
          echo "bitcoind.rpcpass=$(cat ${secretsDir}/bitcoin-rpcpassword-${rpcUser})"
          ${optionalString (cfg.getPublicAddressCmd != "") ''
            echo "externalip=$(${cfg.getPublicAddressCmd})"
          ''}
        } >> '${cfg.dataDir}/lnd.conf'
```

PR #714, opened 2024-07-08
bitcoind-rpc-public-whitelist: add `getpeerinfo`

erikarvstedt:
Required by lnd 0.18.0.

The bug fixed here can only be reproduced on a fully-synced mainnet/testnet node.
lnd error:
```
[ERR] LNWL: Encountered err while fetching fee filters: status code: 403, response: ""
```


jonasnick (approved):
utACK 1f32792d3177a12036df049cf53c42451c695ba5

PR #716, opened 2024-07-11
README: don't mention IRC channel which is now unused

mmilata:
There's 3-4 people and one of them is me. There hasn't been actual discussion in months.

Would be also nice to put link to the matrix channel in the topic but that would probably require assistance from libera.chat or nix people who own the `#nix-` prefix?


jonasnick:
Concept ACK. It doesn't make sense to direct users to a dead channel. We'll also need to change this for nixbitcoin.org (https://github.com/fort-nix/nixbitcoin.org).

erikarvstedt (approved):
Concept ACK. I've mirrored this to nixbitcoin.org.

PR #722, opened 2024-07-27
joinmarket: Add option `settings`

### LINKED ISSUE #707: Allow for modification of the "merge_algorithm" for JoinMarket policy
Nix-Bitcoin should alllow for the "merge_algorithm" to be modified to something other than "default". The "gradual" setting is nice for cleaning up UTXOs.

See: https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/35e6f286fb9be0dc90c2c5c7648d380c103a3870/src/jmclient/configure.py#L256

I believe I can manually modify the configuration after boot and restart the service but on rebuild it will get overridden. Is that right?

erikarvstedt:
Joinmarket settings can now be freely specified:
```nix
services.joinmarket = {
  enable = true;
  settings = {
    POLICY = {
      merge_algorithm = "gradual";
      tx_fees = 5;
    };
    LOGGING = {
      console_log_level = "DEBUG";
    };
  };
};
```
Closes #707

The updated `joinmarket-ob-watcher` is deployed at https://nixbitcoin.org and is working fine.

@seberm, can you help reviewing and testing this?

### DIFF
--- a/docs/services.md
+++ b/docs/services.md
@@ -543,8 +543,8 @@ See [here](https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/master
     services.joinmarket.yieldgenerator = {
       enable = true;
       # Optional: Add custom parameters
-      txfee = 200;
       cjfee_a = 300;
+      cjfee_r = 0.00003;
     };
     '';
     ```
--- a/modules/joinmarket-ob-watcher.nix
+++ b/modules/joinmarket-ob-watcher.nix
@@ -19,6 +19,19 @@ let
       default = "/var/lib/joinmarket-ob-watcher";
       description = "The data directory for JoinMarket orderbook watcher.";
     };
+    settings = mkOption {
+      type = with types; attrsOf anything;
+      example = {
+        LOGGING = {
+          console_log_level = "DEBUG";
+        };
+      };
+      description = ''
+        Joinmarket settings.
+        See here for possible options:
+        https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/v0.9.11/src/jmclient/configure.py#L98
+      '';
+    };
     user = mkOption {
       type = types.str;
       default = "joinmarket-ob-watcher";
@@ -40,17 +53,6 @@ let
   secretsDir = config.nix-bitcoin.secretsDir;
 
   inherit (config.services) bitcoind joinmarket;
-
-  configFile = builtins.toFile "config" ''
-    [BLOCKCHAIN]
-    blockchain_source = bitcoin-rpc
-    network = ${bitcoind.network}
-    rpc_host = ${bitcoind.rpc.address}
-    rpc_port = ${toString bitcoind.rpc.port}
-    rpc_user = ${bitcoind.rpc.users.joinmarket-ob-watcher.name}
-
-    ${joinmarket.messagingConfig}
-  '';
 in {
   inherit options;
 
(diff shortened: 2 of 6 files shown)


erikarvstedt:
I've removed option `txfee_contribution`. It's deprecated and should no longer be used:
https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/v0.9.3/docs/release-notes/release-notes-0.9.3.md

docs/services.md
seberm, line 1: I think we should also show the basic usage of `services.joinmarket.settings` in the docs. Maybe show the user example how to add another messaging server?

seberm (commented):
Hello @erikarvstedt , I've tried to run the code on my testing node. All of the settings were deployed successfuly and the services seems to be running fine. I think it's good to go.

modules/joinmarket.nix
```diff
@@ -61,19 +78,20 @@ let
...
     inherit (nbLib) cliExec;
 
     yieldgenerator = {
-      enable = mkEnableOption "JoinMarket yield generator bot";
+      enable = mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          Enable the JoinMarket yield generator bot.
+          Documentation: https://github.com/JoinMarket-Org/joinmarket-clientserver/v0.9.11/master/docs/YIELDGENERATOR.md
```
jonasnick, line 92: Link doesn't work.

erikarvstedt: Fixed.

modules/joinmarket-ob-watcher.nix
```diff
@@ -19,6 +19,19 @@ let
...
+      type = with types; attrsOf anything;
+      example = {
+        LOGGING = {
+          console_log_level = "DEBUG";
+        };
+      };
+      description = ''
+        Joinmarket settings.
+        See here for possible options:
+        https://github.com/JoinMarket-Org/joinmarket-clientserver/blob/v0.9.11/src/jmclient/configure.py#L98
```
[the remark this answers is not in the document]
erikarvstedt: Yeah, I've changed to URL to point to `master`. We're also doing this for other services: `cd nix-bitoin-src; rg blob/master`.

docs/services.md
erikarvstedt: Note there's already an [example setting](https://github.com/erikarvstedt/nix-bitcoin/blob/d90a2102150ae8d89c1dad088a2ce76d9682e10a/modules/joinmarket.nix#L52).

Are there other messaging servers besides the ones from the default config?

jonasnick (approved):
utACK 9d9e92132a6364175da18be8ecaf4e9ead4425c0

docs/services.md
erikarvstedt: @seberm, I'd love to hear your feedback.

erikarvstedt:
I think we can merge this now.

docs/services.md
seberm: I don't believe there are any other default messaging servers. I was hoping you could add an example showing how to use the settings option.

As you mentioned, there is already an example configuration that I didn't notice.

That said, I think we're good to go. And sorry for my late response - I am currently away from my pc.

erikarvstedt: Thanks! I think this use case is too obscure to warrant extra docs. It's relatively straightforward to figure out how to add a custom server by reading the `joinmarket.nix` source.

jonasnick (approved):
utACK 9d9e92132a6364175da18be8ecaf4e9ead4425c0

PR #723, opened 2024-07-31
bitcoind-rpc-public-whitelist: Add `getnodeaddresses`

### LINKED ISSUE #721: Add `getnodeaddresses` to public bitcoind RPC whitelist?
We recently added `getpeerinfo` to the public whitelist.
If we also added `getnodeaddresses`, lnd could use the public whitelist instead of defining a custom whitelist ([src](https://github.com/fort-nix/nix-bitcoin/blob/7dad87527f7aa366c3ecb67073405f15aa61c7f8/modules/lnd.nix#L189)).

In my view, RPC calls **should not** be added to the public whitelist if they
- affect the behavior of the bitcoind instance
- allow reading private wallet data

`getnodeaddresses` (https://chainquery.com/bitcoin-cli/getnodeaddresses) doesn't match these criteria and thus can be added to the public whitelist.

Was there a reason for not adding these calls in the first place?

cc @jonasnick

erikarvstedt:
Fixes #721.

### DIFF
--- a/modules/bitcoind-rpc-public-whitelist.nix
+++ b/modules/bitcoind-rpc-public-whitelist.nix
@@ -35,6 +35,7 @@
   "getnetworkhashps"
   # Network
   "getnetworkinfo"
+  "getnodeaddresses"
   "getpeerinfo"
   # Rawtransactions
   "analyzepsbt"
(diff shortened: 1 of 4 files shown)


jonasnick (approved):
ACK da625fc13d31e0c9213872656d1477ddd812f768

PR #738, opened 2024-09-25
docs: add ASmap process via extraSources

jurraca:
This is one of the few options where we need to copy a file over, so thought it would be worth a mention in this section of the docs.

Follow up might be an explicit option in the module, but not sure how to amend the `extraSources` based on a module config, if possible.


0xB10C:
fwiw: For my personal use, I'm packaging `asmap_latest.dat` as package (which only contains the asmap_latest.dat binary). See https://github.com/0xB10C/nix/blob/master/pkgs/asmap-data/default.nix. I guess that could be an alternative to copying the file.

jurraca:
closing in favor of #746

PR #740, opened 2024-10-05
add RestartMode=direct

spalfs:
i've noticed when my clightning crashes, services that have it as a dependency (clightning-rest, rtl) will be deactived without restarting:
2024-10-05-210614_1280x1024_scrot
if we set RestartMode="direct", it should restart the services along with the parent:
https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#RestartMode=
i just started testing this on my server, will report back once clightning crashes again (in a day or two) to see if I have the expected result
would we want this on all dependency services?
these are the only two i am using and can test.

### DIFF
--- a/modules/clightning-rest.nix
+++ b/modules/clightning-rest.nix
@@ -96,6 +96,7 @@ in {
         User = clightning.user;
         Restart = "on-failure";
         RestartSec = "10s";
+        RestartMode = "direct";
         ReadWritePaths = [ cfg.dataDir ];
         inherit (nbLib.allowNetlink) RestrictAddressFamilies;
       } // nbLib.allowedIPAddresses cfg.tor.enforce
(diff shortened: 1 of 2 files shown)


spalfs:
yup, looks to work:
2024-10-10-085611_1280x1024_scrot

jonasnick (commented):
Hey @spalfs. Thanks for the PR.

*Sigh*... how can it be so hard to make systemd behave reasonably. It's not like we're doing anything out of the ordinary...

If clightning crashes, then the dependent services are restarted correctly. This can be tested by sending SIGKILL or some other signal to the lightningd process.

The problem appears to be that the dependent services are "shut down" when clightning fails to start up due to the ExecPostStart script timing out. When clightning successfully restarts after that failure, then contrary to expectation, the dependent services are not restarted as well.

I haven't managed to reproduce this, but I would like to, because we have other services that may fail in a similar way. The docs of `RestartMode` don't seem to indicate to me that this is the right way to solve this.

erikarvstedt:
Superseded by #761 which works around this bug at the level of the depending service, not in the dependency itself.

PR #745, opened 2024-11-06
lnd: make `ExecStartPost` extensible

erikarvstedt:
#### Copy of commit msg
By wrapping the value in a list, users can add other `ExecStartPost` scripts.
The previous scalar value could only be replaced, but not merged with other definitions.

#### Example
```bash
run-tests.sh -s '{
  services.lnd.enable = true;
  systemd.services.lnd.serviceConfig.ExecStartPost = lib.mkAfter [ "${pkgs.hello}/bin/hello" ];
}' container --run c journalctl -u lnd
```
Result:
```
...
Nov 06 16:51:05 nb-test lnd[533]: 2024-11-06 16:51:05.096 [INF] LTND: Waiting for chain backend to finish sync, start_height=0
Nov 06 16:51:05 nb-test hello[545]: Hello, world!
Nov 06 16:51:05 nb-test systemd[1]: Started lnd.service.
```

### DIFF
--- a/modules/lnd.nix
+++ b/modules/lnd.nix
@@ -262,9 +262,8 @@ in {
         ExecStartPost = let
           curl = "${pkgs.curl}/bin/curl -fsS --cacert ${cfg.certPath}";
           restUrl = "https://${nbLib.addressWithPort cfg.restAddress cfg.restPort}/v1";
-        in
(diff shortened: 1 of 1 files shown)


jonasnick (approved):
ACK ca96e1959eb42a94ad45e1c719449d3f20d40420

PR #753, opened 2024-12-17
update nixpkgs

jonasnick:
Pkg updates in nixpkgs stable:
btcpayserver: 1.13.5 -> 1.13.7
nbxplorer: 2.5.7 -> 2.5.16

Pkg updates in nixpkgs unstable:
clboss: 0.14.0 -> 0.14.1
clightning: 24.08.2 -> 24.11
electrs: 0.10.6 -> 0.10.7
elementsd: 23.2.1 -> 23.2.4

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -44,11 +44,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1733808091,
-        "narHash": "sha256-KWwINTQelKOoQgrXftxoqxmKFZb9pLVfnRvK270nkVk=",
+        "lastModified": 1734323986,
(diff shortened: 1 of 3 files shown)


jonasnick:
> I'll later add a more detailed bug report to the clightning issue.

Thanks!

> Maybe we should pin clightning to the previous version (i.e. nixpkgs-stable) for now, including hints how to switch to the latest version.

Does that help? According to the issues, v24.8.1 is affected as well. What would be the downside of temporarily increasing the startup timeout?

erikarvstedt:
> v24.8.1 is affected as well

Ah, you're right. Let's go ahead with merging then.

erikarvstedt (approved):
ACK 64bfc3f3a0449c86c239f84a962397b0cd9fbe12

PR #754, opened 2024-12-17
container example: add clightning to flake, remove electrs

msgilligan:
Some of the commands listed in `usage.sh` require clightning to be enabled. This commit enables clightning so they will work.

### DIFF
--- a/examples/container/flake.nix
+++ b/examples/container/flake.nix
@@ -61,7 +61,7 @@
               # Enable some services.
               # See ../configuration.nix for all available features.
               services.bitcoind.enable = true;
-              services.electrs.enable = true;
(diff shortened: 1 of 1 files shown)


erikarvstedt:
Thanks for catching this!
Could you also remove `services.electrs.enable = true;` which is unused in `usage.sh`?

erikarvstedt (approved):
ACK 306b30cf62b09989092c9e297a9c76f035b1d067

Thank you!

msgilligan:
> Could you also remove `services.electrs.enable = true;` which is unused in `usage.sh`?

Done! force-pushed and I also updated the title of the PR.

PR #760, opened 2025-01-22
Fix test `persistentContainerExample`

erikarvstedt:
#### Copy of main commit msg
This test requires `electrs` to be enabled.

Also, by using `electrs` instead of `clightning` we avoid the bug where clightning hangs at startup when internet access is unavailable. Due to technical limititations, internet access is available in NixOS containers only after the container has started.


jonasnick (approved):
ACK b8f6343ee4b924273bba11877e208e2ad4dd3adb

PR #761, opened 2025-01-25
services: use `wants` dependency where possible

erikarvstedt:
#### Copy of commit msg
Let A be a service that depends on another service B.
When A can gracefully handle failures and restarts of B, use
```
wants = [ "B.service" ];
after = [ "B.service" ];
```
instead of
```
requires = [ "B.service" ];
after = [ "B.service" ];
```
in the definition of A.

This way, A keeps running when B is stopped or restarted after a failure.
With `requires`, A is instead stopped when B is stopped or restarted due to a failure.

This brings two benefits:

1. Improved uptime
   Examples:
   - RTL keeps running when one lightning node has failed
   - btcpayserver keeps running and accepting on-chain payments when the lightning node has crashed

2. Avoids a systemd bug where depending units (`A.service` in the above example) are not restarted when their dependency fails
   (issue github/systemd#18856, no full link to avoid spamming the issue).
   In real world nix-bitcoin deployments, this issue was only likely to appear when clightning failed during activation, causing depending units (like `RTL`) to stop and not be restarted.
   All services depending on `clightning` have now been changed to use `wants`, thereby avoiding the bug.

Services `electrs` and `lightning-loop` fail when their respective dependencies stop, so these services have not been changed.
I also haven't changed services `joinmarket` and `joinmarket-yieldgenerator`. Further manual testing is needed to determine if they can be switched to `wants`.

#### Manual testing
<details>
<summary>Commands used for manual testing</summary>

```bash
run-tests.sh --scenario '{
imports = [ scenarios.regtestBase ];
services.btcpayserver.enable = true;
services.clightning-rest.enable = true;
services.fulcrum.enable = true;
services.liquidd.enable = true;
services.lnd = {
  enable = true;
  port = 9736;
};
services.mempool.enable = true;
services.rtl = {
  enable = true;
  address = "0.0.0.0";
  nodes.clightning.enable = true;
  nodes.lnd.enable = true;
};
}' container --run c

# Check that all services are running
systemctl status

time=$(date +%s.%6N)
systemctl stop bitcoind
sleep 10
systemctl status bitcoind
lightning-cli getinfo
lncli getinfo # this hangs when bitcoind is stopped
systemctl start bitcoind

# Check that no service failed
journalctl --grep 'Failed with result'

## Check that services gracefully handle connection loss to bitcoind
journalctl -u nbxplorer --since=@$time
journalctl -u btcpayserver --since=@$time
# clightning shows bitcoind rpc failures as '... exited with status 1'
journalctl -u clightning --since=@$time
lightning-cli getinfo
journalctl -u fulcrum --since=@$time
journalctl -u liquidd --since=@$time
journalctl -u lnd --since=@$time
lncli getinfo
journalctl -u mempool --since=@$time

# Check stopping and starting clightning and lnd
time=$(date +%s.%6N)
systemctl stop clightning lnd
systemctl status clightning lnd
journalctl --grep 'Failed with result'
# open RTL to check that it gracefully handles node connection errors
http://10.225.255.2:3000
systemctl start clightning lnd
journalctl --grep 'Failed with result'
journalctl -u clightning-rest --since=@$time


### Test units that fail when their requirement fails
run-tests.sh --scenario '{
imports = [ scenarios.regtestBase ];
services.lightning-loop.enable = true;
}' container --run c

time=$(date +%s.%6N)
systemctl stop bitcoind lnd
sleep 10
journalctl -u electrs --since=@$time
journalctl -u lightning-loop --since=@$time
```
</details>

### DIFF
--- a/modules/btcpayserver.nix
+++ b/modules/btcpayserver.nix
@@ -171,8 +171,9 @@ in {
       '';
     in rec {
       wantedBy = [ "multi-user.target" ];
-      requires = [ "bitcoind.service" "postgresql.service" ] ++ optional cfg.btcpayserver.lbtc "liquidd.service";
-      after = requires ++ [ "nix-bitcoin-secrets.target" ];
+      requires = [ "postgresql.service" ];
+      wants = [ "bitcoind.service" ] ++ optional cfg.btcpayserver.lbtc "liquidd.service";
+      after = requires ++ wants ++ [ "nix-bitcoin-secrets.target" ];
       preStart = ''
         install -m 600 ${configFile} '${cfg.nbxplorer.dataDir}/settings.config'
         {
@@ -223,11 +224,12 @@ in {
         lbtcexplorerurl=${nbExplorerUrl}
         lbtcexplorercookiefile=${nbExplorerCookie}
       '');
-    in let self = {
+    in rec {
       wantedBy = [ "multi-user.target" ];
-      requires = [ "nbxplorer.service" "postgresql.service" ]
-                 ++ optional (cfg.btcpayserver.lightningBackend != null) "${cfg.btcpayserver.lightningBackend}.service";
-      after = self.requires;
+      requires = [ "postgresql.service" ];
+      wants = [ "nbxplorer.service" ]
+              ++ optional (cfg.btcpayserver.lightningBackend != null) "${cfg.btcpayserver.lightningBackend}.service";
+      after = requires ++ wants;
       serviceConfig = nbLib.defaultHardening // {
         ExecStart = ''
           ${cfg.btcpayserver.package}/bin/btcpayserver --conf=${configFile} \
@@ -244,7 +246,7 @@ in {
       } // nbLib.allowedIPAddresses cfg.btcpayserver.tor.enforce;
       startLimitIntervalSec = 30;
       startLimitBurst = 10;
-    }; in self;
+    };
 
     users.users.${cfg.nbxplorer.user} = {
       isSystemUser = true;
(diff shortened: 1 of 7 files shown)


jonasnick (approved):
Thanks for finding and commenting on github/systemd#18856! Agree that the current systemd behavior is highly unexpected.

Anyway, you're right: for many of our services `wants` is better than `requires`.

I think there's an argument to be made for keeping `requires` for lightning -> bitcoind. We know that it's a fundamental property of Lightning that it bitcoind must run and be synced to the current blockchain. Otherwise, channel balances could be stolen.
If we'd change the lightning -> bitcoind relationship to `wants` as in this PR, then users might miss more easily that bitcoind is not actually running. In the case of clightning it's only indicated in the logs, but RTL does not show it (at least I couldn't find it). Maybe with lnd the situation is slightly better because lncli hangs and RTL gives an error. If we'd keep `requires` then it would be more obvious to the user that the bitcoind service is not running.

What do you think?

Btw, I think the first scenario for manual testing can be extended with something like the following because the PR changes the btcpaypayserver -> nbxplorer and mempool -> fulcrum relationships from `requires` to `wants`:

```
systemctl stop nbxplorer
systemctl status btcpayserver # check that btcpayserver handles connection loss to nbxplorer
systemctl start nbxplorer
systemctl status btcpayserver # check that btcpayserver reconnects correctly

systemctl stop fulcrum
systemctl status mempool # check that mempool handles connection loss to fulcrum
systemctl start fulcrum
systemctl status mempool # check that mempool reconnects correctly
```

erikarvstedt:
> If we'd keep `requires` then it would be more obvious to the user that the bitcoind service is not running.

Yes, that's sensible. I've changed clightning and lnd back to `requires`.

I manually tested `btcpaypayserver -> nbxplorer` and `mempool -> fulcrum` before but forgot the list the commands. Both services work fine with `wants`.

jonasnick (approved):
ACK 4aaef5fdf4dd2c6093f388d5d575e90d23c974be

Issue #664, opened 2023-12-12
Support Bitcoin Knots as alternative to Bitcoin Core?


kristapsk:
So that people have choice.

erikarvstedt:
Here's how to use Knots instead of Core:
```nix
services.bitcoind.package = let
  pkgsUnstable = import (builtins.fetchTarball {
    # unstable (2023-12-12)
    url = "https://github.com/NixOS/nixpkgs/archive/666fc80e7b2afb570462423cb0e1cf1a3a34fedd.tar.gz";
    sha256 = "1j4ahdm49pbw5s1sw0vc0l12s1zxlnw0sjvnljxhdjbsvgbjsl7m";
  }) {};
in
  pkgsUnstable.bitcoind-knots;

```

When the latest Knots 25.1 release has arrived in the stable nixpkgs channel, you can simply do this instead:
```nix
services.bitcoind.package = pkgs.bitcoind-knots;
```

plebhash:
> To use Knots instead of Core, add this to your config:
> 
> services.bitcoind.package = config.nix-bitcoin.pkgs.bitcoind-knots;

@erikarvstedt is there a way to run both?

erikarvstedt:
Let's discuss this on Matrix and share back the results later. I've sent you a PM on Matrix.

PR #771, opened 2025-04-29
[bitcoind] Add option to run Bitcoin Knots

steepdawn974:
This PR adds support for running Bitcoin Knots as an alternative to the standard bitcoind full node in nix-bitcoin.
- Introduces a configuration option to select Bitcoin Knots instead of bitcoind
- Uses the official Nixpkgs `bitcoind-knots` package

Example for adding this snippet to an existing NixOS's `configuration.nix`:

```
  # Enable bitcoind from upstream nix-bitcoin
  services.bitcoind = { 
    enable = true;
    implementation = "knots";
    #more Options
  };
```

PS: this is my first time with nix-bitcoin; I'm not super familiar with the whole code base and may not have handled the `flake.lock` correctly. I just ran `nix flake lock --update-input nix-bitcoin` .. but later noticed there are some helper scripts that probably do sth similar


jonasnick (commented):
Hi @steepdawn974, thanks for your contribution.

How is this better than https://github.com/fort-nix/nix-bitcoin/issues/664#issuecomment-1851831390 ? I agree that we should at least document how to use knots.

steepdawn974:
> Hi @steepdawn974, thanks for your contribution.
> 
> How is this better than [#664 (comment)](https://github.com/fort-nix/nix-bitcoin/issues/664#issuecomment-1851831390) ? I agree that we should at least document how to use knots.

@jonasnick I honestly didn't see that comment.
Was coding up the change for personal use, and though it might be worthwhile for others too.

I guess the new `implementation` option is a more generic approach and would make it more accessible for users to switch between core/knots.
https://github.com/fort-nix/nix-bitcoin/blob/7efc48b510911febe4a44247e85eb4774a4dbda4/modules/bitcoind.nix#L267C1-L275C9

But I'm not mad if its just closed either ;-)

Tracachang:
@steepdawn974 

Your PR is the way to go. My comment was just trying to add some information to #664, essentially documenting the steps I went through when figuring out how to install the latest version of Knots, in case it helps others.

jonasnick:
Since this PR would add built-in support for knots to nix-bitcoin, it misses inclusion of the `implementation` option and knots into our test framework. This adds complexity and maintenance overhead as the tests need to pass for every nix-bitcoin release. However, I am not aware of a good reason to run knots, so I don't think that this additional complexity is justified.

However, @erikarvstedt suggested an alternative to me in private chat, which I am currently in favor of:
1. Add bitcoin knots to pinned.nix. That means you can always access the newest knots in the nixpkgs by simply
   setting

   ```
   services.bitcoind.package = config.nix-bitcoin.pkgs.bitcoind-knots;
   ```

   without having to go through the steps that @Tracachang mentioned.
2. Mention knots in the documentation of the bitcoin module's package option. This should make it easier to find.

What do you think? Would you be interested in implementing this?

erikarvstedt:
> Add bitcoin knots to pinned.nix ...
> Mention knots in the documentation ...

ACK. Will implement this very soon.

steepdawn974:
Understood. I suppose there is nothing needed from my side.
Feel free to close this, once the alternative implementation from @erikarvstedt  is merged.

esomore:
> Since this PR would add built-in support for knots to nix-bitcoin, it misses inclusion of the `implementation` option and knots into our test framework. This adds complexity and maintenance overhead as the tests need to pass for every nix-bitcoin release. However, I am not aware of a good reason to run knots, so I don't think that this additional complexity is justified.
> 
> However, @erikarvstedt suggested an alternative to me in private chat, which I am currently in favor of:
> 
> 1. Add bitcoin knots to pinned.nix. That means you can always access the newest knots in the nixpkgs by simply
>    setting
>    ```
>    services.bitcoind.package = config.nix-bitcoin.pkgs.bitcoind-knots;
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    without having to go through the steps that @Tracachang mentioned.
> 2. Mention knots in the documentation of the bitcoin module's package option. This should make it easier to find.
> 
> What do you think? Would you be interested in implementing this?

It's concerning that community members, myself included, seem to be facing restrictions commenting on Bitcoin Core PRs. This, along with Core's consideration of removing filters—which could increase chain bloat—may explain the notable increase in node runners adopting Bitcoin Knots. Supporting Knots through initiatives like this Nix package seems crucial. I wasn't aware of this package until now but will familiarize myself and look for ways to help.

jonasnick:
Thanks for your input @esomore. However, I still don't see a good reason to run knots.

1. The Bitcoin Core PR was locked because of spam and brigading that made a discussion on the technical aspects impossible. Besides, github PRs have not been considered the place for philosophical discussions at least since since the blocksize debate of 2017. That's what the Bitcoin mailing list is for, which has never been locked. So I don't see the current restrictions on the Bitcoin Core PR to be concerning at all.
2. I have no reason to believe that relaxing the filters for OP_RETURN transaction relay would increase the chain bloat. What relaxed rules could cause is that the existing bloat shifts to OP_RETURN, which is much less harmful to the network than othe methods of storing bloat. If you look at the mailing list, no one has put forward a convincing argument against the relaxation of the OP_RETURN relay rules.
   Nobody wants wasted chain space for JPGs (including virtually every active contributor to Bitcoin Core), but a restricted OP_RETURN relay policy does not achieve that.

That said, running nix-bitcoin with knots is possible and it will be even simpler in the future.

erikarvstedt:
https://github.com/fort-nix/nix-bitcoin/pull/773: pinned pkgs: add `bitcoind-knots`

erikarvstedt:
Implemented in #771.
To enable knots, add this to your config:
```nix
services.bitcoind.package = config.nix-bitcoin.pkgs.bitcoind-knots;
```

PR #773, opened 2025-05-12
pinned pkgs: add `bitcoind-knots`

erikarvstedt:


jonasnick (approved):
ACK 69090cddfac31f9c547409e7213ca2332ca9daa7

Was also able to confirm that knots currently passes our test suite.

PR #778, opened 2025-05-17
Update configuration.nix

Tracachang:
To clarify the configuration, this update explicitly presents the option to choose Bitcoind Knots.

### DIFF
--- a/examples/configuration.nix
+++ b/examples/configuration.nix
@@ -26,6 +26,10 @@
   ### BITCOIND
   # Bitcoind is enabled by default via secure-node.nix.
   #
+  # If you wish to run Bitcoind Knots instead of the default Bitcoind core,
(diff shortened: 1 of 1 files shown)


jonasnick (commented):
Thanks @Tracachang. I do not think setting bitcoind-knots should be part of an example config because it's an expert option (with subtle trade-offs) and we do not test it.

Tracachang:
Thanks @jonasnick . Totally understand the hesitation around implying official support or encouraging use of untested options. Still, I think including the bitcoind-knots line (commented out with a clear warning) would be a good way to make the option visible without suggesting it’s maintained. For example:

# Expert option — use Bitcoind Knots instead of Bitcoin Core (untested by nix-bitcoin maintainers):
# services.bitcoind.package = config.nix-bitcoin.pkgs.bitcoind-knots;

PR #780, opened 2025-05-20
clightning: set service as active only after clnrest has started

### LINKED ISSUE #777: tests: `wireguard-lndconnect` does not pass
`run-tests.sh -s wireguard-lndconnect` at 1586a34e902706b6834b059c9bd6d80caffce84a fails with

```
client: must succeed: curl -fsS --max-time 3 --insecure --header 'rune: ou66jj5XD3g9A32l822kpE-SWGyAvikIXAhohSHYYfs9MA==' -X POST https>
client # curl: (7) Failed to connect to 10.10.0.1 port 3010 after 0 ms: Could not connect to server
client: output:
Test "lnconnect-clnrest-wg" failed with error: "command `curl -fsS --max-time 3 --insecure --header 'rune: ou66jj5XD3g9A32l822kpE-SWGyA>
```

However, I when stepping through the tests manually with `run-tests.sh -s wireguard-lndconnect debug`, none of the tests fail. The test succeeded for me with v0.0.122 (which is still cached, I haven't rerun it). This indicates that there's an issue with the test (timing?) or my environment.

erikarvstedt:
Fixes #777

### DIFF
--- a/modules/clightning-plugins/clnrest.nix
+++ b/modules/clightning-plugins/clnrest.nix
@@ -54,6 +54,7 @@ let
   };
 
   cfg = config.services.clightning.plugins.clnrest;
+  nbLib = config.nix-bitcoin.lib;
   inherit (config.services) clightning;
 
   runePath = "${clightning.networkDir}/admin-rune";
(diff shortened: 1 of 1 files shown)


prusnak:
Surely there is a better idiomatic way how to wait for an open port rather than using while loop?

Maybe something like this (suggested by Claude)

```nix
{
  systemd.services.clightning = {
    # Keep your existing configuration...
    
    # Replace the postStart with just the rune creation
    postStart = mkAfter (
      optionalString cfg.createAdminRune ''
        if [[ ! -e '${runePath}' ]]; then
          rune=$(${clightning.cli}/bin/lightning-cli createrune | ${pkgs.jq}/bin/jq -r .rune)
          install -m 640 <(echo "$rune") '${runePath}'
        fi
      ''
    );
  };
  
  # Add this new socket unit
  systemd.sockets.clightning-rest = {
    wantedBy = [ "sockets.target" ];
    socketConfig = {
      ListenStream = "${nbLib.address cfg.address}:${toString cfg.port}";
      Accept = false;
    };
  };
  
  # Update the service to depend on the socket
  systemd.services.clightning.requires = [ "clightning-rest.socket" ];
  systemd.services.clightning.after = [ "clightning-rest.socket" ];
}
```

erikarvstedt:
The only way to solve this without polling is watching the clightning journal output for string `REST server running`:
```bash
if ! { exec 3>/dev/tcp/${nbLib.address cfg.address}/${toString cfg.port}; } &>/dev/null; then
  clightningStartTime=$(
    /run/current-system/systemd/bin/systemctl show --value -p InactiveExitTimestamp --timestamp=us+utc clightning.service
  )
  /run/current-system/systemd/bin/journalctl --since "$clightningStartTime" -u clightning.service -f | 
    ${pkgs.gawk}/bin/awk '/REST server running at/ { exit }'
fi
```
But this must run as root because the clightning user has no journal access, which further complicates things.

Given that the polling loop only runs when `clnrest` is first started (see the code comment from this PR), I think we should go for the simple polling solution.

Ideally, `clightning` would implement `systemd-notify`, extended to plugins.
@jonasnick, could you suggest this upstream? `bitcoind` and `lnd` also support `systemd-notify`.

The approach suggested by Claude is flawed in many ways and not usable. I'll leave it to another AI to refute it.

jonasnick (approved):
ACK ce6244cc698c841c1179e4c85054d094118cdc5c

Thanks very much @erikarvstedt. Yes, I will suggest systemd-notify upstream.

PR #794, opened 2025-06-21
albyhub: add module

bleetube:
## Project description

This app was previously [nostr-wallet-connect](https://github.com/getAlby/nostr-wallet-connect) which implemented protocols ([nip-47](https://github.com/nostr-protocol/nips/blob/master/47.md) and [lud16](https://github.com/lnurl/luds/blob/luds/16.md), among others) for controlling an lnd lightning wallet over the distributed network protocol called [nostr](https://nostr.com/). The Alby "hub" expands the scope of the project to include support for more lightning wallet backends in addition to lnd: pheonixd and an internal ldk node.

While Alby started out providing custodial lightning software, they have since developed more self-custodial software such as this.

In full transparency, I've observed some rough edges in the build process, some of which have been smoothed out by deploying this as a nix package. Many other rough edges were smoothed out by updates over the last six months, so I have a lot more confidence in this project as of late.

* The upstream build [process](https://github.com/getAlby/hub/blob/master/Dockerfile) involves [pulling in pre-compiled](https://github.com/getAlby/hub/blob/master/build/docker/copy_dylibs.sh) library modules for ldk-node from separate repositories. By contrast, the new nix package handles this gracefully by creating builds inputs for albyhub and building all library modules from source.
* Minor auditability concern: I'm not a C or GO developer, but I wish [go-secp256k1-zkp](https://github.com/vulpemventures/go-secp256k1-zkp) imported the C library as a gitmodule, rather than maintaining the entire tree in repo. Then it would've been easier for me to reason about any changes they may have made to secp256k1-zkp. I did mention this concern in an [issue](https://github.com/vulpemventures/go-secp256k1-zkp/issues/20
) and was dismissed summarily (possibly for not explaining it well).

## Metadata

* homepage URL: https://albyhub.com/
* source URL: https://github.com/getAlby/hub
* user guide: https://guides.getalby.com/user-guide/alby-hub
* license: asl20
* platforms: amd64, aarch64

### DIFF
--- a/modules/albyhub.nix
+++ b/modules/albyhub.nix
@@ -164,7 +164,7 @@ in
       type = with types; nullOr str;
       default =
         if config.services.mempool.enable then
-          "http://${nbLib.addressWithPort config.services.mempool.address config.services.mempool.port}"
+          "http://${nbLib.addressWithPort config.services.mempool.address config.services.mempool.port}/api"
         else
           null;
       description = "Mempool API endpoint.";
--- a/modules/albyhub.nix
+++ b/modules/albyhub.nix
@@ -13,10 +13,6 @@ let
   lnd = config.services.lnd;
   # Use loopback for client connections if lnd is bound to wildcard addresses
   lndRpcClientAddress = builtins.replaceStrings [ "0.0.0.0" "[::]" ] [ "127.0.0.1" "[::1]" ] lnd.rpcAddress;
-  torRelays = [
-    "ws://oxtrdevav64z64yb7x6rjg4ntzqjhedm5b5zjqulugknhzr46ny2qbad.onion"
-    "ws://bitcoinr6de5lkvx4tpwdmzrdfdpla5sya2afwpcabjup2xpi5dulbad.onion"
-  ];
 
   envFileContent =
     let
@@ -127,15 +123,11 @@ in
     };
 
     relay = mkOption {
-      type = with types; nullOr (either str (listOf str));
+      type = with types; nullOr (listOf str);
       example = [ "wss://relay.damus.io" "wss://offchain.pub" "wss://relay.primal.net" ];
-      default = if cfg.tor.proxy then torRelays else null;
+      default = null;
       apply = value: if builtins.isList value then concatStringsSep "," value else value;
-      description = ''
-        The default nostr relays to use. Find more relays:
-        - https://nostrwat.ch/
-        - https://github.com/0xtrr/onion-service-nostr-relays
-      '';
+      description = "The nostr relays to use. Note that onion relays are not supported due to proxy limitations in hub. Find more relays: https://nostrwat.ch/";
     };
 
     logLevel = mkOption {
--- a/modules/albyhub.nix
+++ b/modules/albyhub.nix
@@ -61,11 +61,11 @@ let
       ${optionalString (cfg.baseUrl != null) "BASE_URL=${cfg.baseUrl}"}
       ${optionalString (cfg.frontendUrl != null) "FRONTEND_URL=${cfg.frontendUrl}"}
       ${optionalString (cfg.logEvents != null) "LOG_EVENTS=${boolToString cfg.logEvents}"}
+      ${optionalString (cfg.relay != null) "RELAY=${cfg.relay}"}
+      ${optionalString (cfg.boltzApi != null) "BOLTZ_API=${cfg.boltzApi}"}
       ${optionalString (
         cfg.enableAdvancedSetup != null
       ) "ENABLE_ADVANCED_SETUP=${boolToString cfg.enableAdvancedSetup}"}
-      RELAY=${cfg.relay}
-      BOLTZ_API=${cfg.boltzApi}
       ${backendOpts}
       ${optionalString (cfg.extraConfig != null) cfg.extraConfig}
     '';
@@ -124,7 +124,8 @@ in
 
     relay = mkOption {
       type = with types; nullOr str;
-      default = "wss://relay.getalby.com/v1";
+      default = null;
+      example = "wss://relay.getalby.com,wss://relay2.getalby.com";
       description = "The default nostr relay.";
     };
 
@@ -167,15 +168,21 @@ in
     baseUrl = mkOption {
       type = with types; nullOr str;
       default = null;
-      description = "Base URL for the Alby Hub. Required if you want to connect to your Alby account.";
+      description = ''
+        Public base URL of your Alby Hub backend (used as `BASE_URL`).
+        This must be reachable by the browser and is required when using a custom Alby OAuth client,
+        because the OAuth callback is `${baseUrl}/api/alby/callback`.
+      '';
       example = "http://localhost:8082";
     };
 
     frontendUrl = mkOption {
       type = with types; nullOr str;
       default = null;
-      description = "Frontend URL for the Alby Hub";
-      example = "http://127.0.0.1:8082";
+      description = ''
+        Public URL of the Hub frontend UI used for browser redirects (for example after OAuth or logout).
+        If unset, redirects fall back to `baseUrl`.
+      '';
     };
 
     logEvents = mkOption {
@@ -198,11 +205,10 @@ in
 
     boltzApi = mkOption {
       type = with types; nullOr str;
-      default = "wss://api.boltz.exchange/v2/ws";
+      default = null;
+      example = "wss://api.boltz.exchange/";
       description = "Boltz API endpoint.";
     };
-
-    # Removed: relayBridge and boltzBridge options. Bridges are auto-managed when tor.proxy is enabled.
 
     extraConfig = mkOption {
       type = with types; nullOr str;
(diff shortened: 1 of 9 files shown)


modules/albyhub.nix
```diff
@@ -0,0 +1,372 @@
...
+
+with lib;
+let
+  cfg = config.services.albyhub;
+  nbLib = config.nix-bitcoin.lib;
+  secretsDir = config.nix-bitcoin.secretsDir;
+  bitcoind = config.services.bitcoind;
+  lnd = config.services.lnd;
+
+  boolToString = b: if b then "true" else "false";
```
prusnak, line 11: not needed because `lib.boolToString` exists

bleetube:
tl;dr a lot of testing finally yielded (possibly) a little progress here. and I need to do a lot of testing and cleanup

hub's go http client honors HTTP(S)PROXY and supports SOCKS5 proxies configured through env vars. 

go-nostr's HTTP/WebSocket stack honors HTTP(S)PROXY but does not support SOCKS5 proxies

I haven't run any tests for ldk-node yet, but it has client libs lightning-net-tokio which supports socks but not http proxies, and reqwest which is an http client. There's also bdk_esplora in ldk-node, but esplora-client should be talking to localhost only anyway.

socks5 env vars like `HTTP_PROXY=socks5://127.0.0.1:9050` would presumably work for ldk-node, but not go-nostr

```
Sep 08 15:22:42 squirtle albyhub[342031]: {"iteration":16,"level":"info","msg":"Connecting to the relay","relay_url":"wss://relay.getalby.com/v1","time":"2025-09-08T15:22:42-07:00"}
Sep 08 15:22:49 squirtle albyhub[342031]: {"error":"error opening websocket to 'wss://relay.getalby.com/v1': failed to WebSocket dial: failed to send handshake request: Get \"https://relay.getalby.com/v1\": connection took too long","iteration":16,"level":"error","msg":"Failed to connect to relay","retry_seconds":60,"time":"2025-09-08T15:22:49-07:00"}
```

torsocks does not work for hub's go http client:

```
Sep 08 15:54:29 squirtle torify[361169]: {"error":"could not connect to boltz ws at wss://api.boltz.exchange/v2/ws: dial tcp 143.202.162.204:443: i/o timeout","level":"error","msg":"Failed to connect to boltz websocket, retrying in 2s...","time":"2025-09-08T15:54:29-07:00"}
Sep 08 15:55:02 squirtle torify[361169]: {"error":"Get \"https://getalby.com/api/internal/info\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)","level":"error","msg":"Failed to fetch /info","time":"2025-09-08T15:55:02-07:00"}
```

> Go binaries don’t use glibc for networking, so LD_PRELOAD-based wrappers like torsocks/proxychains can’t reliably intercept sockets

Then I tried to get creative. Since go-nostr doesn't support socks5, tried using an http proxy. I used privoxy, even though tor now also has a `HTTPTunnelPort` option now (I'm assuming it basically behaves the same way).

using env vars to utilize both tor socks5 + privoxy http proxy (ie. `ALL_PROXY=socks5` and `HTTPS_PROXY=http`) still doesn't work for go-nostr (or probably ldk-node for that matter):

```
root@squirtle /e/nixos# doas -u nobody env HTTPS_PROXY=http://127.0.0.1:8118 curl -I https://relay.getalby.com/v1
HTTP/1.1 200 Connection established

HTTP/2 200 


root@squirtle /e/nixos# systemctl cat --runtime albyhub.service | grep -E '^ExecStart|^Env|^IPAddress' | grep -v PATH
Environment="ALL_PROXY=socks5://127.0.0.1:9050"
Environment="HTTPS_PROXY=http://127.0.0.1:8118"
Environment="HTTP_PROXY=http://127.0.0.1:8118"
Environment="NO_PROXY=127.0.0.1,localhost,::1"
ExecStart=/nix/store/pvpqkj6qzg36nrpp5x3wyx2612g13mjp-albyhub-1.19.2/bin/albyhub
ExecStartPre=+/nix/store/gib8jj2qfdj6fsi6q00drkcr4smcc3mj-albyhub-setup
IPAddressAllow=127.0.0.1/32
IPAddressAllow=::1/128
IPAddressAllow=169.254.0.0/16
IPAddressDeny=any
root@squirtle /e/nixos# journalctl -fu albyhub | grep -i error
Sep 09 10:32:38 squirtle albyhub[388745]: {"error":"error opening websocket to 'wss://relay.getalby.com/v1': failed to WebSocket dial: failed to send handshake request: Get \"https://relay.getalby.com/v1\": connection took too long","iteration":6,"level":"error","msg":"Failed to connect to relay","retry_seconds":60,"time":"2025-09-09T10:32:38-07:00"}
```

Regarding ldk-node, Lightning P2P via lightning-net-tokio does not honor ALL_PROXY/HTTP(S)PROXY and won’t be torified by those env vars alone.

so I think both go-nostr and ldk-node either need torsocks, or socks support configured/added directly to their tcp client library.

ldk-node uses several dependencies with uneven support for proxy configuration:

- reqwest (HTTP client) - Supports HTTP_PROXY/HTTPS_PROXY environment variables
- lightning-net-tokio - Uses tokio-socks for SOCKS proxy support
- bdk_esplora and esplora-client - HTTP clients that respect proxy settings (although electrum would always be via localhost and not proxied in nix-bitcoin)

I also tried ugly hacks like using both env vars and torsocks, which can result in double proxying. go-nostr was still broken in that configuration.

ldk-node might work if all proxy vars are configured to use tor socks5, or we use torsocks. But torsocks won't work with Go's http client (in hub), so that leaves us with the best option of using socks5 env vars and leaving go-nostr in a broken state.

I think the next move is to submit a PR to go-nostr to add socks5 support.
After that, the big goal would be to ask for (or build) native tor configuration directly in hub.

Oh wait, maybe part of the issue is that you can't do websocket connections over an http tunnel. I tried another workaround where I bridge the websocket connections to get them to talk over socks5 using this syntax:

```
websocat -b --ping-interval=30 --exit-on-eof \
  --socks5 127.0.0.1:9050 \
  ws-l:127.0.0.1:8485 \
  wss://relay.getalby.com/v1
```

That's go-nostr -> websocat -> privoxy -> tor. The big downside being that there's no TLS between hub and websocat, but that's inside the local environment at least.

I stopped seeing those errors, so maybe it works. I'll continue testing as time permits.

quick thoughts on some cleanup I should probably do: I don't think anything cares about all_proxy, so I'll probably remove that. I should probably cleanup the service.albyhub.after properties is probably overkill and should be removed too. the getPublicAddressCmd option is probably pointless in nix-bitcoin too.



Apologies for the commit spam, I only realized this morning that I can just test my changes without pushing them somewhere by using `nix-bitcoin.url = "path:/etc/nixos/nix-bitcoin";` in my flake (and pushing my changes there instead of here).

bleetube:
the websocket bridge workaround adds more complexity (and maintenance overhead) than I like and breaks the UX of NWC setup, since the user gets a localhost relay address rather than the configured relay uri. the right move here is to get PRs merged upstream to add socks5 support for websockets to the GO HTTP client libs for go-nostr.

naturallaw777:
I created this issue on go-nostr: https://github.com/nbd-wtf/go-nostr/issues/191

bleetube:
interesting, I'll have to retest with that. I've been meaning to get back to this PR, just fell down the rabbit hole of reworking all my systems to use nix flakes, sops-nix, and a bunch of other yak shaving.

bleetube:
I finally just got nostr relay connectivity working over tor.

When Tor enforcement is enabled, Hub’s nostr relay connection fails behind an HTTP proxy because `go-nostr` takes a different websocket dial path when a custom request header is set; that path bypasses proxy-from-environment and attempts direct DNS/TCP egress. go-nostr's repo is archived, so my interim workaround is to stop setting the websocket `User-Agent` header in Hub (`service/start.go`), forcing `go-nostr` to use its default dial path that honors `HTTP_PROXY`/`HTTPS_PROXY` and restores nostr relay connectivity in nix-bitcoin.

```
    services.albyhub = {
      enable = true;
      package = unstablePkgs.albyhub.overrideAttrs (old: {
        postPatch = (old.postPatch or "") + ''
          substituteInPlace service/start.go --replace-fail '"net/http"' ""
          sed -i '/nostr.WithRequestHeader(http.Header{/,/}),/d' service/start.go
        '';
      });
      tor = {
        enforce = true;
        proxy = true;
      };
    };
```

I've confirmed my zaps are working with this configuration, along with nix-bitcoin's LND instance.

bleetube:
Here's quick summary of this module's status.

**Defects**:
- nostr connectivity is broken, but an upstream [PR has been submitted](https://github.com/getAlby/hub/pull/2190) to fix this already. if it gets merged, then `hub` is at least usable with tor.enforce along with LND. Maybe pheonix as well, but I haven't tested that
- LDK node probably won't work when tor is on (upstream needs proxy support)

**Features**:
- Optionally generates an onion for the web interface (I still need to verify it works)
- Autoconfigures LND integration if lnd is enabled
- if Tor isn't enforced, everything including the internal LDK node works
- Everything mentioned in the PR description above remains true.

I think if that PR gets merged upstream, or some comparable change, then this module would be ready for consideration. So I'm marking this draft as ready for review.

bleetube:
Upstream didn't like the workaround. Closing due to lack of support for tor enforcement.

PR #825, opened 2025-12-24
LND: add neutrino backend support

PatMulligan:
Add option to use neutrino instead of bitcoind for fetching blockchain data. This allows running LND as a lightweight client that connects to remote Bitcoin full nodes.

New options:
- services.lnd.backend: choose between "bitcoind" (default) or "neutrino"
- services.lnd.neutrino.addpeers: list of Bitcoin node peers to connect to
- services.lnd.neutrino.feeUrl: fee estimation URL

Example configuration:

```nix
  services.lnd = {
    enable = true;

    backend = "neutrino";
    neutrino.addpeers = [
      "btcd1.lnolymp.us"
      "btcd2.lnolymp.us"
      "btcd-mainnet.lightning.computer"
      "uswest.blixtwallet.com"
      "node.eldamar.icu"
      "noad.sathoarder.com"
      "bb1.breez.technology"
      "bb2.breez.technology"
      "neutrino.shock.network"
    ];

    neutrino.feeUrl = "https://nodes.lightning.computer/fees/v1/btc-fee-estimates.json" # default value

  };
```

### DIFF
--- a/modules/lnd.nix
+++ b/modules/lnd.nix
@@ -150,12 +150,12 @@ let
       description = "The backend to use for fetching blockchain data.";
     };
     neutrino = {
-      addpeers = mkOption {
+      peers = mkOption {
         type = types.listOf types.str;
         default = [];
         example = [ "192.168.1.1:8333" "btcd.example.com:8333" ];
         description = ''
-          List of Bitcoin full node peers to connect to via neutrino.addpeer.
+          List of Bitcoin full node peers to connect to for neutrino.
           Multiple peers provide redundancy for maximum uptime.
         '';
       };
@@ -206,7 +206,7 @@ let
       bitcoind.zmqpubrawtx=${zmqHandleSpecialAddress bitcoind.zmqpubrawtx}
     '' else ''
       bitcoin.node=neutrino
-      ${lib.concatMapStringsSep "\n" (peer: "neutrino.addpeer=${peer}") cfg.neutrino.addpeers}
+      ${lib.concatMapStringsSep "\n" (peer: "neutrino.addpeer=${peer}") cfg.neutrino.peers}
       fee.url=${cfg.neutrino.feeUrl}
     ''}
 
@@ -232,10 +232,10 @@ in {
           services.lnd.port to a port other than 9735.
         '';
       }
-      { assertion = cfg.backend != "neutrino" || cfg.neutrino.addpeers != [];
+      { assertion = cfg.backend != "neutrino" || cfg.neutrino.peers != [];
         message = ''
           When using neutrino backend, you must configure at least one peer
-          in services.lnd.neutrino.addpeers.
+          in services.lnd.neutrino.peers.
         '';
       }
     ];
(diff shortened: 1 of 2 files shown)


modules/lnd.nix
```diff
@@ -144,6 +144,30 @@ let
       description = "LND TLS certificate path.";
     };
     tor = nbLib.tor;
+    backend = mkOption {
+      type = types.enum [ "bitcoind" "neutrino" ];
+      default = "bitcoind";
+      description = "The backend to use for fetching blockchain data.";
+    };
+    neutrino = {
+      addpeers = mkOption {
```
prusnak: wondering whether this should rather be `peers` instead of `addpeers`?

modules/lnd.nix
```diff
@@ -288,7 +324,7 @@ in {
     users.users.${cfg.user} = {
       isSystemUser = true;
       group = cfg.group;
-      extraGroups = [ "bitcoinrpc-public" ];
+      extraGroups = optional (cfg.backend == "bitcoind") "bitcoinrpc-public";
```
prusnak: `optional` is very cryptic and it is not obvious it is adding an element into the list, suggesting using `optionals` instead:

```suggestion
      extraGroups = optionals (cfg.backend == "bitcoind") [ "bitcoinrpc-public" ];
```

modules/lnd.nix
PatMulligan: yeah i think it makes sense!

modules/lnd.nix
```diff
@@ -144,6 +144,37 @@ let
...
     tor = nbLib.tor;
+    backend = mkOption {
+      type = types.enum [ "bitcoind" "neutrino" ];
+      default = "bitcoind";
+      description = "The backend to use for fetching blockchain data.";
+    };
+    neutrino = {
+      peers = mkOption {
+        type = types.listOf types.str;
+        default = [];
```
ekzyis, line 155: nit: could add defaults here, too, or was it intentionally left out? I see there's a default for `feeUrl`.

modules/lnd.nix
```diff
@@ -144,6 +144,37 @@ let
       description = "LND TLS certificate path.";
     };
     tor = nbLib.tor;
+    backend = mkOption {
+      type = types.enum [ "bitcoind" "neutrino" ];
+      default = "bitcoind";
+      description = "The backend to use for fetching blockchain data.";
+    };
```
ekzyis, line 151: nit: I think `lnd.neutrino.enable` xor `lnd.bitcoind.enable` would be more common instead of a new `lnd.backend` option.

ekzyis (approved):
I'm running this [rebased on master](https://github.com/ekzyis/nix-bitcoin/tree/add-lnd-neutrino) and my new LND v0.20.1-beta node synced fine. Neutrino config written to /var/lib/lnd/lnd.conf is as expected.

Left two nits, else LGTM!

> should I add any more config options or maybe just let it be handled in services.lnd.extraConfig...?

I think the existing basic config is fine. No need to attempt to include all config options. I really like how simple this was to setup. Thank you for this PR! You saved me time.

PR #827, opened 2026-02-06
update nixpkgs

jonasnick:
bitcoin: 30.0 -> 30.2
bitcoind: 30.0 -> 30.2
bitcoind-knots: 29.2.knots20251110 -> 29.3.knots20260210
btcpayserver: 2.2.1 -> 2.3.4
clightning: 25.09.2 -> 25.12.1
elementsd: 23.2.4 -> 23.3.2
lnd: 0.19.3-beta -> 0.20.1-beta

## Draft

This is a draft because it contains a workaround to deal with the following problem:

After updating LND from 0.19.3-beta to 0.20.0-beta, one of the test fails because both lightning-loop and lightning-pool crash on startup.

Sequence of events:
1. LND appears to be available (lncli getinfo works)
2. Both lightning-loop and lightning-pool fail with: `rpc error: code = Unknown desc = chain notifier RPC is still in the process of starting`
   and they don't seem to recover.

In the failing test scenario, 100 blocks are mined at startup and LND needs to sync to these blocks.

With the new lnd version are we supposed to explicitly check that `lncli getinfo 2>&1 | grep -q '"synced_to_chain": true'` before starting up lightning-loop/pool?

CC original contributor of the pool pkg/module @sputn1ck

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -44,11 +44,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1767480499,
-        "narHash": "sha256-8IQQUorUGiSmFaPnLSo2+T+rjHtiNWc+OAzeHck7N48=",
+        "lastModified": 1770136044,
+        "narHash": "sha256-tlFqNG/uzz2++aAmn4v8J0vAkV3z7XngeIIB3rM3650=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "30a3c519afcf3f99e2c6df3b359aec5692054d92",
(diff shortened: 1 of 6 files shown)


jonasnick:
I updated nixpkgs again and that replaced lnd 0.20.0 with 0.20.1. Unfortunately, I still see the same issues with the loop and pool modules.

jonasnick:
The hack for loop isn't necessary anymore because the problem was fixed in https://github.com/lightninglabs/loop/commit/bde49d7a which shipped in version 0.31.6 and the current nixpkgs includes version 0.33.0.

jonasnick:
superseded by #840

